PR #13, opened 2021-03-10
Port infoview.

gebner:
This PR resurrects the infoview from the Lean 3 extension.


infoview/goal.tsx
```diff
@@ -0,0 +1,66 @@
...
+                goals.push(match[1])
+            }
+        } while (match)
+    }
+
+    if (!(reFilters.length === 0 || filterIndex === -1)) {
+        goals = goals.map((g) =>
+            // this regex splits the goal state into (possibly multi-line) hypothesis and goal blocks
+            // by keeping indented lines with the most recent non-indented line
+            g.match(/(^(?!  ).*\n?(  .*\n?)*)/mg).map((line) => line.trim())
```
Kha, line 42: Should we make `goals` even more structured? For example, I thought about using the case name, if available, as the heading. We should also expose the pre-state while we're at it, so maybe `goals` isn't the best name in the first place.

Kha: Yeah, the benefit would be negligible. By "pre-state" I meant the goals before execution of the tactic. Currently `goals` is always the post-state.

Issue #22, opened 2021-04-20
Imports fail with 'no such file or directoy'


tydeu:
When running Lean 4 nightly (commit `26dda3f63d88`) with `vscode-lean4` v0.0.24 on Windows, I am encountering the following error on any `import`:

```
`leanpkg print-paths` failed:

stderr:
configuring <package>
> sh -c "<home>\.elan\toolchains\leanprover-lean4-nightly\bin/leanmake" LEAN_OPTS="" LEAN_PATH="././build" "build\<Foo.olean>" >&2    # in directory .
uncaught exception: no such file or directory (error code: 2)
```

It seems to be failing because it is adding `build` both to `LEAN_PATH` and to the path of `Foo.olean`. Reopening VS Code, reloading the window, restarting the lean server, and/or refreshing the dependencies does not fix the issue. 

Furthermore, running `leanpkg print-paths` myself in the directory works:
```
$ leanpkg print-paths
configuring <package>
././build
./.
```
So I am not quite sure what is going on.

However, I can report that this error is not present when using the stable build of Lean 4.

Kha:
The missing file is most likely `sh`, which would explain why it's not failing in your bash. And no, we really shouldn't depend on `sh` and `make` on Windows. I will likely revisit that part of `leanpkg` soon.

Note that this is an issue with Lean, not vscode-lean4.

tydeu:
@Kha Which part is calling `sh`? Because I can run `leanpkg print-paths` fine within the VS Code integrated terminal (which I have setup to be MSYS2). Is `vscode-lean4` running `leanpkg print-paths` or is this some process spawned by lean? If `vscode-lean4` is running it, it should be able to inherit the environment of the integrated terminal and work correctly. 

Also, why is this just a problem with the nightly, but not with the stable release? Did the `sh` dependency come later?

Kha:
The Lean server is now calling `leanpkg print-paths`, which will call `sh/make`, by itself to make sure that imports are up to date. So if VS Code itself (and not its terminals) does not have `sh` in its PATH, neither will the Lean server and leanpkg started from it, leading to the error. Meaning you could running VS Code (after closing all instances) from *inside* an msys2 shell and it should work, possibly?

tydeu:
@Kha Running VS Code inside a msys2 shell does work. However, this completely messes with my environment. This problem could be better resolved if I could configure the environment of the Lean server process. Luckily, while digging through the extension codebase, I found that this can be done. 

[...]

PR #23, opened 2021-04-21
Added option to configure Lean server environment PATH

tydeu:
Addresses #22, enabling users to add paths to PATH environment variable used for the Lean 4 language server. I needed this, so I decided to go ahead and code it myself.


tydeu:
This can be further generalized to allow the configuration of arbitrary environment variables. However, care still has to be taken with `Path`,  as one generally wants to add to it not replace it completely. I pushed an additional commit that allows demonstrates such an extension.

Kha:
FWIW this change looks good to me even if longer term this should be resolved by https://github.com/leanprover/lean4/issues/419

package.json
```diff
@@ -59,6 +59,24 @@
...
+          "description": "Environment variables to add to the Lean 4 server environment",
+          "additionalProperties": {
+            "type": "string",
+            "description": "environment variable to add"
+          }
+        },
+				"lean4.serverEnvPaths": {
+					"type": "array",
+					"default": [],
+					"description": "Paths to add to the Lean 4 server environment.",
```
gebner, line 74:
```suggestion
					"description": "Paths to add to the Lean 4 server environment PATH variable.",
```

src/leanclient.ts
```diff
@@ -61,17 +62,23 @@ export class LeanClient implements Disposable {
         if (this.isStarted()) {
             await this.stop()
         }
+        const env = Object.assign({}, process.env, serverEnv());
+        const paths = serverEnvPaths()
+        if (paths.length != 0) {
+            const PathKey = env.Path ? 'Path' : 'PATH'
+            env[PathKey] = paths.join(path.delimiter) + path.delimiter + process.env.Path
```
gebner, line 69:
```suggestion
            const pathKey = process.platform === 'win32' ? 'Path' : 'PATH'
            env[pathKey] = paths.join(path.delimiter) + path.delimiter + process.env.Path
```

gebner (approved):
Please also run `npm i` to reformat the `package.json` file.

tydeu:
@gebner Done! 

P.S. I had previously used `npm ci` because I wasn't sure you would want an updated lock file, which `npm i` creates.

gebner:
Thanks.  I don't care too much about the lockfile, but the package.json file should be clean.

PR #24, opened 2021-04-29
add user paths for version check, fix path field access

ammkrn:
The initial version check is done before adding the path variables specified in the user's config file, so if the user is counting on that to set the lean path, the version check will always fail. `exec` defaults to `/bin/sh`, so if that's not your login shell, the vscode config is kind of the only way to do it. Alternatively, the extension could use SHELL to figure out what shell is actually being used and pass that to exec.
```
    const user_shell = process.env.SHELL
...
        const { stdout, stderr } = await promisify(exec)(cmd, {cwd: folderPath, shell: user_shell })
```

I took the code to update the process path from `leanclient.ts`, but I think `paths.join(path.delimiter) + path.delimiter + process.env.Path` needs to be broken up into either `process.env.Path` (for windows) and `process.env.PATH` for everything else, as you did with the pathKey getter. As of 122c7e3, the process ends up with `:undefined` tacked onto the end of PATH for non-windows systems.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,12 +1,11 @@
 import { workspace, commands, window, languages, ExtensionContext, TextDocument } from 'vscode'
 import { promisify } from 'util'
-import { exec } from 'child_process'
+import { execFile } from 'child_process'
 import { AbbreviationFeature } from './abbreviation'
 import { executablePath, addServerEnvPaths } from './config'
 import { LeanClient } from './leanclient'
 import { InfoProvider } from './infoview'
 import { LeanTaskGutter } from './taskgutter'
-import * as path from 'path'
 
 async function checkLean4(): Promise<boolean> {
     const folders = workspace.workspaceFolders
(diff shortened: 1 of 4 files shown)


gebner:
Oh wait, exec uses the shell?  That's really unnecessary.  Unless you want to look into that, I'll refactor it.

Could you please extract the whole environment variable setup into a separate function, so that it is not duplicated between the two files?

> As of 122c7e3, the process ends up with :undefined tacked onto the end of PATH for non-windows systems.

Well spotted!

ammkrn:
> Could you please extract the whole environment variable setup into a separate function, so that it is not duplicated between the two files?

It's now in src/config.ts with the other utility functions for the user config.

> Oh wait, exec uses the shell? That's really unnecessary. Unless you want to look into that, I'll refactor it.

The docs say `exec` spawns a shell by default but `execFile` doesn't:
```
The child_process.execFile() function is similar to child_process.exec() except that it does not spawn a shell by default. Rather, the specified executable file is spawned directly as a new process making it slightly more efficient than child_process.exec().
```

f7feac1 swaps out execFile for exec.

src/extension.ts
```diff
@@ -13,17 +13,20 @@ async function checkLean4(): Promise<boolean> {
     if (folders) {
         folderPath = folders[0].uri.fsPath
     }
-    const cmd = `${executablePath()} --version`
+    
+    const env = addServerEnvPaths(process.env);
+    const cmd = `${executablePath()}`
```
gebner, line 18:
```suggestion
    const cmd = executablePath()
```
Or am I missing something?

ammkrn: You're not missing anything, no.

PR #25, opened 2021-05-26
Expose configurable elaboration delay

Vtec234:
Users may set how long they wish the server to wait after the last edit before starting elaboration. See also Zulip [thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Has.20Lean.204.20gotten.20slower.3F).


gebner:
Ah, I didn't know that this option was already implemented.  Thanks!

PR #28, opened 2021-06-09
Adding a changelog

mhuisi:
This PR adds a changelog to the README. I regret not starting one when I transferred the repo to leanprover-community, so I reconstructed a full change log now. @Kha also mentioned to me that it would be helpful to keep the pinned Nix version up to date. The PR also updates the functionality section in the README.

For two changes I couldn't figure out what it is they did without further context. I've marked those with `(?)` to indicate that they still need an understandable description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ As of now, the extension supports the following features:
 - Diagnostics
 - Syntax highlighting with basic Lean 4 syntax rules
 - `\foo` unicode shortcuts
-- A simple infoview since [leanprover/lean4@ec903f58](https://github.com/leanprover/lean4/commit/ec903f58d294cc7c57008a9cf754745fa7024f8c) (2021-02-04)
+- Semantic highlighting since [leanprover/lean4@5df753f3](https://github.com/leanprover/lean4/commit/5df753f3381dca714d0e81727d38feba765d6234) (2021-03-17)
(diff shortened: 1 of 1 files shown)


gebner:
Thanks for making the effort to update the readme!

But I fear that such a fine-grained changelog will just get out of date.  AFAICT, this is just a list of the commits, right?  I don't think anybody cares when I removed the `webview-rpc` library for example.  (I also don't want to update the changelog every time I make a commit.)

I believe it would be more useful to have just big features and new configuration options in the changelog (together with links to relevant PRs and issues on the Lean repo).  If somebody wants to see the full history, they can always run `git log`.

mhuisi:
Oh right, one reason why I included fixes and the like is because I thought that it might be of use for @Kha to quickly determine whether the pinned Nix version needs to be updated.

gebner:
> I suppose one could just write the changelog in a way that combines multiple versions if all but one of them are trivial.

I think that's a good idea.  I'm making releases for every tiny fix.

> Should fixes be documented in the changelog?

I think that just drowns out the other information.  I'd prefer if they weren't included.  (Including fixes also increases the effort required to keep the changelog up to date.)

> whether the pinned Nix version needs to be updated.

The answer to this question is always "yes".

Kha:
Yeah, this was just a single instance of me wanting to quickly find out what version a commit was in, which for Lean I do via the autogenerated changelogs. I used the the comparison button instead.

mhuisi:
I hope that this is easier to maintain. I also added links to discussions of the respective feature where I found the discussion to be important.

gebner:
This is great, thanks!

PR #30, opened 2021-06-25
LSP-based infoview

Vtec234:
This PR splits the infoview into a separate NPM package independent of VSCode, and based on a more general editor<->infoview API based on LSP. Changes include:
- The new [infoview API](https://github.com/Vtec234/vscode-lean4/blob/44a9bf3a8c7aaea99bd54c52f322859d96935507/lean4-infoview/src/infoviewApi.ts) tries to provide more general functionality, basing most of it on LSP requests and the ability to listen in on incoming/outgoing LSP notifications from within the infoview. This is intended to allow using custom requests from widget code.
  - This is not as nice to implement on the VSCode extension side as one might like. See [here](https://github.com/Vtec234/vscode-lean4/blob/44a9bf3a8c7aaea99bd54c52f322859d96935507/vscode-lean4/src/infoview.ts#L42) for why. This may be fixable in upstream `vscode-jsonrpc`.
  - I expect the new API to be more easily implementable in any other text editor capable of displaying a webpage.
- Fixes in various edge cases. For example, we now delete pins when the corresponding file is closed as the server cannot produce goal states for them any more.
- We handle more stuff in the infoview. For example, pins are synced there without going through the extension by subscribing to `textDocument/didChange`.
- Outsourced tachyons CSS and codicons to NPM packages. I took the artistic liberty of changing the codicon for copy-to-comment to "quote" (see it [here](https://microsoft.github.io/vscode-codicons/dist/codicon.html)) as I couldn't find the current one (and maybe "quote" is a little clearer?).
- Some refactoring, primarily on `info.tsx`. UI code is difficult, but I tried to separate the concerns of pausing and fetching goals into different and maybe reusable components.
- The repository is now organised into a "monorepo" containing two packages. We use Lerna to manage them.
- Probably other things I forgot.

I apologise for making the development a bit hard to follow. I made commits to a separate repo before copying the files here. They can be [browsed](https://github.com/Vtec234/lean4-infoview/commits/master).

There are quite a lot of changes, and while I tried testing all functionality I expect there to be some breakage. Still to be fixed:
- [x] Re-send diagnostics to infoview on startup
- [x] Implement "copy-to-comment" (~~now a general "make text edit" API~~)


vscode-lean4/src/leanclient.ts
```diff
@@ -1,36 +1,20 @@
...
-
-    /**
-     * Array containing the parts of the file which are still being processed.
-     * The array should be empty if and only if the server is finished processing.
-     */
-    processing: LeanFileProgressProcessingInfo[];
-}
-
-const processingMessage = 'processing...'
+import { PlainGoal, PlainTermGoal, LeanFileProgressParams } from '@lean4/infoview';
```
gebner, line 17: This import doesn't work for me (red squiggles from tsserver).  I've run `npx lerna bootstrap`.

vscode-lean4/src/rpc.ts
```diff
@@ -18,7 +18,7 @@ export class Rpc {
                 try {
                     this.sendMessage({ seqNum, result: await this.methods[name](...args) })
                 } catch (ex) {
-                    this.sendMessage({ seqNum, exception: ex === undefined ? 'error' : ex })
+                    this.sendMessage({ seqNum, exception: ex === undefined ? 'error' : ex.toString() })
```
gebner, line 21: Is there a reason for this change?  I'm not sure what it fixes, but it breaks exceptions like `{code: 3278}` which are now rendered as `[object Object]`.

vscode-lean4/src/rpc.ts
```diff
@@ -33,13 +33,15 @@ export class Rpc {
     invoke(name: string, args: any[]): Promise<any> {
         this.seqNum += 1
         const seqNum = this.seqNum
-        this.sendMessage({seqNum, name, args})
-        return new Promise((resolve, reject) => this.pending[seqNum] = {resolve, reject})
+        return new Promise((resolve, reject) => {
```
gebner, line 36: I don't think there's a semantic difference here, but ok.

lean4-infoview/src/infoview/messages.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+const MessageView = React.memo(({uri, diag}: MessageViewProps) => {
+    const ec = React.useContext(EditorContext);
+    const fname = escapeHtml(basename(uri));
+    const {line, character} = diag.range.start;
+    const loc: Location = { uri, range: diag.range };
+    const shouldColorize = diag.severity === DiagnosticSeverity.Error;
+    let text = escapeHtml(diag.message)
+    text = shouldColorize ? colorizeMessage(text) : text;
+    const severityClass: {[name: number]: string} = diag.severity ? {
```
gebner, line 23:
```suggestion
    const severityClass: {[name: number]: string} = diag.severity ? {
```
This type annotation is almost certainly not what is intended.  (Surprisingly it is still correct though, because every string is also a {[_:number]: string}.)

vscode-lean4/src/rpc.ts
Vtec234: I think there's a potential race if `sendMessage` comes back with `messageReceived` immediately and tries to read `this.pending[seqNum]` before the promise has a chance to store `{resolve, reject}`. Putting both in the promise linearizes this so it cannot happen.

vscode-lean4/src/rpc.ts
Vtec234: Yes, the extension-side was receiving exceptions with only closure members in the prototype (like a custom `ex.toString`) which couldn't be JSON-serialized and the infoview saw an empty object. But I see that there are other cases in which we do not want to call that. I'll add some cases.

vscode-lean4/src/leanclient.ts
Vtec234: Hmm, I'm not quite sure. Which version of NPM are you using? I'm on 7.19.0. I can check with another one. There was a separate ESLint directory issue but TS proper should work.

lean4-infoview/src/infoview/index.css
```diff
@@ -1,16 +1,22 @@
 
 /* Some styles specific to the infoview. */
 
+html,body {
+    height: 100%;
```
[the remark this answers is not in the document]
Vtec234: It allows page elements to stretch to the bottom of the page, see [here](https://stackoverflow.com/a/579123).

vscode-lean4/src/leanclient.ts
gebner: I'm using 7.18 here.

vscode-lean4/src/rpc.ts
gebner: Ok, I think I understand.  Note that this can only happen if you call messageReceived manually from sendMessage.

lean4-infoview/src/infoview/util.ts
```diff
@@ -0,0 +1,180 @@
...
+}
+
+export function useServerNotificationEffect<T>(method: string, f: (params: T) => void, deps?: React.DependencyList): void {
+  const ec = React.useContext(EditorContext);
+  React.useEffect(() => {
+    void ec.api.subscribeServerNotifications(method).then(
+      () => {},
+      (ex) => { console.error(`Failed subscribing to server notification '${method}': ${ex}`); }
+    );
+    ec.events.gotServerNotification.on(([thisMethod, params]: [string, T]) => {
```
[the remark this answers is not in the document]
Vtec234: Thanks!

lean4-infoview/src/infoview/goal.tsx
```diff
@@ -30,19 +26,19 @@ function emphasizeMessage(goal: string): string {
         .replace(/^([^:\n< ][^:\n⊢{[(⦃]*) :/mg, '<strong class="goal-hyp">$1</strong> :')
 }
 
-export function Goal({plainGoals}: GoalProps): JSX.Element {
+export function Goal({plainGoal}: {plainGoal: PlainGoal}): JSX.Element {
+    if(!plainGoal) return <></>;
```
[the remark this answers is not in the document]
Vtec234: Ah yes, this is a leftover, thank you. Actually `plainGoal` should never be null now (strict type checking is on), so the check can be removed.

lean4-infoview/src/infoview/goal.tsx
```diff
@@ -69,7 +65,7 @@ export function Goal({plainGoals}: GoalProps): JSX.Element {
 }
 
 export function TermGoal({termGoal}: {termGoal: PlainTermGoal}): JSX.Element {
-    if (!termGoal?.goal) return null;
+    if(!termGoal) return <></>;
```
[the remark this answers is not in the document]
Vtec234: Same as above, check is unnecessary.

vscode-lean4/src/rpc.ts
Vtec234: I thought it could still happen if a `message` event with a response comes in before the `Promise` executes.

lean4-infoview/src/infoview/info.tsx
```diff
@@ -0,0 +1,266 @@
...
+    return (
+    <Details initiallyOpen>
+        <InfoStatusBar {...props} triggerUpdate={triggerDisplayUpdate} isPaused={isPaused} setPaused={setPaused} copyGoalToComment={copyGoalToComment} />
+        <div className="ml1">
+            {status === 'error' && error &&
+                <div className="error">
+                    Error updating: {error}.
+                    <a className="link pointer dim" onClick={e => { e.preventDefault(); triggerDisplayUpdate(); }}>Try again.</a>
+                </div>}
+            {status !== 'error' && goal &&
```
gebner, line 128: I didn't notice this before because github doesn't show the diff.  There is a good reason why we are using `display: none` in addition to conditional rendering: with `display: none`, the state of the details element is preserved if you move around the file (i.e. whether it's collapsed or not).  This allows you e.g. to disable the error message display, or the term goal display.

Vtec234: Apologies about the diff, I was be careful to make the move commit contain Renames rather than Add/Delete, but alas it doesn't seem to work for all files. Yes, I see. I think this means we should put the stateful part (`Details`) in `Info` and the pure part in `Goal`. I will fix.

lean4-infoview/src/infoview/info.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+    const [status, setStatus] = React.useState<InfoStatus>('loading');
+    const [goal, setGoal] = React.useState<PlainGoal>();
+    const [termGoal, setTermGoal] = React.useState<PlainTermGoal>();
+    const [error, setError] = React.useState<string>();
+
+    const messages = useMessagesFor(pos);
+    const [serverIsProcessing, _] = useServerNotificationState(
+        '$/lean/fileProgress',
+        false,
```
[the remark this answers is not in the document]
Vtec234: I think this is the same problem as with diagnostics, namely that the editor needs to re-send the notification it received from the server on infoview startup. I'm not seeing a nice way to solve this within LSP, so I'll do what the current version does which is re-send diagnostics and progress on infoview startup.

vscode-lean4/src/infoview.ts
```diff
@@ -18,98 +17,135 @@ export class InfoProvider implements Disposable {
...
+                return;
+            }
 
-        reveal: async (loc) =>
-            await this.revealEditorPosition(Uri.parse(loc.uri), loc.line, loc.character),
+            // NOTE(WN): For non-custom notifications we cannot call LanguageClient.onNotification
+            // here because that *ovewrites* the notification handler rather than registers an extra one.
+            // So we have to add a bunch of event emitters to `LeanClient.`
+            if (method.startsWith('$')) {
+                const h = this.client.client.onNotification(method, (params) => {
```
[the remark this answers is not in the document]
Vtec234: What do you mean by "default handler"? As far as I could tell there can only ever be one handler (`vscode-jsonrpc` code).

vscode-lean4/src/leanclient.ts
Vtec234: That should definitely work. Do you have `node_modules/@lean4/infoview/` in `vscode-lean4/`? It should contain a symlink to the infoview directory.

gebner: Yes, I have that directory.

eslint is complaining about a missing tsconfig though, could this be a problem for typescript as well?
```
Parsing error: Cannot read file '/home/gebner/vscode-lean4/tsconfig.json'.
```

vscode-lean4/src/infoview.ts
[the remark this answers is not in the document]
gebner: There's something about StarNotificationHandler in vscode-languageserver-node, which looks like a fallback handler that is used if no other notification handler applies.

lean4-infoview/src/infoview/info.tsx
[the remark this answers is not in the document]
gebner: It will also be initialized with the wrong value of the Info is opened while the extension is running (e.g. when pinning).  The false here should really be something like `isLoading(currentFileProgressParams, pos)`.

vscode-lean4/src/infoview.ts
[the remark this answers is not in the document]
Vtec234: Oh yeah, there is a `starNotificationHandler` but it only gets called when there is no handler registered for that particular method, so we still need the hacks for builtin ones. But I guess we can go with the middleware hack for didChange and publishDiagnostics, and use a default for all custom ones. Good idea!

vscode-lean4/src/leanclient.ts
gebner: Hah! It works once you run `npm run build`.

Vtec234: Oh yeah, the infoview `dist/` has to be available. I'll add that to the README.

lean4-infoview/src/infoview/info.tsx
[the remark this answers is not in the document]
Vtec234: I think you're right, this should be top-level infoview state like the diagnostics rather than per-component state.

vscode-lean4/src/rpc.ts
Vtec234: I can't find the case with closures anymore, reverting this change.

README.md
```diff
@@ -1,93 +1,12 @@
-# Lean 4 VSCode Extension
-This extension provides VSCode editor support for the Lean 4 programming language.
-It is based on [vscode-lean](https://github.com/leanprover/vscode-lean).
+This repository contains NPM packages providing editor integration for the Lean 4 programming language.
```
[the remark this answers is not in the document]
Vtec234: I suppose we could also point people at the `vscode-lean4/` folder, but for now I moved the README back to root.

vscode-lean4/src/leanclient.ts
```diff
@@ -163,9 +160,6 @@ export class LeanClient implements Disposable {
                 }
             },
         }
-        this.setProgress(Object.assign({}, ...workspace.textDocuments
```
Vtec234, line 157: Removing this changes the semantics slightly in that yellow bars only appear when the server actually begins compiling the header. I think this is nice, as visually indicates things are working.

lean4-infoview/src/infoviewApi.ts
```diff
@@ -40,10 +46,10 @@ export interface EditorApi {
    */
   //emulateServerNotification(method: string, params: any): Promise<void>;
 
-  /** Apply edits to the workspace. */
-  applyEdits(edits: TextDocumentEdit[]): Promise<void>
+  /** Insert text into a document. When `pos` is not present, write at the current cursor location. */
+  insertText(text: string, kind: TextInsertKind, pos?: TextDocumentPositionParams): Promise<void>
```
Vtec234, line 50: Sadly the range of things which can be done with `TextEdit` without knowing the file contents (and for now we don't really need to track them in the infoview) is quite small, so reinstated this API.

vscode-lean4/src/leanclient.ts
gebner: I don't have extremely strong feelings about this, but it was very much intentional.  When opening a Lean file, I'm interested in whether there are errors or not.  The progress bars allow me to distinguish the states of "no errors because Lean hasn't compiled this part yet" and "no errors because the proof is fine".

This is the same reason why I said that the fileProgress notification should be sent *before* processing is started (in https://github.com/leanprover/lean4/issues/503).  The important information is not that Lean is still running, but that it is not finished yet.

Vtec234: I see your point. The only real difference is that with this change, if the server fails to start the gutters won't appear, whereas before they'd be there even though nothing is actually being processed. You still know that the lack of errors is due to server failure because there will be the "server failed to start" popup. IMO this way it's clearer when things have failed, and when things are actually compiling (where we do send server notifs before starting compilation as you note), and you can still clearly tell why there are no errors. But if you still disagree, I'm happy to revert this change.

gebner: Ah, if this only happens with a dead server then it's a non-issue.  In that case you already get a lot of errors anyhow.

Vtec234: Well, there are also no gutters between `exec lean --server` and "`lean --worker` sends `progress @ 0`". This is a few ms right now, but I suppose it could be a problem with a bad plugin which takes long to initialize. We could show [progress](https://github.com/microsoft/vscode-extension-samples/tree/main/progress-sample/src) then.

gebner:
I've tried it out a bit now.  Some comments:
 - The `LICENSE` file should go back into the root directory.
 - The `release.sh` script should also go back into the root directory, and run `npm run build` after `npm i`.  (Ideally, there'd be a github action which uploads the release when a tag is pushed, but we can do that another time.)
 - There's quite a bit of flickering when you move the cursor through a file (it switches between "No info found." and something else).  This might be related to https://leanprover.zulipchat.com/#narrow/stream/147302-lean4-dev/topic/Auto.20completion/near/233509746 and eafe2e1980467d3851f175754ffb56c27d4c332b
 - I'm not sure what I think about the grayed-out identifiers.  Graying out the hypothesis name `foo✝` in `foo✝ : Nat` is a positive change in my eyes (since you can easily spot the accessible hypotheses).  But graying them out in the rest of the term looks very distracting (since it's the only highlighting so it sticks out like a sore thumb).
 - The pinned info shows `Loading...` for a short time every time you type something.
 - Copy-to-comment produces an extra newline.
 - vscode now uses 230% CPU (with 4 processes)

Kha:
>     * But graying them out in the rest of the term looks very distracting (since it's the only highlighting so it sticks out like a sore thumb).

In Emacs I quickly got used to it, and I believe @leodemoura liked them as well. I haven't seen this PR in action yet though, it might be a case of color schemes. Here is what it looks like in Emacs for me currently: 
image

leodemoura:
> > ```
> > * But graying them out in the rest of the term looks very distracting (since it's the only highlighting so it sticks out like a sore thumb).
> > ```
> 
> In Emacs I quickly got used to it, and I believe @leodemoura liked them as well. I haven't seen this PR in action yet though, it might be a case of color schemes. Here is what it looks like in Emacs for me currently:
> image

I get something very similar on my machine, and yes I like it too :)

gebner:
It's a much lighter gray here:

dead_highlighting

With #777 as the gray color for the light theme it looks better for me.

Vtec234:
> The LICENSE file should go back into the root directory.
> vscode now uses 230% CPU (with 4 processes)

fixed

> The release.sh script should also go back into the root directory, and run npm run build after npm i. (Ideally, there'd be a github action which uploads the release when a tag is pushed, but we can do that another time.)

I made an attempt at how this should work but cannot test it without the PAT, so feel free to change things. Note the slightly weird `--scope=lean4` because the extension package is named `lean4`. Ideally we'd rename it to something like `@lean4/vscode-extension` but not necessary if it would cause trouble with the extension marketplace.

Vtec234:
> Copy-to-comment produces an extra newline

This is also the current behaviour, but I removed the newline.

> There's quite a bit of flickering when you move the cursor through a file

> The pinned info shows Loading... for a short time every time you type something.

I removed the `Loading...` indicator, I think it is nicer this way. In particular we still get a gold font on the status line when things are loading so no information is lost.

gebner:
> > Copy-to-comment produces an extra newline
> 
> This is also the current behaviour, but I removed the newline.

Oops, I didn't notice that.  You're right of course.

> > There's quite a bit of flickering when you move the cursor through a file
> 
> > The pinned info shows Loading... for a short time every time you type something.
> 
> I removed the `Loading...` indicator, I think it is nicer this way. In particular we still get a gold font on the status line when things are loading so no information is lost.

I think that was a different bug.  I mean the following:
```
#eval 1 + Nat.zero
```
1. Put the cursor on `Nat`, and pin the info.
2. Insert spaces before 1.
  - Every space increases the position of the pin by two!
  - On every keystroke, the expected type briefly disappears.
3. After some spaces, the pin is beyond the end of the line.

Vtec234:
I fixed the type disappearing. As for the double-move, I also observed that before but can't anymore. I will wishfully think it disappeared along with one of the recent changes, but let me know if it's still buggy. Also the pin movement is still incorrect if you insert a newline before `1` and then immediately remove it, but that also seems broken in the current code so is no regression.

gebner:
I still get the double-moves.

Vtec234:
Fixed now, hopefully. This was only observable on development builds because it was (intentionally, to catch this kind of bug) triggered by `React.StrictMode`.

Vtec234:
@gebner I've been test-driving this since the PR and it seems to work fine. I think it's good to go. Though I probably wouldn't release a new version until there are actually new features (WIP) using the LSP API.

gebner:
Great!  I'll make a release so that we get some free testing!

PR #32, opened 2021-07-31
feat: support server RPC refs

Vtec234:
Adds support for receiving and using [`RpcRef`s](https://github.com/leanprover/lean4/blob/32bea737089f5153cef65e480601e2e610973222/src/Lean/Data/Lsp/Extra.lean#L71) in the client but does not do anything with it yet. Some points:
- We have to bump the target to ES2021 to hook into the JS GC with `FinalizationRegistry`. I don't think this is an issue -- we only intend to support recent Electron webviews and potentially Webkit ones.
- The "RPC" name collision is a bit confusing since we now have webview<->editor RPC and webview<->Lean RPC. Any ideas?

### DIFF
--- a/lean4-infoview/src/infoview/rpcSessions.tsx
+++ b/lean4-infoview/src/infoview/rpcSessions.tsx
@@ -1,0 +1,195 @@
+import React from "react"
+import { DocumentUri, TextDocumentPositionParams } from "vscode-languageserver-protocol"
+import { RpcPtr } from "../lspTypes"
+import { EditorContext, RpcContext } from "./contexts"
+import { EditorConnection } from "./editorConnection"
+import { DocumentPosition, useServerNotificationEffect } from "./util"
+
+interface RpcConnectParams {
+    uri: DocumentUri
+}
+
+interface RpcConnected {
+    uri: DocumentUri
+    sessionId: string
+}
+
+interface RpcCallParams extends TextDocumentPositionParams {
+    sessionId: string
+    method: string
+    params: any
+}
+
+interface RpcReleaseParams {
+    uri: DocumentUri
+    sessionId: string
+    ref: RpcPtr<any>
+}
+
+const RpcNeedsReconnect = -32900
+
+class RpcSession {
+    #ec: EditorConnection
+    #uri: DocumentUri
+    #finalizers: FinalizationRegistry<RpcPtr<any>>
+
+    constructor(readonly sessionId: string, uri: DocumentUri, ec: EditorConnection) {
+        this.#ec = ec
+        this.#uri = uri
+        // Here we hook into the JS GC and send release-reference requests whenever
+        // the GC finalizes an `RpcPtr`. Requires ES2021.
+        this.#finalizers = new FinalizationRegistry(ptr => {
+            const params: RpcReleaseParams = {
+                uri: this.#uri,
+                sessionId: this.sessionId,
+                ref: ptr,
+            }
+
+            // TODO(WN): batch multiple of these into a single notif
+            void this.#ec.api.sendClientNotification('$/lean/rpc/release', params)
+        })
+    }
+
+    async call(pos: DocumentPosition, method: string, params: any): Promise<any> {
+        const rpcParams: RpcCallParams = {
+            ...DocumentPosition.toTdpp(pos),
+            sessionId: this.sessionId,
+            method: method,
+            params: params,
+        }
+        return this.#ec.api.sendClientRequest('$/lean/rpc/call', rpcParams)
+            .then(val => {
+                // const s = JSON.stringify(val)
+                // console.log(`'${method}(${JSON.stringify(params)})' at '${pos.line}:${pos.character}' -> '${s.length < 200 ? s : '(..)'}'`)
+                return val
+            })
+    }
+
+    registerRef(ptr: RpcPtr<any>) {
+        this.#finalizers.register(ptr, RpcPtr.copy(ptr))
+    }
+}
+
+/** Provides an interface for making RPC calls to the Lean server. */
+export class RpcSessions {
+    /** Maps each URI to the connected RPC session, if any, at the corresponding file worker. */
+    #connected: Map<DocumentUri, RpcSession>
+    /** Like `#connected`, but for sessions which are currently connecting. */
+    #connecting: Map<DocumentUri, Promise<RpcSession>>
+    #ec: EditorConnection
+    setSelf: (_: (_: RpcSessions) => RpcSessions) => void
+
+    constructor(ec: EditorConnection) {
+        this.#connected = new Map()
+        this.#connecting = new Map()
+        this.#ec = ec
+        this.setSelf = () => { }
+    }
+
+    private async sessionAt(uri: DocumentUri): Promise<RpcSession | undefined> {
+        if (this.#connected.has(uri)) return this.#connected.get(uri)!
+        else if (this.#connecting.has(uri)) return this.#connecting.get(uri)!
+        else return undefined
+    }
+
+    private connectAt(uri: DocumentUri): void {
+        // NOTE(WN): a request would be slightly simpler to handle than a notification,
+        // but would complicate the server more.
+
+        if (this.#connecting.has(uri)) {
+            throw `already connecting at '${uri}'`
+        }
+        if (this.#connected.has(uri)) {
+            this.#connected.delete(uri)
+        }
+        const newSesh = new Promise<RpcSession>(resolve => {
+            const h = this.#ec.events.gotServerNotification.on(([method, params]) => {
+                if (method !== '$/lean/rpc/connected') return
+                const conn: RpcConnected = params
+                if (conn.uri !== uri) return
+                resolve(new RpcSession(params.sessionId, conn.uri, this.#ec))
+                h.dispose() // this works, somehow
+            })
+        })
+        this.#connecting.set(uri, newSesh)
+        const connParams: RpcConnectParams = { uri }
+        this.#ec.api.sendClientNotification('$/lean/rpc/connect', connParams)
+        newSesh.then(newSesh => {
+            this.#connected.set(uri, newSesh)
+            this.#connecting.delete(uri)
+
+            // Trigger a React update when we connect.
+            this.setSelf(rs => {
+                // The `RpcSessions` object is intentionally global with only a shallow copy here
+                const newRs = new RpcSessions(rs.#ec)
+                newRs.#connected = rs.#connected
+                newRs.#connecting = rs.#connecting
+                newRs.setSelf = rs.setSelf
+                return newRs
+            })
+        })
+    }
+
+    /**
+     * Executes an RPC call in the context of `pos`. If the relevant RPC session is not yet
+     * connected, returns `undefined` and then triggers a React update. See also `registerRef`.
+     */
+    async call<T>(pos: DocumentPosition, method: string, params: any): Promise<T | undefined> {
+        const sesh = await this.sessionAt(pos.uri)
+        if (!sesh) {
+            this.connectAt(pos.uri)
+            return undefined
+        }
+
+        try {
+            const ret = await sesh.call(pos, method, params)
+            return ret
+        } catch (ex) {
+            if (ex.code === RpcNeedsReconnect) {
+                // Are we reconnecting yet?
+                if (!this.#connecting.has(pos.uri)) {
+                    this.connectAt(pos.uri)
+                }
+                return undefined
+            }
+            console.error(`RPC error: ${JSON.stringify(ex)}`)
+            throw ex
+        }
+    }
+
+    /**
+     * All {@link RpcPtr}s received from the server must be registered for garbage
+     * collection through this method. Not doing so is safe but will leak memory.
+     */
+    registerRef(pos: DocumentPosition, ptr: RpcPtr<any>): void {
+        void this.sessionAt(pos.uri).then(sesh => {
+            if (sesh) sesh.registerRef(ptr)
+            else throw `tried to register ref in non-existent RPC session ${pos.uri}`
+        })
+    }
+
+    /**
+     * Returns an identifier for the RPC session at `uri`. When this changes, all components
+     * which may use RPC in the context of that document must be re-created to avoid using
+     * outdated references.
+     */
+    sessionIdAt(uri: DocumentUri): string {
+        if (this.#connected.has(uri)) return this.#connected.get(uri)!.sessionId
+        return ''
+    }
+}
+
+/** Manages a Lean RPC connection by providing an {@link RpcContext} to the children. */
+export function WithRpcSessions({ children }: { children: React.ReactNode }) {
+    const ec = React.useContext(EditorContext)
+    const [sessions, setSessions] = React.useState<RpcSessions>(new RpcSessions(ec))
+    sessions.setSelf = setSessions
+
+    useServerNotificationEffect('$/lean/rpc/connected', _ => {
+        // Just subscribe here, but actually handle internally
+    }, [])
+
+    return <RpcContext.Provider value={sessions}>
+        {children}
+    </RpcContext.Provider>
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,5 +12,7 @@
     "eslint.workingDirectories": [
         "./lean4-infoview",
         "./vscode-lean4",
-    ]
+    ],
+    "files.insertFinalNewline": true,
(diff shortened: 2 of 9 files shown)


gebner:
Web browser support shouldn't be a problem: https://caniuse.com/mdn-javascript_builtins_finalizationregistry

lean4-infoview/src/infoview/rpcSessions.tsx
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    private async sessionAt(uri: DocumentUri): Promise<RpcSession | undefined> {
+        if (this.#connected.has(uri)) return this.#connected.get(uri)!
+        else if (this.#connecting.has(uri)) return this.#connecting.get(uri)!
+        else return undefined
+    }
+
+    private connectAt(uri: DocumentUri): void {
+        // NOTE(WN): a request would be slightly simpler to handle than a notification,
```
gebner, line 96: I think a request is the better choice here, if explicit initialization is necessary.  (And doesn't seem to be any harder to implement in the server, whether you call `handleRpcConnect` in `handleRequest` or `handleNotification` shouldn't make a big difference.)

gebner (commented):
The two RPC names are fine with me.

This is not finished yet, right?  I don't see any calls to registerRef for example.

To be honest, I'm quite lost here.  Do you have some document that explains the RPC protocol?  Some questions after reading this PR:
 - Why do you have to initialize an RPC session?
 - Can you have more than one RPC session per file?
 - Can you close an RPC session?

There seems to be a lot of "magic" in the `call` method.  It fails the first time it is called, but then initializes a session instead as a side effect.

What happens if the infoview is closed, are the references released then?

lean4-infoview/src/infoview/rpcSessions.tsx
Vtec234: The issue is that in the server we currently maintain that notifications are handled synchronously and can modify server state while requests run in thread-pools and cannot. We'd need a special case for this one request, or a bigger overhaul to allow synchronous requests in general. It's doable but I'd rather keep it as-is for now given that the notification works fine.

Vtec234:
Sorry to delay this response; there is quite a bit of broken stuff on the server side I am still fixing.

> This is not finished yet, right? I don't see any calls to registerRef for example.

I wanted to make a smaller PR with just the RPC session manager and then add widgets later.

> Do you have some document that explains the RPC protocol?

The design is still a bit up in the air so not really, though you can consult [`Lean.Data.Lsp.Extra`](https://github.com/leanprover/lean4/blob/3c2b576ffd489664f77125141c38829da4fc1bd9/src/Lean/Data/Lsp/Extra.lean#L81) for the messages.

> Can you have more than one RPC session per file?

No. A file worker always has exactly one open session.

> Why do you have to initialize an RPC session? Can you close an RPC session? There seems to be a lot of "magic" in the call method. It fails the first time it is called, but then initializes a session instead as a side effect.

The server initializes RPC eagerly. The client *connects* to an already initialized session. The point is to deal with restarts of both client and server (due to user action or a crash). The logic is as follows:
- The client only sends `connect` once. On receiving the first client `connect`, a file worker sends the current session ID. On receiving a subsequent `connect`, it assumes that the client has restarted, makes a new RPC session (dropping all old references), and then sends the new session ID.
- On connecting, the client gets back an RPC session ID. It must use this ID in all RPC requests. If the file worker restarts, on receiving a `call` request it will note the old session ID and complain with `rpcNeedReconnect`. This is what the magic part of `call` does -- if it sees this error, it reconnects lazily. It is very convenient but a little confusing. I can write this up in a comment once the design is finalized.
- So it will be have correctly and won't leak when either side restarts, but you're right that we don't handle closing since all of the above is in the infoview, but closing would need to go in the VSCode extension. If we have a `hasInfoview` configuration flag, we could use [`didChangeConfiguration`](https://microsoft.github.io/language-server-protocol/specification#workspace_didChangeConfiguration) to tell the client that the infoview has closed and it should release all references, as well as stop sending new ones in diagnostics to avoid the leak you noted [here](https://github.com/leanprover/lean4/pull/596#pullrequestreview-720945831).

Vtec234:
Another option to handle closing the infoview (as well as not opening it in the first place) suggested earlier by Dany would be to have a heartbeat notification from the client. The server will eagerly send refs in diagnostics, but if it doesn't receive any heartbeat for N seconds, it will release the refs and stop sending new ones. I'm not sure which variant will end up simpler.

gebner:
> I wanted to make a smaller PR with just the RPC session manager and then add widgets later.

I agree, it's good to split the widgets into a separate PR.  But this PR should have the complete RPC implementation (i.e., no memory leaks).

> though you can consult Lean.Data.Lsp.Extra for the messages. [...]

Thanks for the explanation!  I think I got a bit confused by the names; `connect` is more like `getSessionId`, and `sessionId` seems to be essentially `processId`.

> If we have a hasInfoview configuration flag, we could use didChangeConfiguration to tell the client that the infoview has closed and it should release all references

This seems really fragile and limiting.  What if we want to have more than one infoview?  The RPC mechanism is really cool.  What if we want to use it to implement other editor features that are not in the infoview (like better hovers, etc.)?

Another option to deal with the "releasing refs in diagnostics with closed infoview" issue would be to have an explicit `getInteractiveDiagnostics` method (that takes a range as argument and returns all diagnostics in that range, typically a single line).  This would also cut down on the data transferred by `publishDiagnostics`.  That notification is triggered often and is already a bottleneck right now if you have a large number of errors.  Eagerly sending the interactive version for every diagnostic more than duplicates the notification payload.

> The server will eagerly send refs in diagnostics, but if it doesn't receive any heartbeat for N seconds, it will release the refs and stop sending new ones.

If we don't want to modify the protocol, we could also have the vscode extension (i.e., `InfoProvider`) keep track of the received references and release them when the infoview is closed.

Vtec234:
> But this PR should have the complete RPC implementation (i.e., no memory leaks).

We should deal with the leaking diagnostics issue, however for RPC procedures, because the shape of incoming data and embedded `RpcPtr`s is variable, there is no general mechanism for registering them for garbage collection. Instead we need type-specific bindings which do this. On the Lean side there is a nice `deriving`, while in TS it is sadly hand-written for now (though I hope to autogenerate these eventually). For example
```typescript
export type TaggedText<T> = { text: string } | { append: TaggedText<T>[] } | { tag: [T, TaggedText<T>] }
function registerFormatRefs(rs: RpcSessions, pos: DocumentPosition, f: TaggedText<RpcPtr<any>>): void {
    const go = (f: TaggedText<RpcPtr<any>>) => registerFormatRefs(rs, pos, f)
    if ('append' in f) { for (const a of f.append) go(a) }
    else if ('tag' in f) { rs.registerRef(pos, f.tag[0]); go(f.tag[1]) }
}
```

Vtec234:
> an explicit `getInteractiveDiagnostics` method

This is doable and would also allow us to get rid of the diagnostics-passing hacks in the extension. We'd only need some caching on the server side to avoid formatting messages twice (once as interactive, once as not). However it would add some latency since the infoview will have to wait for the diagnostics to come back rather than render what's already there. The question is whether it'll be noticeable.

> If we don't want to modify the protocol, we could also have the vscode extension (i.e., InfoProvider) keep track of the received references and release them when the infoview is closed.

I'd rather add the heartbeat than an additional reference-tracking layer inbetween. It seems like it could reasonably reliably handle the closing issue, though it'd act weird in cases where the timer jumps -- say the user puts their laptop to sleep.

gebner:
> However it would add some latency since the infoview will have to wait for the diagnostics to come back rather than render what's already there.

Note that we already have exactly the same latency for the goal view, which doesn't seem to be a problem in practice.

> there is no general mechanism for registering them for garbage collection. Instead we need type-specific bindings which do this.

Ah, then I misunderstood your comment on the `RpcRef` structure.  I thought you wanted to do this generically: traverse the object, find all `p` properties, and just assume that they're references.  (I would've suggested a less common name than `p` though.)

The problem with type-specific bindings is that we cannot add new fields to existing structures, at least if they contain references.  Because otherwise older versions of the extension would leak memory.

> It seems like a heartbeat could reasonably reliably handle the closing issue

It only handles the closing issue if there's a single infoview and nothing else uses the RPC.  This assumption is true now, but fairly limiting for future developments.  What if somebody wants two different tabs for diagnostics and the goal view?  What if we want to use the RPC outside of the infoview?

Vtec234:
> It only handles the closing issue if there's a single infoview and nothing else uses the RPC.

We could have multiple per-session heartbeats, no?

Vtec234:
I implemented the session expiration mechanism. The client is expected to send a notification (at least) every 10s, and the server kills the connection after not seeing any notifications for 30s (three periods). It is super simple to implement and deals with any potential leaks -- if there is no client, the connection is guaranteed to terminate. The only issue now is that we can no longer push RPC refs to the client ahead of time, so I will add `getInteractiveDiagnostics`. However I will cache the interactive diags on the server so the processing time for this is just the roundtrip and not the pretty-print time. What do you think?

gebner:
> We could have multiple per-session heartbeats, no?

I'm not completely sure how that would work.  Would we have multiple "subsessions" where the references are scoped to a specific "subsession", and there is one heartbeat per "subsession"?  In that case we could just skip the heartbeats and have the infoview tell the extension about the "subsession", and the extension can then terminate the "subsession" when the infoview is closed.

> I implemented the session expiration mechanism. The client is expected to send a notification (at least) every 10s, and the server kills the connection after not seeing any notifications for 30s (three periods). It is super simple to implement and deals with any potential leaks -- if there is no client, the connection is guaranteed to terminate.

I am not happy about putting in a workaround for a vscode limitation into the RPC protocol.  Also 30 seconds is long enough that you can leak memory by closing and reopening the infoview.

> However I will cache the interactive diags on the server so the processing time for this is just the roundtrip and not the pretty-print time. What do you think?

I don't think it makes a difference when the server produces the interactive diagnostics.  The goal view is also not produced until it's requested and it doesn't seem to be an issue.  But AFAICT, this is just an implementation detail on the server and doesn't affect the API so we can always change it later on.

gebner:
> hook into the JS GC with FinalizationRegistry

BTW, have you measured how long it takes for the finalizer to be called?  I imagine that an object could survive quite long if it makes it through a few GC cycles and into an old generation.  I've tried to find any real-world measurements of how the FinalizationRegistry performs in practice, but there doesn't seem to be anything relevant.  There's a lot of scary warnings though, like this one from [v8](https://v8.dev/features/weak-references#progressive-enhancement):

> In general, avoid writing code that depends on the garbage collector cleaning up a WeakRef or calling a finalizer at any predictable time — it can’t be done! Moreover, whether an object is garbage-collectible at all may depend on implementation details, such as the representation of closures, that are both subtle and may differ across JavaScript engines and even between different versions of the same engine.

Vtec234:
> Would we have multiple "subsessions" where the references are scoped to a specific "subsession", and there is one heartbeat per "subsession"?

Pretty much, though I call these "sessions". References are already scoped to a session.

>  In that case we could just skip the heartbeats and have the infoview tell the extension about the "subsession", and the extension can then terminate the "subsession" when the infoview is closed.

This will not work when the extension crashes (which in a remote setup will not necessarily kill the remote server).

> I am not happy about putting in a workaround for a vscode limitation into the RPC protocol. Also 30 seconds is long enough that you can leak memory by closing and reopening the infoview.

Sorry, I'm not quite getting this. Which vscode limitation are you referring to? There is no (or at least there shouldn't be a) situation in which this approach leaks memory. Guarenteeing this even if the client/infoview crashes/misbehaves/.. is the whole point. Currently, since we only allow one session, the story at the end [here](https://github.com/leanprover-community/vscode-lean4/pull/32#issuecomment-893119875) still applies. If we implement support for multiple sessions in the future, this will happen:
- On startup the infoview sends a `connect` request which creates RPC session 1 (really a random id but no matter) on the server. While open, the infoview will keep session 1 alive with the heartbeat.
- When you reopen the infoview, the new webview no longer has any knowledge of session 1. It will send a `connect` again and get session 2. On the server, session 1 will expire after 30s due to a lack of heartbeat.

Vtec234:
> BTW, have you measured how long it takes for the finalizer to be called? I imagine that an object could survive quite long if it makes it through a few GC cycles and into an old generation

This will vary a lot per JS engine. In coarse terms, in current VSCode Insiders I observed more or less this:
- when you move off a goal state once, the refs for its types will generally stay alive (no GC pass yet)
- when you move the cursor about a few more times, the memory pressure will increase and the previous refs will be GC'd

> There's a lot of scary warnings though

Yes, the warning is not to rely on finalizers being called for correctness. But we are only using the GC to collect our (Lean-side) garbage, which is fine. This does make the memory usage non-deterministic, but I see no better option. I *really* don't want to RC manually in JS. In the future we may have Lean compiling to JS and emitting its RC instructions instead, but that will not happen now.

Vtec234:
> Ah, then I misunderstood your comment on the RpcRef structure. I thought you wanted to do this generically: traverse the object, find all p properties, and just assume that they're references. (I would've suggested a less common name than p though.) The problem with type-specific bindings is that we cannot add new fields to existing structures, at least if they contain references. Because otherwise older versions of the extension would leak memory.

Forgot to address this, sorry. I originally considered and dismissed the duck-typing approach as a bit hacky, but my opinion here is not strong. We could potentially do it (with a name unlikely to clash, as you write). I do think though that given the fast development pace of Lean 4, at the moment we should try to match client and server versions rather than worry about backwards compatibility. It will be an important consideration eventually, but I don't think that time is now.

gebner:
> Pretty much, though I call these "sessions". References are already scoped to a session.

I don't care about the name.  I tried to avoid talking about sessions for clarity, because these are currently in 1-to-1 correspondence with worker processes.

> This will not work when the extension crashes (which in a remote setup will not necessarily kill the remote server).

Is this a realistic concern? (I've never had the extension crash, ever.)

To be fair, I don't have a good idea of the various remote plugins/versions of vscode (there's a docker/ssh plugin, a live share plugin, the code-server version, etc.).  I thought the main extension code always runs on the same machine as the lean server, and only the infoview is shipped remotely?

> Sorry, I'm not quite getting this. Which vscode limitation are you referring to?

That the infoview cannot clean up after itself when it's closed.  If the infoview got a shutdown event or was part of the main editor process, we wouldn't have this problem.

> When you reopen the infoview, the new webview no longer has any knowledge of session 1. It will send a connect again and get session 2. On the server, session 1 will expire after 30s due to a lack of heartbeat.

Thanks for the clarification!  I didn't realize that the keep-alive token contained the session id, and that restarting the infoview created a new session.

> but I see no better option. I really don't want to RC manually in JS.

One realistic alternative/additional safeguard is to have finer-grained sessions.  For example, open a new session every time we load a new goal or new diagnostics (so you'd typically have two or three open sessions at a time).  With such a setup we might even get away without releasing any references at all (since the sessions are closed quickly).

> we should try to match client and server versions rather than worry about backwards compatibility. It will be an important consideration eventually, but I don't think that time is now.

At least forward compatibility becomes important the moment the RPC protocol is released.  Users want to open files from older Lean versions.  Matching client and server versions is not really an option.  (Though breaking compatibility after a year or two is okay in my eyes.)

Vtec234:
> Is this a realistic concern? (I've never had the extension crash, ever.)

Honestly probably not, but the nice thing about keep-alive is that we no longer have to worry about whether this or any other weird failure may happen. This will also hold for other editors.

> If the infoview got a shutdown event or was part of the main editor process, we wouldn't have this problem.

That's fair enough. However since keep-alive is editor-independent, we will not have to worry about shutdown events in this context in any Lean 4 extension.

> One realistic alternative/additional safeguard is to have finer-grained sessions

This would be very complicated. We would have to wrangle around the specific sessions in request handlers and have the lifetime management problem again for each of them. I would really rather not do this.

> Users want to open files from older Lean versions. Matching client and server versions is not really an option

While this is definitely true in Lean 3, the part I am not so sure is that it currently holds for Lean 4. It is in rapid development and the only larger project, Mathlib 4, pretty much tracks master. However I agree this will be an issue in the future. Regarding matching client/server, I am still hoping to bundle widget code together and have the server *push* the client JS as a dynamically loadable React component. This would guarantee a match, though I admit the idea is currently vaporware.

Vtec234:
> I tried to avoid talking about sessions for clarity, because these are currently in 1-to-1 correspondence with worker processes.

Btw, a session is *not* just tied to a worker process. This is perhaps the core semantics that I failed to convey. It represents a connection between a client and a worker. A new session is made when the client restarts, so it isn't 1-to-1 with workers. The fact that a worker always has one session state is an implementation detail.

PR #37, opened 2021-09-19
Info-on-hover widgets

Vtec234:
This is the client-side implementation of the Lean 3 feature which allows hovering over terms in the infoview to see their types and additional information such as docstrings. It is based on the server RPC [feature](https://github.com/leanprover/lean4/pull/596).

The client is still quite buggy, so help with testing is needed. To try it out, either clone the branch and build it, or download the VSIX CI artifact and install it in VSCode. Please do report any bugs or strange behaviour that you find.


gebner:
I'll look at this when I'm back from vacation in October.
1. This needs to be backwards-compatible with old Lean 4 versions.
2. The PR seems to include code for interactive traces which are not in Lean 4 yet afaict.

lean4-infoview/src/infoview/rpcInterface.ts
```diff
@@ -0,0 +1,147 @@
...
+    else if ('lazyTrace' in e) rs.registerRef(pos, e.lazyTrace[2])
+}
+
+function TaggedMsg_registerRefs(rs: RpcSessions, pos: DocumentPosition, t: TaggedText<MsgEmbed>): void {
+    const go = (t: TaggedText<MsgEmbed>) => {
+        if ('append' in t) { for (const a of t.append) go(a) }
+        else if ('tag' in t) { MsgEmbed_registerRefs(rs, pos, t.tag[0]); go(t.tag[1]) }
+    }
+    go(t)
+}
```
gebner, line 124: These `registerRefs` functions are 100% boilerplate and are a chore to keep in sync with upstream (and a single typo causes a memory leak).  I think we should just look for `p` attributes and register those.  Then we don't need handcrafted register functions or complicated wrappers (since `RpcSession.call` can register the objects by itself).  I believe you wanted to do it this way originally.  We're also using this approach in the neovim extension and it works great: https://github.com/Julian/lean.nvim/blob/53bd0cc9d5dd27c46f3c1372dc331da4c5372244/lua/lean/rpc.lua#L108-L125

API-wise I think it would be good to include the position in the `RpcContext`.  Right now there's a lot of pointless `pos` arguments everywhere.

vscode-lean4/src/leanclient.ts
```diff
@@ -74,7 +74,7 @@ export class LeanClient implements Disposable {
 
     async restart(): Promise<void> {
         if (this.isStarted()) {
-            await this.stop()
+            void this.stop()
```
gebner, line 77: This looks like a really dangerous idea (if stop was actually async).  I think you should just remove the async modifier on stop instead.

vscode-lean4/src/leanclient.ts
```diff
@@ -253,7 +247,7 @@ export class LeanClient implements Disposable {
 
     async stop(): Promise<void> {
         assert(() => this.isStarted())
-        await this.client.stop()
+        void this.client.stop()
```
gebner, line 250: I don't know what the vscode languageclient does when stopping.  I hope it doesn't do any cleanup, otherwise this is a race condition with the change in start.

lean4-infoview/src/infoview/main.tsx
```diff
@@ -81,22 +80,22 @@ export function renderInfoview(editorApi: EditorApi, uiElement: HTMLElement): In
...
+            editorEvents.sentClientNotification.fire([method, params])
         },
         changedCursorLocation: async loc => editorEvents.changedCursorLocation.fire(loc),
         changedInfoviewConfig: async conf => editorEvents.changedInfoviewConfig.fire(conf),
         requestedAction: async action => editorEvents.requestedAction.fire(action),
-    };
+    }
 
-    const ec = new EditorConnection(editorApi, editorEvents);
+    const ec = new EditorConnection(editorApi, editorEvents)
```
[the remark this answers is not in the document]
Vtec234: Yep sorry, this was before I suggested Prettier-ifying everything.

lean4-infoview/src/infoview/interactiveCode.tsx
```diff
@@ -0,0 +1,181 @@
...
+/** Shows `explicitValue : itsType` and a docstring if there is one. */
+function TypePopupContents({pos, info, redrawTooltip}: {pos: DocumentPosition, info: InfoWithCtx, redrawTooltip: () => void}) {
+  const rs = React.useContext(RpcContext)
+  const [ip, setIp] = React.useState<InfoPopup>()
+
+  React.useEffect(() => {
+    InteractiveDiagnostics_infoToInteractive(rs, pos, info).then(val => {
+      if (val) {
+        setIp(val)
+        // We let Tippy.js know that the tooltip should be re-rendered,
```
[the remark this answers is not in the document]
Vtec234: Yes, it's because tippy has code to auto-show and auto-hide popups on hover and keep them open when the user is hovering over the popup itself, while popper only renders the popups. Unfortunately tippy isn't quite customizable enough to do exactly what I want for the UI so there is some extra code doing that, but doing it all ourselves would still be quite a bit more complicated.

vscode-lean4/src/leanclient.ts
Vtec234: I don't remember exactly why I made this change, but it had something to do with startup delay. I think I wanted to rapidly restart the server for development purposes. However we don't expect anyone to have to do this, so I may well revert the change. Thoughts?

lean4-infoview/src/infoview/interactiveCode.tsx
```diff
@@ -0,0 +1,181 @@
...
+  return <Tippy {...computedProps} ref={ref} />
+})
+
+/** Shows `explicitValue : itsType` and a docstring if there is one. */
+function TypePopupContents({pos, info, redrawTooltip}: {pos: DocumentPosition, info: InfoWithCtx, redrawTooltip: () => void}) {
+  const rs = React.useContext(RpcContext)
+  const [ip, setIp] = React.useState<InfoPopup>()
+
+  React.useEffect(() => {
+    InteractiveDiagnostics_infoToInteractive(rs, pos, info).then(val => {
```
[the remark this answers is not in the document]
Vtec234: Sorry, `async` where? `React.useEffect` does not accept async callbacks.

lean4-infoview/src/infoview/rpcInterface.ts
Vtec234: No, the intention was always to do it explicitly this way. I was (and still am) hoping to auto-generate the code from Lean using a JS backend (see comment at top of file). But there are more important things to work on atm so, although I have reasons to believe it would be significantly easier than an LLVM backend, say, I admit a Lean JS emitter is vaporware. It would also increase the implementation cost for Lua.

Then as a pragmatic matter, I'll agree we can use an ad-hoc traversal; as you mentioned before, `p` is too likely to clash -- if we go with this approach, we should change it to something like `_rpcptr`. This will unfortunately increase payload size.

vscode-lean4/src/leanclient.ts
gebner: If the only reason for the change was to make development easier, then I'd prefer to keep the `await` version.  Spawning async tasks without waiting for them feels like a race condition bound to happen.

lean4-infoview/src/infoview/interactiveCode.tsx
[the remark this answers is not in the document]
gebner: Okay, web development is messy after all.

lean4-infoview/src/infoview/interactiveCode.tsx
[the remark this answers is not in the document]
gebner: Uggh.  I just saw the `.then` and assumed this could be easily refactored into an `await`.  And `(async () => { ... })()` would be even worse here.

lean4-infoview/src/infoviewApi.ts
```diff
@@ -96,9 +96,11 @@ export interface InfoviewApi {
   /**
    * Must fire whenever the user moves their cursor or makes a selection.
    * When no selection is made, `loc.range.start` must equal `loc.range.end`.
+   * Must fire with `undefined` when the cursor moves to "nowhere" (e.g. all
+   * files are closed).
```
gebner, line 100: Should this also specify what happens with multiple cursors?

vscode-lean4/src/rpc.ts
```diff
@@ -39,8 +39,15 @@ export class Rpc {
...
                     this.sendMessage({ seqNum, result: await this.methods[name](...args) })
                 } catch (ex) {
-                    // TODO(WN): somehow `ex.message` is lost here on going through `webview.postMessage`
-                    this.sendMessage({ seqNum, exception: ex === undefined ? 'error' : ex })
+                    if (ex === undefined) {
+                        this.sendMessage({ seqNum, exception: 'error' })
+                        return
+                    }
+                    /* Somehow certain properties (such as `ex.message`) disappear along the way
+                     * through `Webview.postMessage` but make it through in a new object. */
```
gebner, line 47: This is interesting.  I wanted to suggest using `{...ex}`, but look what happens:
```
> {...new Error('my message')}
{}
```

After studying the ECMA specification, I believe this because the `message` attribute is not *enumerable*:
```
> Object.getOwnPropertyDescriptor(new Error('my message'), 'message')
{
  value: 'my message',
  writable: true,
  enumerable: false,
  configurable: true
}
```

gebner:
Bug: sometimes opening the tooltip inserts a newline.  For example if you open the tooltip on `id` here:
```lean
example : id 2 = 1 + 1 := rfl
                        --^ open infoview here
```

lovettchris:
@Vtec234 I'd like to start working on this PR, I've merged it with master here: https://github.com/lovettchris/vscode-lean4/tree/rpc-widgets and I can keep working on it there if you like....

Vtec234:
Should be good now :) I tested it to work with pre-widgets nightly.

PR #39, opened 2021-09-28
Fix dot bug

SteveUlin:
There are some Black Very Small Squares “⬝” (U+2B1D) in this file instead of Middle Dots “·” (U+00B7).

Only middle dots work with the simple lambda notation, i.e. ( .  + 1).


alexjbest:
I wouldn't say this is a bug, the other dot is a different symbol which has different uses (e.g. in mathlib 3 anyway). If you hover over the symbol in vscode it should tell you how to type it.

SteveUlin:
Fair fair, I don't want to break mathlib.

Currently I get this message when I hover over the error in ( \dot + 1)  
"""
expected tokenLean 4
Type ⬝ using \tr or \dot or \con or \cdot
"""

SteveUlin:
Hovering over a working solution tells the user how to type the correct unicode.

PR #44, opened 2021-09-30
fix: infoview stays closed if user closes it

lovettchris:
fix: The infoview no longer reopens when the Lean server is restarted, and it no longer opens for markdown files.


vscode-lean4/src/infoview.ts
```diff
@@ -248,6 +254,7 @@ export class InfoProvider implements Disposable {
 
     private async sendPosition() {
         if (!window.activeTextEditor || !languages.match(this.leanDocs, window.activeTextEditor.document)) { return null; }
+        void this.autoOpen();
```
Vtec234, line 257: Why add this here?

lovettchris: if user changes this preference then next time they move the position in the editor the infoview will popup.

image

Vtec234: Ah, then maybe it should go into the `workspace.onDidChangeConfiguration` hook? If that ends up starting the infoview too soon (i.e. while the user is still changing settings) then let's keep it in `sendPosition` but maybe rename it to `onPositionChanged` or some such since it now does other things besides just sending the position.

lovettchris: Yeah, exactly, that felt too soon, and it crowds out the settings editor...

gebner: > it crowds out the settings editor...

If you're in the settings editor, then the infoview wouldn't open (afaict, since it's the active editor and not a lean file).

It took me a while to understand what the actual issue here was.  The PR fixes two things:
 - The infoview reopens when the Lean server is restarted,
 - and it opens for markdown files.

Please just note this in the PR description / commit message.  We should also update the description of the setting to reflect the current behavior.

I think it would be clearer to add the `autoOpen` handler to the `onDidChangeActiveTextEditor` handler.

vscode-lean4/src/infoview.ts
```diff
@@ -19,6 +19,7 @@ export class InfoProvider implements Disposable {
     private subscriptions: Disposable[] = [];
 
     private stylesheet: string = '';
+    private autoOpened: boolean;
```
gebner, line 22:
```suggestion
    private autoOpened: boolean = false;
```

vscode-lean4/src/infoview.ts
lovettchris: Ok, I tried moving it to onDidChangeActiveTextEditor , but it is happening too soon (before LeanClient calls languages.setTextDocumentLanguage(doc, 'lean4')), so I added a new didSetLanguage event to the client which the InfoView listens to.  Now the following scenario works:
1. open vs code on lean package with setting AutoOpen turned off
2. open a .lean source file (note: no InfoView), but lean server is started.
3. close the .lean source file
4. Change the AutoOpen setting to Checked.
5. Open a .lean source file (InfoView pops up when it realizes this is a lean4 document).

Vtec234 (approved):

vscode-lean4/src/leanclient.ts
```diff
@@ -227,7 +230,8 @@ export class LeanClient implements Disposable {
         if (doc.languageId === 'lean') {
             // Only change the id for *visible* documents,
             // because this closes and then reopens the document.
-            await languages.setTextDocumentLanguage(doc, 'lean4')
+            await languages.setTextDocumentLanguage(doc, 'lean4');
+            this.didSetLanguageEmitter.fire(undefined);
```
Vtec234, line 234: Oh I just noticed this, since it's `EventEmitter<string>` maybe `fire` with `'lean4'` here? Or change the type to `EventEmitter<undefined>`.

lovettchris: Oh right, 'lean4' makes perfect sense here. Fixed.

PR #50, opened 2021-10-13
restart lean server when workspace folder changes.

lovettchris:
work in progress... open to feedback...


gebner:
This PR doesn't implement what the title says.  The server is restarted when a file in a different workspace folder is opened.

So when you open a file workspace folder A, then a file from workspace folder B, then one from A, etc., the server will be restarted every time.  And every time in the same directory, so it doesn't make much of a difference.

It might be useful to restart the server when the workspace folders change (or actually the rootPath¹).  Or maybe just show a warning popup, since restarting the server takes a lot of time and CPU and we might not want to do that to the user without asking first.

There is also the more general question of what *should* happen when you open multiple workspace folders in vscode.  Should we start the server in the first folder and use it for everything.  Should we start one server per folder?  (The directory the server is started in has an effect on the Lean version that is used, and on the dependencies that can be imported, so it's important.)  I don't use multiple workspace folders myself, so I can't really say what the expected behavior here is.

¹ We're using the vscode-languageserver-node library to start the server, and I'm not completely sure where it starts the server when there are multiple workspace folders.

Issue #47, opened 2021-10-08
Documentation view with "Try This" to run code snippets locally.


lovettchris:
## Documentation view

The book "Theorem Proving in Lean" can be read by hitting ctrl+shift+p and searching for "Lean: Open Documentation View".

Any Lean project with an html/ folder can be displayed in this panel. For example, try out the book Mathematics in Lean.

![image](https://user-images.githubusercontent.com/18707114/136627914-fb9d7a62-1b64-4307-8f4d-ff71edcfb41b.png)

PR #63, opened 2021-10-30
Add docView and `Lean4 Show All Abbreviations` that displays the current list of abbreviations 

lovettchris:
Add "Lean 4: Show All Abbreviations" command that writes to the docview and from there you can use the handy vscode search feature.

image


Vtec234:
I am very positive about the feature of displaying all abbreviations (sometimes I have to look them up in the .json file!), however I think a "fake" `\help` abbreviation is quite unexpected/non-discoverable UI. I would propose the following alternative.
- We add a "Lean 4: Show all abbreviations" command to the command palette.
- We point the user towards this command in the "Type .. using .." popup. For example, on hovering over `←` the user would see a popup "Type ← using `\l` or `\leftarrow` or `\l-` or `\gets` or `\<-`. Click here to show all abbreviations." with the "here" being a link that executed the `Lean 4: Show all abbreviations` command if such a thing is possible. If not, we could instead write "Type ← using `\l` or `\leftarrow` or `\l-` or `\gets` or `\<-`. See also `Lean 4: Show all abbreviations`".

What do you think?

gebner:
100% agreed with Wojciech.  I would expect `\help` to input `🚑` (or maybe some other character), not perform an action and open something.

I wouldn't have put the table in the infoview either (in my mind, the infoview is for information about the current file, not for documentation).  I'd prefer to put the table in a separate tab (i.e., another, separate webview).  But I don't feel too strongly on this.

Vtec234:
I also agree regarding putting documentation in its own webview. It might fit quite naturally into the docview (#47).

Kha:
Emacs actually shows the entire list as well, but it does so after `\<TAB>` - and analogously when completing any other ambiguous prefix. 
image

Vtec234:
@Kha fwiw, I _hate_ this feature in Emacs because every time I mispress Tab while trying to complete an abbreviation despite there being none to complete, that list will appear and I then have to go close it.

lovettchris:
thanks for all the great feedback, I will play with all these ideas and see which one I like best, but I agree \help is not discoverable or expected.

lean4-infoview/src/infoview/main.tsx
```diff
@@ -21,6 +21,14 @@ function Main(props: {}) {
     if (!props) { return null }
     const ec = React.useContext(EditorContext);
 
+    const [htmlBlob, setHtmlBlob] = React.useState<string>('')
```
gebner, line 24: The changes to the infoview are no longer necessary now.

vscode-lean4/src/abbreviation/AbbreviationProvider.ts
```diff
@@ -11,6 +11,9 @@ export class AbbreviationProvider implements Disposable {
 	private readonly disposables = new Array<Disposable>();
 	private cache: Record<string, string | undefined> = {};
 
+    private abbreviationCompletedEmitter = new EventEmitter<TextEditor>();
```
gebner, line 14: This event is unused now.

vscode-lean4/src/abbreviation/AbbreviationProvider.ts
```diff
@@ -31,6 +38,19 @@ export class AbbreviationProvider implements Disposable {
 		};
 	}
 
+	getAbbreviations() : Map<string, string> {
```
gebner, line 41: Please just make `symbolsByAbbreviation` public instead.

gebner (commented):
Styling-wise: I think single column was the right decision.  I'm not sure if it's a good idea to increase the font size, the list is long and it's probably more useful to see more abbreviations per page (you can always ctrl-= to increase the font size).  It might also make sense to use the same (monospace) font that is also used in the editor.  I would also print `\alpha` instead of `alpha` (i.e. include with the leader character).

vscode-lean4/src/abbreviation/AbbreviationProvider.ts
lovettchris: True, but this is twice now that I've found it might be useful to have so I might just leave it for a bit and see what else pops up that needs this event.

vscode-lean4/src/abbreviation/AbbreviationProvider.ts
lovettchris: Done.

lean4-infoview/src/infoview/main.tsx
lovettchris: Fixed

lovettchris:
> Styling-wise: I think single column was the right decision. I'm not sure if it's a good idea to increase the font size, the list is long and it's probably more useful to see more abbreviations per page (you can always ctrl-= to increase the font size). It might also make sense to use the same (monospace) font that is also used in the editor. I would also print `\alpha` instead of `alpha` (i.e. include with the leader character).

Fixed.

vscode-lean4/src/abbreviation/AbbreviationProvider.ts
gebner: We can remove the `getAbbreviations` and `getAbbreviationsNames` functions.

gebner (commented):
LGTM otherwise.

vscode-lean4/src/abbreviation/AbbreviationProvider.ts
lovettchris: Woops, right, removed.

PR #64, opened 2021-10-30
fix: print server stderr messages once

Vtec234:


vscode-lean4/src/leanclient.ts
```diff
@@ -240,17 +240,8 @@ export class LeanClient implements Disposable {
         } as any, null);
 
         // HACK
-        (this.client as any)._serverProcess.stderr.on('data', (msg) => {
-            const s = msg.toString();
-            // ignore this normal message that happens on server restart.
-            if (s.trim() !== 'Watchdog error: Cannot read LSP message: Stream was closed'){
-                this.client.outputChannel.appendLine(s);
```
Vtec234, line 247: This didn't do what it was supposed to since the `LanguageClient` already appends all `stderr` data to the output channel. So the only effect is messages which are not `Watchdog error: ..` would be printed twice instead of once. Instead, I filed leanprover/lean4#754 stop the server from printing the troublesome errors.

vscode-lean4/src/leanclient.ts
```diff
@@ -240,17 +240,8 @@ export class LeanClient implements Disposable {
         } as any, null);
 
         // HACK
```
gebner, line 242:
```suggestion
        // Reveal the standard error output channel when the server prints something to stderr.
        // The vscode-languageclient library already takes care of writing it to the output channel.
```
Let's improve this comment here so that the next person knows what's up.

Issue #58, opened 2021-10-20
InfoView stops updating afer "Lean 4: Resstart Server" command


lovettchris:
Steps:
1. install vscode lean4 extension v0.0.39
1. load a lean4 folder
1. check everything works fine, including the infoview.
1. run `Lean 4: Resstart Server" command`
1. now move caret around the .lean file and check the infoview still works.

Someones you see this output:
```
Watchdog error: cannot find open document 'file:///d%3A/Temp/lean_examples/Foo/Foo.lean'
[Error - 11:57:39 AM] Connection to server got closed. Server will not be restarted.
```
and the infoview stops updating.

I'm wondering if this needs to be re-initialized after client restart since the initializeResult will be different after server restart...
```
await webviewPanel.api.initialize(this.client.initializeResult, this.getLocation(editor))
```

Vtec234:
Looks like this has been fixed in our recent changes. If anyone is still observing the issue, please reopen.

lovettchris:
Agreed

PR #66, opened 2021-11-01
Close RPC sessions on server restart.

### LINKED ISSUE #52: Use the new lineRange parameter of the interactive diagnostics command.
This would fix this issue that's still present in the vscode extension: https://github.com/Julian/lean.nvim/pull/165#issuecomment-939107439
```lean
import Lean

open Lean Elab in
elab "slow" : command => do
  IO.sleep (ms := 1000)
  logInfo "I should be able to see this message in the infoview before the whole file is done."

slow
slow
slow
slow
slow
slow
slow
slow
slow
slow
slow
slow
slow
slow
slow
slow
slow
```

gebner:
Fixes #52.


lovettchris:
Cool, does this fix issue #58 ?

vscode-lean4/src/infoview.ts
```diff
@@ -136,6 +136,7 @@ export class InfoProvider implements Disposable {
                 if (this.webviewPanel && this.client.initializeResult && window.activeTextEditor){
                     await this.webviewPanel.api.initialize(this.client.initializeResult, this.getLocation(window.activeTextEditor));
                 }
+                void this.webviewPanel?.api?.serverRestarted()
```
lovettchris, line 139: Why isn't `webviewPanel.api.initialize` enough?  With this branch I'm seeing the inital startup case is broken where infoview starts working then this even comes in and effectively kills it, leaving the infoview in a non-working state.

lovettchris (changes requested):

vscode-lean4/src/infoview.ts
lovettchris: This is what I mean by infoview not working, notice the numbers are not updating as I edit the .lean file:

https://user-images.githubusercontent.com/18707114/139720209-a52e5617-1ad8-4fda-b94b-eb44db11abe9.mp4

Vtec234:
I believe this is intended to fix a different issue, namely the one where RPC sessions are not closed on server restart -- see [Zulip](https://leanprover.zulipchat.com/#narrow/stream/147302-lean4-dev/topic/vscode.3A.20select.20interpreter/near/259853444). However this is not the right fix. The issue seems to be that:
- either the infoview is not reinitialized, i.e. somehow `clear()` in `infoview.ts` doesn't work; or
- the infoview is reinitialized correctly, and an RPC session for `B.lean` starts but `B` is not opened in the server (?!); or
- something else is going on.

lovettchris:
> I believe this is intended to fix a different issue, namely the one where RPC sessions are not closed on server restart -- see [Zulip](https://leanprover.zulipchat.com/#narrow/stream/147302-lean4-dev/topic/vscode.3A.20select.20interpreter/near/259853444). However this is not the right fix. The issue seems to be that:
> 
> * either the infoview is not reinitialized, i.e. somehow `clear()` in `infoview.ts` doesn't work; or
> * the infoview is reinitialized correctly, and an RPC session for `B.lean` starts but `B` is not opened in the server (?!); or
> * something else is going on.

I put some console.logging in RpcSession constructor then replayed the scenario of opening 2 files, restarting the server then switching the files and the log output shows the expected RpcSessions are created and each rpc call is passing the correct document Uri:
```
    child process exited with code 0
    client starting
    client running
[opening first file named Mymath.lean]
    Creating new RpcSession if 12972191147399796795 for document    
        file:///d%3A/Temp/lean_examples/MyMath/MyMath.lean
[opening second file named util.lean]
    Creating new RpcSession if 2205546381138363094 for document         
        file:///d%3A/Temp/lean_examples/MyMath/MyMath/util.lean
[restarting the server]
    client has stopped or it failed to start
    client starting
    client running
    Creating new RpcSession if 12904032966990033144 for document         
        file:///d%3A/Temp/lean_examples/MyMath/MyMath/util.lean
[select second file]
    rejected promise not handled within 1 second: Error [ERR_STREAM_WRITE_AFTER_END]: write after end
    Creating new RpcSession if 18268304082738162278 for document         
        file:///d%3A/Temp/lean_examples/MyMath/MyMath.lean
```

Vtec234:
A print statement [here](https://github.com/leanprover/vscode-lean4/blob/38a8c3fe7def7fc4ca02474a0138b29776284293/vscode-lean4/webview/index.ts#L14) is not triggered for me on server restart, so indeed it looks like `clear()` does not work. Notably the documentation for `Webview.html` states: `This should be a complete, valid html document. Changing this property causes the webview to be reloaded.` but it does not look like it is actually reloaded. @lovettchris is it working for you?  It looks like a VSCode bug, likely dependent on the version.

vscode-lean4/src/infoview.ts
gebner: Ah, this is probably a race condition.  It works just fine here on my machine.

gebner:
From what I can tell the `clear()` indeed doesn't work.  However if `clear()` worked, then the infoview would lose config, diagnostics, progress, position information when the server is restarted.

lovettchris:
I found and fixed a related bug here: https://github.com/leanprover/vscode-lean4/pull/69

Vtec234:
> From what I can tell the `clear()` indeed doesn't work. However if `clear()` worked, then the infoview would lose config, diagnostics, progress, position information when the server is restarted.

We should send the diagnostics, progress and position on `clear` in the same way we already do on infoview startup. I am in favor of this solution because it is low effort - it will allow the infoview to assume that unless it had received a `didClose`, file workers stay open (although they may still restart). An alternative would be to send a fake `didClose` for all previously open files other than the active editors (the only ones which VSCode sends a `didOpen` for on restart, thus creating new workers), or the `serverRestarted` event you proposed. But again, I am in favour of just nuking and reiniting the infoview as on server restart there cannot possibly be state we may persist in the infoview anyway. A similar approach works reasonably well for the file workers in the server.

lovettchris (changes requested):

lean4-infoview/src/infoview/rpcSessions.tsx
```diff
@@ -223,10 +228,14 @@ export class RpcSessions implements Disposable {
         })
     }
 
-    dispose() {
+    ensureAllSessionsClosed() {
         this.#connected.forEach(rs => rs.dispose())
         this.#connecting.forEach(fut => fut.then(rs => rs.dispose()))
```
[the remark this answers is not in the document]
gebner: Well spotted.

gebner:
> I think the `void this.webviewPanel?.api?.serverRestarted()` event should be fired before calling `this.webviewPanel.api.initialize`

There were a few issues related to the message protocol that got in the way here.  For example the `initialize` call is only supposed to be called a single time, otherwise we tell react to mount the infoview component multiple times (which apparently replaces the old one but I'm not sure if it cleans up all the event handlers).  I think part of the issue was also that the WithDiagnosticsContext only updates on the publishDiagnostics notification, creating a race condition with the serverRestarted event.  The infoview also requested interactive diagnostics for other files (in addition to the current one).

I have now refactored the code a bit so that the diagnostics/progress contexts are always created, even if there is no server connected.  This prevents race conditions that are caused by react's asynchronous updating.

Since I was already refactoring the diagnostics context, I've also implemented #52.

> there cannot possibly be state we may persist in the infoview anyway.

I think we want to persist all UI state, open details, pins, tactic state filters, etc.  The intention is that you can restart the server frequently and easily in Lean 4.

lean4-infoview/src/infoview/rpcSessions.tsx
```diff
@@ -268,6 +279,8 @@ export function WithRpcSessions({ children }: { children: React.ReactNode }) {
         []
     )
 
+    useEvent(ec.events.serverRestarted, () => sessions.ensureAllSessionsClosed())
```
Vtec234, line 282: Instead of adding even more complexity to `RpcSessions`, why don't we leave it as-is, and instead call `dispose()`, then make a `new RpcSessions(ec)` here?

gebner: That's not a bad idea!  But I believe the closed flag will still be necessary to prevent references from the old session to be released.

lovettchris:
> I think we want to persist all UI state, open details, pins, tactic state filters, etc. The intention is that you can restart the server frequently and easily in Lean 4.

Thanks your new bits work really well, switching Lean server versions is now very smooth, and the bug I reported before is fixed, the new edits in the .lean file are being picked up and new diagnostics are being shown correctly.

vscode-lean4/src/infoview.ts
lovettchris: Infoview now works with your latest changes, so I'm resolving this conversation.

lean4-infoview/src/infoview/info.tsx
```diff
@@ -5,7 +5,7 @@ import { Goals as GoalsUi, Goal as GoalUi, goalsToString } from './goals';
 import { basename, DocumentPosition, RangeHelpers, useEvent, usePausableState } from './util';
 import { Details } from './collapsing';
 import { EditorContext, ProgressContext, RpcContext, VersionContext } from './contexts';
-import { MessagesAtFile, useMessagesFor } from './messages';
+import { MessagesAtFile, MessagesList, useMessagesFor } from './messages';
```
lovettchris, line 8: MessagesAtFile is  no longer used.

lean4-infoview/src/infoview/main.tsx
```diff
@@ -91,49 +98,36 @@ export function renderInfoview(editorApi: EditorApi, uiElement: HTMLElement): In
...
-            )
-        }
-    })
+    editorEvents.initialize.on((loc: Location) => ec.events.changedCursorLocation.fire(loc))
+
+    ReactDOM.render(<React.StrictMode>
+        <EditorContext.Provider value={ec}>
+            <Main/>
+        </EditorContext.Provider>
+    </React.StrictMode>, uiElement)
```
lovettchris, line 130: nice, much simpler!

vscode-lean4/src/infoview.ts
```diff
@@ -228,17 +232,8 @@ export class InfoProvider implements Disposable {
             webviewPanel.webview.html = this.initialHtml();
 
             // this.client.initializeResult can be null if the server failed to start.
-            if (this.client.initializeResult) {
-                await webviewPanel.api.initialize(this.client.initializeResult, this.getLocation(editor))
```
lovettchris, line 232: Much nicer, separating out the this.client.initializeResult until the separate serverRestarted event makes a lot of sense.

lovettchris (approved):

lean4-infoview/src/infoviewApi.ts
```diff
@@ -96,6 +96,9 @@ export interface InfoviewApi {
   /** Like {@link gotServerNotification}, but fires on subscribed client->server notifications. */
   sentClientNotification(method: string, params: any): Promise<void>;
 
+  /** Fires when the server is restarted. */
```
Vtec234, line 99:
```suggestion
  /** Must fire with the server's initialization message when the server is started or restarted. */
```
```suggestion
  /** Fires when the server is restarted. */
```

vscode-lean4/src/infoview.ts
```diff
@@ -228,17 +232,8 @@ export class InfoProvider implements Disposable {
             webviewPanel.webview.html = this.initialHtml();
 
             // this.client.initializeResult can be null if the server failed to start.
```
Vtec234, line 234: Outdated comment

vscode-lean4/src/infoview.ts
```diff
@@ -128,16 +128,20 @@ export class InfoProvider implements Disposable {
     constructor(private client: LeanClient, private readonly leanDocs: DocumentSelector, private context: ExtensionContext) {
         this.updateStylesheet();
         this.subscriptions.push(
-            this.client.restarting(async () => {
-                this.clear();
```
Vtec234, line 132: The `clear()` method can also be removed.

lean4-infoview/src/infoview/rpcSessions.tsx
Vtec234: Ah yes, unfortunately. But we should still make a new `RpcSessions` object to make sure the promise chain in `connectAt` does not make modifications to `this` after a server restart (I don't think this can happen, but then again I didn't think the bugs we've seen would happen either). In that case `dispose` does not need to clear out `#connected/#connecting`, just has to clear the timeouts and set `#closed` so the `FinalizationRegistry` stops sending `release`s.

lean4-infoview/src/infoview/contexts.ts
```diff
@@ -1,18 +1,17 @@
...
 import { InfoviewConfig, defaultInfoviewConfig } from "../infoviewApi";
 import { RpcSessions } from "./rpcSessions";
-import { InteractiveDiagnostic } from "./rpcInterface";
 import { ServerVersion } from "./serverVersion";
 
 // Type-unsafe initializers for contexts which we immediately set up at the top-level.
 export const EditorContext = React.createContext<EditorConnection>(null as any);
 export const RpcContext = React.createContext<RpcSessions>(null as any);
-export const VersionContext = React.createContext<ServerVersion>(null as any);
+export const VersionContext = React.createContext<ServerVersion | undefined>(null as any);
```
Vtec234, line 13:
```suggestion
export const VersionContext = React.createContext<ServerVersion | undefined>(undefined);
```
```suggestion
export const VersionContext = React.createContext<ServerVersion | undefined>(null as any);
```

lean4-infoview/src/infoview/info.tsx
```diff
@@ -245,7 +245,7 @@ function InfoAux(props: InfoProps) {
         setStatus('updating');
 
         let allReq = undefined
-        if (sv.hasWidgetsV1()) {
+        if (sv?.hasWidgetsV1()) {
```
Vtec234, line 248: Very minor, but I would use `sv!` here keeping in mind the invariant that the message UI is not rendered if `sv === undefined`.

lean4-infoview/src/infoview/main.tsx
```diff
@@ -52,11 +52,16 @@ function Main(props: {}) {
         []
     );
 
+    const serverInitializeResult = useEventResult(ec.events.serverRestarted);
```
Vtec234, line 55: If `useEventResult` were higher-order and taking a function that maps the event value, we could also use it for `uri`/`events.changedCursorLocation` above.

lean4-infoview/src/infoview/messages.tsx
```diff
@@ -64,13 +64,8 @@ function mkMessageViewProps(uri: DocumentUri, messages: InteractiveDiagnostic[])
     return addUniqueKeys(views, v => DocumentPosition.toString({uri: v.uri, ...v.diag.range.start}));
 }
 
-export interface MessagesAtFileProps {
-    uri: DocumentUri;
-    messages: InteractiveDiagnostic[];
-}
-
 /** Shows the given messages for the given file. */
```
Vtec234, line 67: Comment is no longer true?

lean4-infoview/src/infoview/messages.tsx
```diff
@@ -109,7 +113,7 @@ export function AllMessages({uri: uri0}: { uri: DocumentUri }) {
                 </a>
             </span>
         </summary>
-        <MessagesAtFile uri={uri} messages={diags} />
+        <MessagesAtFile uri={uri} />
```
Vtec234, line 116: This change breaks the pausing logic. Note that `diags` is frozen when the UI is paused, but with this change we ignore the frozen value. Since the pausable state is in this class rather than `MessagesAtFile`, one option may be to move `useMessagesForFile` up into this class and get rid of `MessagesAtFile`.

lean4-infoview/src/infoview/messages.tsx
```diff
@@ -118,41 +122,42 @@ export function AllMessages({uri: uri0}: { uri: DocumentUri }) {
...
+            try {
+                const diags = await getInteractiveDiagnostics(rs, { uri, line: 0, character: 0 },
+                    line ? { start: line, end: line + 1 } : undefined)
+                if (diags) {
+                    setDiags(diags)
+                }
+            } catch (err: any) {
+                console.log('getInteractiveDiagnostics error ', err)
+            }
+        }
```
Vtec234, line 154: This breaks backwards compatibility with pre-line-filterable-diags servers, no? Since we no longer make a request if `!hasWidgetsV1()`.

lean4-infoview/src/infoview/messages.tsx
```diff
@@ -118,41 +122,42 @@ export function AllMessages({uri: uri0}: { uri: DocumentUri }) {
  * Provides a `DiagnosticsContext` by subscribing to server diagnostic notifications
  * and querying for interactive diagnostics whenever (LSP standard) diagnostics arrive.
```
Vtec234, line 123: Outdated comment.

lean4-infoview/src/infoview/main.tsx
gebner: Good suggestion!  But there's no higher-order argument necessary, you can just do a `?.uri` afterwards.

gebner: Added the higher-order version anyhow to reduce rerenders.

lean4-infoview/src/infoview/rpcSessions.tsx
```diff
@@ -268,6 +273,11 @@ export function WithRpcSessions({ children }: { children: React.ReactNode }) {
         []
     )
 
+    useEvent(ec.events.serverRestarted, () => {
+        sessions.dispose();
+        setSessions(new RpcSessions(ec));
```
lovettchris, line 278: How does this deal with the possibility that there could be some pending async calls to `sendClientRequest` or `await sesh.call` that could complete after the new RpcSessions has been created, then you could have competing async requests, etc...  I'd feel better if RpcSessions had a this.disposed flag that each async completion checked and bailed if it is set to true.

gebner: Good point.  A lot of this code assumes that the `#connected` and `#connecting` maps stay the same (even though the RpcSessions object is recreated every time a session is created).  I don't want to refactor this now.

lean4-infoview/src/infoview/messages.tsx
gebner: This is ugly.. we don't want to move `useMessagesForFile` into this class because this will request all interactive diagnostics for the file (which blocks and causes a lot of unnecessary RPC messages).

gebner: Okay, put in a moderately ugly solution using a thunk for the interactive diagnostics.

gebner:
I've noticed something else that's a bit disturbing.  We use a lot of these React context objects to capture notifications sent from the server.  However this causes serious performance issues.  It's obvious once you think of it:
```jsx
<ProgressContext.Provider value={allProgress}>
  {ret}
</ProgressContext.Provider>
```
Whenever `allProgress` changes, everything in `ret` is being rerendered.  At first I accidentally had `<WithRpcSessions>` inside of the `ProgressContext.Provider`, which apparently causes the RPC session to be reset every time the server updated the progress...

We need to replace all these providers by `Event`s.

Vtec234:
> Whenever `allProgress` changes, everything in `ret` is being rerendered. At first I accidentally had `<WithRpcSessions>` inside of the `ProgressContext.Provider`, which apparently causes the RPC session to be reset every time the server updated the progress...

Isn't the whole point of contexts that only children which `useContext` are rerendered when the context changes? `WithRpcSessions` does not use the progress context, so it shouldn't matter. Could it be that you're only testing in debug mode where React does more updates than necessary to test correctness?

gebner:
> Isn't the whole point of contexts that only children which useContext are rerendered when the context changes?

I don't know, I haven't found a precise description of when react will rerender the components or mount/unmount them.  AFAICT by default `<A value={x}><Foo/></A>` will at least rerender `Foo` and all children when `x` changes (unless you explicitly opt out using memoization).  Whether `A` is a provider or not does not make a difference here.  It also doesn't matter whether it is a debug build or not.

> Could it be that you're only testing in debug mode where React does more updates than necessary to test correctness?

Yes, and this is also recommended by react.  But this doesn't mean we can ignore this.  If the extension is sluggish to the point of being unusable in a development build, then it becomes hard to work on it.

Vtec234:
> by default `<A value={x}><Foo/></A>` will at least rerender `Foo` and all children when `x` changes

I can confirm this happens in development mode. It is unfortunate, conceptually I would expect React to keep track of which children actually `useContext` and only render those, however..

> At first I accidentally had <WithRpcSessions> inside of the ProgressContext.Provider, which apparently causes the RPC session to be reset every time the server updated the progress...

I am more worried about this. This sounds like a bug in our logic more than anything about React. The `WithRpcSessions` component should not be re-created every time it is re-rendered, in particular the `RpcSessions` state should be the same object across its lifetime. Is this still reproducible in the before-last commit?

gebner:
> > by default `<A value={x}><Foo/></A>` will at least rerender `Foo` and all children when `x` changes
> 
> I can confirm this happens in development mode. It is unfortunate, conceptually I would expect React to keep track of which children actually `useContext` and only render those, however..

That's not what React does, and it's well documented.  And to reiterate, it has nothing to do with useContext.  It's also independent of development/production build.  The dev build only enables some additional warnings, it doesn't change the whole semantics.

> > At first I accidentally had  inside of the ProgressContext.Provider, which apparently causes the RPC session to be reset every time the server updated the progress...
> 
> I am more worried about this. This sounds like a bug in our logic more than anything about React. The `WithRpcSessions` component should not be re-created every time it is re-rendered, in particular the `RpcSessions` state should be the same object across its lifetime. Is this still reproducible in the before-last commit?

Oh sorry, false alarm!  I tried my standard stress test: https://gist.github.com/gebner/523230aa3f95a1f40b92aee998494da6

What happened is that I first hovered around in an interactive diagnostic, and this flickers a lot.  I was afraid that this happened because I inadvertently swapped the order of the providers so I tried swapping them back.  Then I hovered in the expected type and everything was fine.

(Of course what actually happened (before and after the swapping) is that the interactive diagnostics are reloaded every time a publishDiagnostics notification arrives, but the expected type is only updated when you move the cursor or the line is still being processed.)

Vtec234:
> That's not what React does, and it's well documented. And to reiterate, it has nothing to do with useContext. It's also independent of development/production build. The dev build only enables some additional warnings, it doesn't change the whole semantics.

I see. Per the docs, "Even if an ancestor uses `React.memo` or `shouldComponentUpdate`, a rerender will still happen starting at the component itself using `useContext`.". So our options are to memoize more components, or implement what I thought React should be doing ourselves with `Event`s (as you suggest).

lean4-infoview/src/infoview/rpcSessions.tsx
```diff
@@ -244,9 +249,15 @@ export class RpcSessions implements Disposable {
         })
     }
 
-    dispose() {
+    ensureAllSessionsClosed() {
         this.#connected.forEach(rs => rs.dispose())
         this.#connecting.forEach(fut => fut.then(rs => rs?.dispose()))
```
Vtec234, line 254: If we are not recreating the `RpcSessions` object, the `then` should include the same lambda as in `ensureSessionClosed`. More simply, `this.#connected.forEach((_, uri) => this.ensureSessionClosed(uri)); this.#connecting.forEach((_, uri) => this.ensureSessionClosed(uri))` should work.

lean4-infoview/src/infoview/messages.tsx
```diff
@@ -81,15 +76,36 @@ export function MessagesAtFile({uri, messages}: MessagesAtFileProps) {
...
+            } catch (err: any) {
+                console.log('getInteractiveDiagnostics error ', err)
+            }
+        }
+        return diags0.map(d => ({ ...(d as LeanDiagnostic), message: { text: d.message } }));
+    }), [sv, rs, uri0, diags0]);
 
-    const [isPaused, setPaused, [uri, diags], _] = usePausableState(false, [uri0, diags0]);
+    const [isPaused, setPaused, [uri, diags, iDiags], _] = usePausableState(false, [uri0, diags0, iDiags0]);
+    if (isPaused) void iDiags()
```
Vtec234, line 108: Is this fetching diagnostics when paused? Should it not do so on `!isPaused`?

lean4-infoview/src/infoview/messages.tsx
```diff
@@ -81,15 +76,36 @@ export function MessagesAtFile({uri, messages}: MessagesAtFileProps) {
...
-    const dc = React.useContext(DiagnosticsContext);
+    const sv = React.useContext(VersionContext);
+    const rs = React.useContext(RpcContext);
+    const dc = React.useContext(LspDiagnosticsContext);
     const config = React.useContext(ConfigContext);
-    let diags0 = dc.get(uri0);
-    if (!diags0) diags0 = [];
+    const diags0 = dc.get(uri0) || [];
+
+    let iDiags0 = React.useMemo(() => lazy(async () => {
```
Vtec234, line 96: I'm not sure this is the right way to do this. Per per React docs: "Remember that the function passed to `useMemo` runs during rendering. Don’t do anything there that you wouldn’t normally do while rendering. For example, side effects belong in `useEffect`, not `useMemo`.".

lean4-infoview/src/infoview/messages.tsx
gebner: Yes, `iDiags` is a thunk and the code above forces it when the component switches from `!isPaused` to `isPaused`.  (It will also force it on every subsequent update when the component is paused, but those will be no-ops.)

The idea is that we don't need to load the diagnostics when the "all messages" view is closed and not paused.  But when it switches to paused, we need to fetch them.

lean4-infoview/src/infoview/messages.tsx
gebner: The memo is fine here.  The `lazy` function is side effect-free, it only creates the thunk but does not execute it.

But maybe we should use useEffect for the iDiags() call below..

Vtec234 (approved):
Thanks!! My takeaway is that we should rethink:
- the state machine/internal logic in `RpcSessions`
- the use of React contexts; `Event`s used correctly may trigger fewer re-renders

lovettchris:
> Thanks!! My takeaway is that we should rethink:
> 
> * the state machine/internal logic in `RpcSessions`
> * the use of React contexts; `Event`s used correctly may trigger fewer re-renders

Is this the right way to see a render count?  When I 'run without debugging" I'm seeing 71 renders during startup, and 36 renders each time I type a character in the .lean file....which does seem a tad excessive... 
```tsx
    const rendersCountRef = React.useRef(0);
    rendersCountRef.current += 1;

    return (
    <ConfigContext.Provider value={config}>
        <VersionContext.Provider value={sv}>
            <WithRpcSessions>
                <WithLspDiagnosticsContext>
                    <ProgressContext.Provider value={allProgress}>
                        {ret}
                    </ProgressContext.Provider>
                </WithLspDiagnosticsContext>
            </WithRpcSessions>
        </VersionContext.Provider>
        <div>
        RendersCounter rendered <b>{rendersCountRef.current}</b>
        </div>
    </ConfigContext.Provider>
```

lovettchris:
By the way, there is a ton of amazing discussion in this PR that I would have loved to capture in a Zulip discussion thread, I think it is time we created a vscode specific channel for these discussions in Zulip and I would love to continue this discussion even after this PR is done.  I've been analyzing the React dependencies and it would be fun to continue this kind of discussion there to help get more shared understanding and to develop a roadmap towards simplification/optimization of this stuff...
image

lean4-infoview/src/infoview/rpcSessions.tsx
```diff
@@ -244,9 +249,13 @@ export class RpcSessions implements Disposable {
...
+        [...this.#connected.keys()].forEach(uri => this.ensureSessionClosed(uri));
+        [...this.#connecting.keys()].forEach(uri => this.ensureSessionClosed(uri));
+    }
+
     dispose() {
-        this.#connected.forEach(rs => rs.dispose())
-        this.#connecting.forEach(fut => fut.then(rs => rs?.dispose()))
+        this.ensureAllSessionsClosed();
     }
 }
```
lovettchris, line 261: I don't understand line 256 the `React.useEffect` containing `return () => sessions.dispose()` can you explain this one to me?

lovettchris:
I think I found a bug:

1. open a vscode folder and open a .lean file and ensure info view is visible
2. close info view 
3. type `Ctrl`+`Shift`+`Enter` to reopen info view
4. Infoview says `Waiting for Lean server to start...` forever...

lean4-infoview/src/infoview/rpcSessions.tsx
lovettchris: Never mind I read up on `useEffect` and see now that the empty dependencies `[]` cause this to be a `componentDidMount` event only and it is returning the function to be invoked on unmount, which is the sessions.dispose...

Vtec234:
> Is this the right way to see a render count? When I 'run without debugging" I'm seeing 71 renders during startup, and 36 renders each time I type a character in the .lean file....which does seem a tad excessive...

I think it is right, and I agree it sounds excessive to call all our render functions this many times. Although it really depends, e.g. if only the top-level component is rendered 36 times but the expensive-to-compute children are memoised and not re-rendered, then it might be fine -- we should probably `useMemo` more components in any case. But ideally even without memoisation there would be fewer redraws.

gebner:
> Is this the right way to see a render count? When I 'run without debugging" I'm seeing 71 renders during startup, and 36 renders each time I type a character in the .lean file....which does seem a tad excessive...

Yes.  The numbers don't surprise me. I think this means that there are around 15 commands (def/...) below the cursor.  The server sends 2 notifications for every processed command (publishDiagnostics and fileProgress) and we do a rerender for each one of them.

lean4-infoview/src/infoview/rpcSessions.tsx
gebner: Indeed, this cleanup handler will only be executed if we tell react to display something else.  It would be great if we could cleanly shutdown the RPC sessions when the infoview is closed, but I don't know if this is possible.  We are currently using a keepalive timer to clean up inactive sessions on the server side for this reason.

gebner:
I'm not sure if this is a regression in this PR, but if I leave vscode running in the background for a few minutes (one minute is not enough), then I get the following message (in the infoview hover) when I come back:
> Error: Cannot decode params in RPC call 'Lean.Widget.InteractiveDiagnostics.infoToInteractive({"p":"1"})'
RPC reference '1' is not valid

lovettchris:
> (I could reproduce this on master, so I've filed it as #73.)

I've seen this also.

PR #69, opened 2021-11-03
This fixes a leak in cleaning up RpcSessions objects

lovettchris:
I think this explains some of the debug output we see after editing lean files for a while.

### DIFF
--- a/lean4-infoview/src/infoview/rpcSessions.tsx
+++ b/lean4-infoview/src/infoview/rpcSessions.tsx
@@ -180,7 +180,8 @@ export class RpcSessions implements Disposable {
                 if (!this.#connecting.has(uri)) {
                     // force a reconnect.
                     if (this.#connected.has(uri)) {
-                       this.#connected.delete(uri);
+                        this.#connected.get(uri)?.dispose()
+                        this.#connected.delete(uri);
                     }
                     this.connectAt(uri);
                 }
@@ -219,7 +220,7 @@ export class RpcSessions implements Disposable {
         this.#connected.delete(uri)
         this.#connecting.get(uri)?.then(sesh => {
             sesh.dispose()
-            this.#connected.delete(uri)
+            this.#connecting.delete(uri)
         })
     }
 
--- a/lean4-infoview/src/infoview/rpcSessions.tsx
+++ b/lean4-infoview/src/infoview/rpcSessions.tsx
@@ -117,7 +117,7 @@ export class RpcSessions implements Disposable {
     /** Maps each URI to the connected RPC session, if any, at the corresponding file worker. */
     #connected: Map<DocumentUri, RpcSession>
     /** Like `#connected`, but for sessions which are currently connecting. */
-    #connecting: Map<DocumentUri, Promise<RpcSession>>
+    #connecting: Map<DocumentUri, Promise<RpcSession | undefined>>
     #ec: EditorConnection
     setSelf: (_: (_: RpcSessions) => RpcSessions) => void
 
@@ -138,12 +138,17 @@ export class RpcSessions implements Disposable {
         if (this.#connecting.has(uri) || this.#connected.has(uri)) {
             throw new Error(`already connecting or connected at '${uri}'`)
         }
-        this.ensureSessionClosed(uri)
         const connParams: RpcConnectParams = { uri }
-        const newSesh: Promise<RpcSession> = this.#ec.api.sendClientRequest('$/lean/rpc/connect', connParams)
+        let newSesh: Promise<RpcSession | undefined> = this.#ec.api.sendClientRequest('$/lean/rpc/connect', connParams)
             .then((conn: RpcConnected) => new RpcSession(conn.sessionId, uri, this.#ec))
+            .catch(() => {
+                this.#connecting.delete(uri)
+                return undefined
+            })
         this.#connecting.set(uri, newSesh)
-        newSesh.then(newSesh => {
+        newSesh = newSesh.then((newSesh: RpcSession | undefined) => {
+            if (!newSesh) return undefined
+
             this.#connected.set(uri, newSesh)
             this.#connecting.delete(uri)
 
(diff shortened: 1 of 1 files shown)


gebner:
This PR also includes the commits from #63.  Please rebase onto master.

lovettchris:
> This PR also includes the commits from #63. Please rebase onto master.

fixed

lean4-infoview/src/infoview/rpcSessions.tsx
```diff
@@ -180,7 +180,8 @@ export class RpcSessions implements Disposable {
                 if (!this.#connecting.has(uri)) {
                     // force a reconnect.
                     if (this.#connected.has(uri)) {
-                       this.#connected.delete(uri);
+                        this.#connected.get(uri)?.dispose()
```
Vtec234, line 183: Thanks! This was definitely an oversight.

lean4-infoview/src/infoview/rpcSessions.tsx
```diff
@@ -219,7 +220,7 @@ export class RpcSessions implements Disposable {
         this.#connected.delete(uri)
         this.#connecting.get(uri)?.then(sesh => {
             sesh.dispose()
-            this.#connected.delete(uri)
+            this.#connecting.delete(uri)
```
Vtec234, line 223: I think this line, on the other hand, was correct. Note that when a new promise is inserted into `#connecting`, it already includes code that will remove it from `#connecting` when the promise resolves, but will add it to `#connected`. Hence the `#connected.delete` here. I spent quite a while thinking about this design and couldn't find a way to simplify it, unfortunately :/

lean4-infoview/src/infoview/rpcSessions.tsx
lovettchris: yep, my fault I think... since I've been running with this fix I have not seen the pesky output errors...

lean4-infoview/src/infoview/rpcSessions.tsx
lovettchris: Ok, I reverted it, but I'm still worried that sessionAt actually looks at what is in the this.#connecting, or are you confident that the connectAt newSesh.then() handler will always fire and cleanup the this.#connecting list.  What happens in the case that line 143 `this.#ec.api.sendClientRequest` fails?  Who cleans up the this.#connecting list in that case?

Vtec234: I am not, and the request can in fact fail if the file is not open. I started working on a fix for this, which I think I can push directly to this branch.

lovettchris: ok, if you have a fix that would be great.  This branch is in my own fork because I don't think I have access to make branches directly on this repo...

Vtec234: I pushed something that should make the session manager more reliable in normal usage. On server restart we will still see the `keepAlive`s because we're not yet restarting the infoview.

Vtec234 (approved):

lean4-infoview/src/infoview/rpcSessions.tsx
```diff
@@ -156,7 +161,16 @@ export class RpcSessions implements Disposable {
...
             })
+
+            return newSesh
         })
+        // NOTE(WN): because we insert into `#connecting` above, this conditional may only fail
+        // if the whole `newSesh` promise chain is executed immediately (if that's even possible).
+        // In that case we are either already connected or know we failed to do so, so don't want
+        // to mark as connecting.
+        if (this.#connecting.has(uri)) {
+            this.#connecting.set(uri, newSesh)
```
lovettchris, line 172: Does the comment match the code?  You say "don't want to make as connecting" but then you go right ahead and set #connecting?

lean4-infoview/src/infoview/rpcSessions.tsx
lovettchris: Thanks I like your improvements and tested them they seem to work fine.  I'm not sure about line 197 `this.connectAt(uri);` inside the call catch block because there are times when switching lean server versions where I see this catch block get hit 4 or 5 times in rapid succession.  Perhaps it should wait for the next call to run connectAt.  Also if I just let the editor sit there for a few minutes while I'm doing nothing, I see a series of RpcNeedsReconnect calls.  If it did not automatically reconnect it wouldn't do this and the "idle" would be more "idle":
image

lean4-infoview/src/infoview/rpcSessions.tsx
gebner: "that case" is the else-branch afaict (where it is indeed marked as not connecting).

gebner (commented):
The whole connected/connecting bookkeeping seems unnecessarily complicated (and error-prone as this PR shows).  We could probably get away with a single `#sessions` map and a `connected: Promise<{}>` field in `RpcSession`.

Vtec234:
> We could probably get away with a single `#sessions` map and a `connected: Promise<{}>` field in `RpcSession`.

I did try this idea when initially implementing `RpcSessions` and couldn't get it to work for reasons I don't remember, but which I think had to do with specifics of `async/await` in JS. I do agree that the `#connecting` invariants are overly complicated.

lean4-infoview/src/infoview/rpcSessions.tsx
Vtec234: > there are times when switching lean server versions where I see this catch block get hit 4 or 5 times in rapid succession. Perhaps it should wait for the next call to run `connectAt`

Note that the first call to `connectAt` will insert the `uri` into `#connecting`, so on further `RpcNeedsReconnect` exceptions we'll just return `undefined` but will not try to connect again.

Vtec234: > Also if I just let the editor sit there for a few minutes while I'm doing nothing, I see a series of `RpcNeedsReconnect` calls. If it did not automatically reconnect it wouldn't do this and the "idle" would be more "idle":

The client should always try to reconnect if it gets the `RpcNeedsReconnect` error from the server, but the server should not send this unless it has crashed or restarted. So seeing this when idling is a bug, but one that we should fix by ensuring the session stays alive rather than removing the reconnection code.

lovettchris: Ok, and I'm seeing that this RpcNeedsReconnect  does not keep happening in your #66

PR #70, opened 2021-11-04
fix: subscription counting and client running check

Vtec234:
An alternative approach to #66 which restarts the whole infoview.

### DIFF
--- a/vscode-lean4/src/infoview.ts
+++ b/vscode-lean4/src/infoview.ts
@@ -21,7 +21,7 @@ export class InfoProvider implements Disposable {
     private stylesheet: string = '';
     private autoOpened: boolean = false;
 
-    // Subscriptions are counted and only disposed when count === 0.
+    // Subscriptions are counted and only disposed of when count becomes 0.
     private serverNotifSubscriptions: Map<string, [number, Disposable]> = new Map();
     private clientNotifSubscriptions: Map<string, [number, Disposable]> = new Map();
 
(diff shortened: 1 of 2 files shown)


vscode-lean4/src/infoview.ts
```diff
@@ -225,26 +219,26 @@ export class InfoProvider implements Disposable {
...
+     * is already running. Do nothing if there is no webview panel. */
+    private async initWebview(editor: TextEditor) {
+        if (!this.webviewPanel) return
 
+        // this.client.initializeResult can be null if the server failed to start.
+        // in this case the infoview displays an error message.
+        await this.webviewPanel.api.initialize(this.client.initializeResult, this.getLocation(editor))
         // The infoview listens for server notifications such as diagnostics passively,
         // so when it is first started we must re-send those as if the server did.
-        await this.sendPosition();
```
Vtec234, line 238: I am not sure why we were re-sending this on revealing the infoview given that even when hidden the infoview should still receive position updates etc, but marking this in case I am missing something.

gebner:
As I've mentioned on #66, I don't think it's a good idea to discard the whole infoview state just because the server needs to be restarted. I think we want to persist all UI state, open details, pins, tactic state filters, etc. The intention is that you can restart the server frequently and easily in Lean 4.

Vtec234:
I see, let's go with that then. I will merge the two misc fixes here.

PR #74, opened 2021-11-08
Send RPC keep-alive from extension instead of infoview.

### LINKED ISSUE #73: RPC session expires after a few minutes
If I leave vscode running in the background for a few minutes (one minute is not enough), then I get the following message (in the infoview hover) when I come back and try to hover over a term in the infoview:

> Error: Cannot decode params in RPC call 'Lean.Widget.InteractiveDiagnostics.infoToInteractive({"p":"1"})'
RPC reference '1' is not valid

gebner:
Fixes #73.

It also allows us to explicitly close sessions in the future instead of / in addition to the keep-alive system.

### DIFF
--- a/lean4-infoview/src/infoview/rpcSessions.tsx
+++ b/lean4-infoview/src/infoview/rpcSessions.tsx
@@ -9,14 +9,6 @@ import { EditorContext, RpcContext } from "./contexts"
 import { EditorConnection } from "./editorConnection"
 import { DocumentPosition, useClientNotificationEffect, useEvent } from "./util"
 
-interface RpcConnectParams {
-    uri: DocumentUri
-}
-
-interface RpcConnected {
-    sessionId: string
-}
-
 interface RpcCallParams extends TextDocumentPositionParams {
     sessionId: string
     method: string
@@ -29,13 +21,6 @@ interface RpcReleaseParams {
     refs: RpcPtr<any>[]
 }
 
-interface RpcKeepAliveParams {
-    uri: DocumentUri
-    sessionId: string
-}
-
-const keepAlivePeriodMs = 10000
-
 const RpcNeedsReconnect = -32900
 
 class RpcSession implements Disposable {
@@ -43,7 +28,6 @@ class RpcSession implements Disposable {
     #uri: DocumentUri
     /** A timeout mechanism for notifying the server about batches of GC'd refs in fewer messages. */
     #releaseTimeout?: number
-    #keepAliveInterval: number
     #refsToRelease: RpcPtr<any>[]
     #finalizers: FinalizationRegistry<RpcPtr<any>>
     #closed: boolean = false
@@ -51,15 +35,6 @@ class RpcSession implements Disposable {
     constructor(readonly sessionId: string, uri: DocumentUri, ec: EditorConnection) {
         this.#ec = ec
         this.#uri = uri
-
-        // We setup a recurring timer to keep the RPC session alive.
-        this.#keepAliveInterval = window.setInterval(() => {
-            const params: RpcKeepAliveParams = {
-                uri: this.#uri,
-                sessionId: this.sessionId,
-            }
-            void this.#ec.api.sendClientNotification('$/lean/rpc/keepAlive', params)
-        }, keepAlivePeriodMs)
 
         // Here we hook into the JS GC and send release-reference notifications
         // whenever the GC finalizes a number of `RpcPtr`s. Requires ES2021.
@@ -113,7 +88,7 @@ class RpcSession implements Disposable {
 
     dispose() {
         this.#closed = true;
-        window.clearInterval(this.#keepAliveInterval)
+        this.#ec.api.closeRpcSession(this.sessionId)
     }
 }
 
@@ -143,9 +118,8 @@ export class RpcSessions implements Disposable {
         if (this.#connecting.has(uri) || this.#connected.has(uri)) {
             throw new Error(`already connecting or connected at '${uri}'`)
         }
-        const connParams: RpcConnectParams = { uri }
-        let newSesh: Promise<RpcSession | undefined> = this.#ec.api.sendClientRequest('$/lean/rpc/connect', connParams)
-            .then((conn: RpcConnected) => new RpcSession(conn.sessionId, uri, this.#ec))
+        let newSesh: Promise<RpcSession | undefined> = this.#ec.api.createRpcSession(uri)
+            .then(sessionId => new RpcSession(sessionId, uri, this.#ec))
             .catch(() => {
                 this.#connecting.delete(uri)
                 return undefined
--- a/lean4-infoview/src/infoview/rpcSessions.tsx
+++ b/lean4-infoview/src/infoview/rpcSessions.tsx
@@ -4,24 +4,10 @@
  */
 import React from "react"
 import { DidCloseTextDocumentParams, Disposable, DocumentUri, TextDocumentPositionParams } from "vscode-languageserver-protocol"
-import { RpcPtr } from "../lspTypes"
+import { RpcPtr, RpcCallParams, RpcNeedsReconnect, RpcReleaseParams } from "../lspTypes"
 import { EditorContext, RpcContext } from "./contexts"
 import { EditorConnection } from "./editorConnection"
 import { DocumentPosition, useClientNotificationEffect, useEvent } from "./util"
-
-interface RpcCallParams extends TextDocumentPositionParams {
-    sessionId: string
-    method: string
-    params: any
-}
-
-interface RpcReleaseParams {
-    uri: DocumentUri
-    sessionId: string
-    refs: RpcPtr<any>[]
-}
-
-const RpcNeedsReconnect = -32900
 
 class RpcSession implements Disposable {
     #ec: EditorConnection
--- a/vscode-lean4/src/infoview.ts
+++ b/vscode-lean4/src/infoview.ts
@@ -3,28 +3,14 @@ import {
     commands, Disposable, DocumentSelector,
     ExtensionContext, languages, Range,
     Selection, TextEditor, TextEditorRevealType,
-    Uri, ViewColumn, WebviewPanel, window, workspace, env, Position, Diagnostic,
+    Uri, ViewColumn, WebviewPanel, window, workspace, env, Position,
 } from 'vscode';
-import { EditorApi, InfoviewApi, LeanFileProgressParams, TextInsertKind } from '@lean4/infoview';
+import { EditorApi, InfoviewApi, LeanFileProgressParams, TextInsertKind, RpcConnectParams, RpcConnected, RpcKeepAliveParams } from '@lean4/infoview';
 import { LeanClient } from './leanclient';
 import { getInfoViewAllErrorsOnLine, getInfoViewAutoOpen, getInfoViewAutoOpenShowGoal,
     getInfoViewFilterIndex, getInfoViewStyle, getInfoViewTacticStateFilters } from './config';
 import { Rpc } from './rpc';
 import * as ls from 'vscode-languageserver-protocol'
-import { LanguageClient } from 'vscode-languageclient/node';
-
-interface RpcConnectParams {
-    uri: ls.DocumentUri;
-}
-
-interface RpcConnected {
-    sessionId: string
-}
-
-interface RpcKeepAliveParams {
-    uri: ls.DocumentUri
-    sessionId: string
-}
 
 const keepAlivePeriodMs = 10000
 
--- a/vscode-lean4/src/infoview.ts
+++ b/vscode-lean4/src/infoview.ts
@@ -181,7 +181,7 @@ export class InfoProvider implements Disposable {
                 // This event is triggered both the first time the server starts
                 // as well as when the server restarts.
 
-                this.clearSessions();
+                this.clearRpcSessions();
 
                 // The info view should auto-open the first time the server starts:
                 await this.autoOpen()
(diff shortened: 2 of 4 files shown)


vscode-lean4/src/infoview.ts
```diff
@@ -10,8 +10,50 @@ import { LeanClient } from './leanclient';
 import { getInfoViewAllErrorsOnLine, getInfoViewAutoOpen, getInfoViewAutoOpenShowGoal,
     getInfoViewFilterIndex, getInfoViewStyle, getInfoViewTacticStateFilters } from './config';
 import { Rpc } from './rpc';
-import { PublishDiagnosticsParams } from 'vscode-languageclient';
 import * as ls from 'vscode-languageserver-protocol'
+import { LanguageClient } from 'vscode-languageclient/node';
+
+interface RpcConnectParams {
```
Vtec234, line 16: Can we put these in `lean4-infoview/src/lspTypes.ts`?

vscode-lean4/src/infoview.ts
```diff
@@ -203,6 +269,11 @@ export class InfoProvider implements Disposable {
         this.serverNotifSubscriptions.clear();
     }
 
+    private clearSessions() {
```
Vtec234, line 272: `clearRpcSessions`? More to the point, `infoview.ts` is becoming a bit of a catch-all blob. What do you think about putting the RpcSession management in a utility class somewhere that `infoview` calls?

lean4-infoview/src/infoview/rpcSessions.tsx
```diff
@@ -52,15 +36,6 @@ class RpcSession implements Disposable {
         this.#ec = ec
         this.#uri = uri
 
-        // We setup a recurring timer to keep the RPC session alive.
-        this.#keepAliveInterval = window.setInterval(() => {
```
Vtec234, line 56: Hypothetically, other editors with different webview implementations may be able to keep the page running and wouldn't need to implement this in the extension. Thus I would keep this code around as it doesn't hurt to send more keep-alives then necessary. I would also note in `EditorApi` that the RPC session management only has to work if the webview tends to go to sleep.

vscode-lean4/src/infoview.ts
gebner: That's a good idea.

vscode-lean4/src/infoview.ts
gebner: Indeed the InfoProvider class is getting more and more complicated.  But splitting up the file won't help with that.

lean4-infoview/src/infoview/rpcSessions.tsx
gebner: The original motivation for the keep-alive approach was that the infoview doesn't get a close event in vscode and the keep-alive was the easiest way to close the session reliably.  This didn't work out, so we need to do this differently and ultimately no longer need keep-alives.  As you've pointed out, other editors might even be able to handle this in a cleaner way.

> other editors with different webview implementations may be able to keep the page running

I believe that's the standard behavior in electron, so we'll encounter it in other editors as well.

> I would also note in EditorApi that the RPC session management only has to work if the webview tends to go to sleep.

There's no supported way to detect this.  Note that the infoview will still respond to messages, it's only the setInterval call which is throttled.

The reason for this throttling is ultimately to reduce CPU and power usage by reducing the number of unnecessary wake-ups.  I don't want to work against that goal if there's a straightforward alternative.  It's not the highest priority, but having a lean file open in the background shouldn't reduce the battery life of my laptop.

vscode-lean4/src/infoview.ts
Vtec234: Perhaps we will need to refactor it then at some future point, but it is low priority.

lean4-infoview/src/infoview/rpcSessions.tsx
Vtec234: > The original motivation for the keep-alive approach was that the infoview doesn't get a close event in vscode and the keep-alive was the easiest way to close the session reliably. This didn't work out, so we need to do this differently and ultimately no longer need keep-alives.

The original motivation was nothing specific to VSCode, it is that we are building a distributed system in which the client may randomly break/loop/catch fire, and we want the server to still clean up its resources in any such case without relying on the client to behave correctly. I don't understand the claim that it "didn't work". It's unfortunate that we have to move the timer to the extension in order to maintain liveness, but it will work there, no?

> I believe that's the standard behavior in electron

Yes, but there are also webkit-based webviews and others.

> There's no supported way to detect this.

What I meant was not a dynamic check but rather an observation/specification of how a particular webview behaves will determine where in the code the timer needs to live.

> The reason for this throttling is ultimately to reduce CPU and power usage by reducing the number of unnecessary wake-ups. I don't want to work against that goal if there's a straightforward alternative. It's not the highest priority, but having a lean file open in the background shouldn't reduce the battery life of my laptop.

I absolutely agree, but note that this came from web browsers in which a malicious website can set up high-frequency timers/bitcoin miners/whatever. And with many tabs open, it makes sense to put old ones to sleep. We are not doing any of that. In any sane implementation, a 10s-interval timer will use close to no CPU or power, so I don't think this is a concern.

vscode-lean4/src/infoview.ts
```diff
@@ -3,16 +3,44 @@ import {
...
-import { EditorApi, InfoviewApi, LeanFileProgressParams, TextInsertKind } from '@lean4/infoview';
+import { EditorApi, InfoviewApi, LeanFileProgressParams, TextInsertKind, RpcConnectParams, RpcConnected, RpcKeepAliveParams } from '@lean4/infoview';
 import { LeanClient } from './leanclient';
 import { getInfoViewAllErrorsOnLine, getInfoViewAutoOpen, getInfoViewAutoOpenShowGoal,
     getInfoViewFilterIndex, getInfoViewStyle, getInfoViewTacticStateFilters } from './config';
 import { Rpc } from './rpc';
-import { PublishDiagnosticsParams } from 'vscode-languageclient';
 import * as ls from 'vscode-languageserver-protocol'
 
+const keepAlivePeriodMs = 10000
```
lovettchris, line 15: How did we come up with 10 seconds?  Can it be 20 seconds?  If you can move it to 20 seconds you will double my laptop battery life...

vscode-lean4/src/infoview.ts
```diff
@@ -3,16 +3,44 @@ import {
...
+}
+
+class RpcSession implements Disposable {
+    keepAliveInterval?: NodeJS.Timeout;
+
+    constructor(client: LeanClient, public sessionId: string, public uri: ls.DocumentUri) {
+        this.keepAliveInterval = setInterval(() => {
+            const params: RpcKeepAliveParams = { uri, sessionId }
+            try {
+                client.sendNotification('$/lean/rpc/keepAlive', params)
```
lovettchris, line 30: I wonder if this client.sendNotification could overlap with another "sendClientNotification" in another thread? Does it matter?  I wonder if "sendClientNotification" should clearInterval and restart the countdown so that a "keepAlive" doesn't slow down whatever work "sendClientNotification" is trying to do ?

vscode-lean4/src/infoview.ts
```diff
@@ -123,6 +153,25 @@ export class InfoProvider implements Disposable {
                 this.client.convertRange(show.selection)
             );
         },
+
+        createRpcSession: async uri => {
+            const sessionId = await rpcConnect(this.client, uri);
+            const session = new RpcSession(this.client, sessionId, uri);
```
lovettchris, line 159: I have a dumb question, do we need to do any keepAlives?  The RpcNeedsReconnect handling seems to work really well, so why not just let work after a 30 second delay in typing.  That way my battery life is increased even more... ?

lovettchris (changes requested):

vscode-lean4/src/infoview.ts
Vtec234: The 10s was chosen arbitrarily, to be more or less at an order of magnitude negligible in terms of power usage but reasonably short to a human. On which point, I am assuming you are joking about the battery life? A short computation every few seconds will have no effect at all at power usage in a normal editing session.

vscode-lean4/src/infoview.ts
Vtec234: Notifications are generally asynchronous, so it could overlap, but I don't think there's any issue with that. The LSP protocol is designed to be asynchronous.

vscode-lean4/src/infoview.ts
lovettchris: Actually I wasn't joking, if there's nothing else going on on my machine, switching from 10 to 20 seconds would literally double my battery life.  of course, that's a big if.  Every other app running on my system would have to have an excellently silent idle loop, and my RAM would have to be consuming very low amount of power (which is also not true).  So it's sort of a joke, but it's more a question of being a "well behaved" application that doesn't needlessly suck power while idle.

vscode-lean4/src/infoview.ts
lovettchris: Previously using the browser window.setTimeout it would not overlap with keystroke handling, that's why I ask.  This is a new multithreaded overlap potential.

Vtec234: I am confused, which keystroke handlers are you referring to? Btw, afaik async JS methods do not run in parallel on multiple threads, although they may interleave.

vscode-lean4/src/infoview.ts
Vtec234: I see, apologies for assuming that then. Still, looking at any modern system in its idle state, one observes multiple background services such as the window manager, OS interrupt handlers, sync services, a browser if you have one open, etc, waking up and doing bits of work multiple times a second. I am really confident that this 10s vs 20s does not matter for power draw at all. So changing this value will not make us any better behaved. Please disregard the humans comment (bad phrasing) and just take 10s to be the arbitrarily chosen value it is, within an order of magnitude of [TCP keepalives](https://en.wikipedia.org/wiki/Keepalive#TCP_keepalive) which inspired the protocol.

vscode-lean4/src/infoview.ts
lovettchris: Sorry, the setTimeout before was a web browser call in infoview, which I believe runs on the UI thread, the same thread that would be processing render operations that result in calls to sendClientNotification.  (Those sendClientNotification calls are usually the result of a a mouse hover or move cursor which also happens on keyboard input like "." and therefore I believe previous sendClientNotification versus keepAlive could not be interleaved).  But if you are saying LSP is already fully multithread safe then this is not an issue.

vscode-lean4/src/infoview.ts
lovettchris: Ok, I'll stop pushing then, and defer to my other dumb question below about whether we need keepalive at all.

lean4-infoview/src/infoview/rpcSessions.tsx
Vtec234: Update: we decided to remove the keepalive code from infoview for now. We can readd it if it is ever needed for another editor.

vscode-lean4/src/infoview.ts
Vtec234: If I am understanding the question correctly, the reason why we need to keep the session alive rather than just reconnect every now and then is that reconnects clear the serverside storage of RPC-referenced objects thus invalidating parts of the UI, for example the state of a running widget.

vscode-lean4/src/infoview.ts
Vtec234: I see, I think the same kind of negligibility argument applies here. Looking at the traffic of a typical LSP session (say by enabling server logging) reveals tens of requests per second made by the editor when making changes. One keepalive is really just lost in the woods here. Note also that only RPC requests keep the RPC session alive. `didChange` sent on keystrokes does not. This is intentional since we don't need a session when the infoview is closed.

Vtec234 (approved):

vscode-lean4/src/infoview.ts
gebner: Another related effect of keep-alives is that they keep the lean process in memory, which is undesirable in low-memory situations (where you want the OS to swap out the process).

Also note that we're not talking about a single process here.  It's extension host + lean watchdog + lean file worker (+ browser tab if we'd do it in the infoview).  At the moment the infoview also keeps sessions connected to all opened files, whether they are used or not (though we could conceivable optimize that), so effectively it's 10 * lean file worker.  And all of that * number of open vscode windows.

Generating the keep-alives in the infoview is also undesirable if you use vscode in a browser, as it will cause constant network traffic.  (I believe the extension host is run remotely.)

> Still, looking at any modern system in its idle state, one observes multiple background services such as the window manager, OS interrupt handlers, sync services, a browser if you have one open, etc, waking up and doing bits of work multiple times a second.

A couple of years ago there was a huge push to reduce wakups in the linux ecosystem (around the time Intel introduced tools like `powertop` with the explicit purpose of finding misbehaving programs that wake up unnecessarily).  Anecdotally, this effort has led to much improved battery runtimes (with 10+ hours being the expected norm instead of the exception).  Of course there are still some buggy programs out there, but I don't want to contribute to the problem if we can easily avoid it.

vscode-lean4/src/infoview.ts
gebner: The keepAlive notification won't make a difference here.  With either implementation, the notification can come between a completion request and it's response.  It shouldn't cause any problems,  (You can also have multiple requests in flights btw.) 

All LSP events are routed through the extension which runs in the extension host.  The only thing that changes is where the notification is "started": i.e., infoview->extension->lean vs. extension->lean.  Since you're asking about completions: these go through the extension as well, i.e. editor->extension->lean.

vscode-lean4/src/infoview.ts
gebner: #73 is an example of a problem we have because the reconnect doesn't work well.

PR #75, opened 2021-11-08
fix: move "show all abbreviations" command to docview.ts and make it more discoverable

lovettchris:
Like this:
image

And still available here:
image

both work and fixed a bug so the "back/forward" stack also works.

### DIFF
--- a/vscode-lean4/src/docview.ts
+++ b/vscode-lean4/src/docview.ts
@@ -165,9 +165,11 @@ export class DocViewProvider implements Disposable {
             const books = {
                 'Theorem Proving in Lean': 'https://leanprover.github.io/theorem_proving_in_lean4/',
                 'Reference Manual': 'https://leanprover.github.io/lean4/doc/',
-                'Mathematics in Lean': 'https://github.com/leanprover-community/mathlib4/',
                 'Abbreviations cheat sheet': mkCommandUri('lean4.docView.showAllAbbreviations')
             };
+            // TODO: add mathlib4 when we have a book about it
+            // 'Mathematics in Lean': 'https://github.com/leanprover-community/mathlib4/',
+
             for (const book of Object.getOwnPropertyNames(books)) {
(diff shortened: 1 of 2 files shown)


vscode-lean4/src/docview.ts
```diff
@@ -158,7 +165,8 @@ export class DocViewProvider implements Disposable {
             const books = {
                 'Theorem Proving in Lean': 'https://leanprover.github.io/theorem_proving_in_lean4/',
                 'Reference Manual': 'https://leanprover.github.io/lean4/doc/',
-                'Mathematics in Lean': 'https://github.com/leanprover-community/mathlib4/'
+                'Mathematics in Lean': 'https://github.com/leanprover-community/mathlib4/',
```
Vtec234, line 168: I would remove this, the book ['Mathematics in Lean'](https://leanprover-community.github.io/mathematics_in_lean/introduction.html) has not been ported to Lean 4. It is distinct from mathlib the library.

Vtec234 (approved):

vscode-lean4/src/docview.ts
lovettchris: right, I realize it is only a gihub repo, and not a book, but it's still a handy link though isn't it?

Vtec234: It is useful to know about, but I'm not sure of what use Mathlib4 would be to someone learning about Lean -- probably none. It seems not being a learning resource, it doesn't really fit with the other links.

lovettchris: I find every bit of sample code written in Lean4 that I can find to be extremely useful in helping to learn Lean4 and write Lean4 programs.

gebner: I completely agree with Wojciech, we should only include long-form documentation for Lean 4 here.  Emphasis on both documentation and Lean 4.  The documentation view is a poor browser for code.

lovettchris: Ok, that's 2 :-)  It is removed.

PR #78, opened 2021-11-18
minor fixes

lovettchris:
- readme updates
- fix how lean workspace folder is found by using getWorkspaceLeanFolderUri helper method.
- remove elan install 1 2 3 prompt since we don't really want user to customize elan install anyway.


README.md
```diff
@@ -1,7 +1,9 @@
 
 # Lean 4 VSCode Extension
-This extension provides VSCode editor support for the Lean 4 programming language.
-It is based on [vscode-lean](https://github.com/leanprover/vscode-lean).
+This extension provides VSCode editor support for the Lean 4 programming language. It is based on
+[vscode-lean](https://github.com/leanprover/vscode-lean) with the addition of an all new [Language
```
gebner, line 4: I wonder if we should keep mentioning the old Lean 3 extension, and so prominently in the first paragraph.  I don't think any user cares about the development history of this extension.  Also, "based on" suggests to me that the Lean 4 extension requires the Lean 3 one at runtime, which is of course not true.

If we want to mention it, I'd emphasize that it is for Lean 3, i.e. something like: "If you want to use Lean 3, then you need the separate [vscode-lean extension for Lean 3](https://github.com/leanprover/vscode-lean)."

gebner (commented):
Otherwise lgtm.

README.md
lovettchris: I agree.

lovettchris: Fixed, we have a later reference to lean 3 in the side by side support feature, so I think we're good without any lean 3 mention in the intro paragraph.

PR #80, opened 2021-12-03
Make progress bar red when processing stopped with an error

### LINKED ISSUE #67: Yellow bar doesn't disappear when there are invalid imports.
Steps to reproduce:

Create a new `foo.lean` file in VSCode, and `import bar` (without having created a `bar.lean`).

There is a red squiggle under `import bar` saying `unknown package 'bar'`, but the yellow bar (and the pop-up message "busily processing") never disappears.

Expected:

The yellow bar should disappear if nothing further is going to change, regardless of how successful processing the file was!

larsk21:
Depends on https://github.com/leanprover/lean4/pull/845, but also works (for me) with the current version of the server (the bar will still be yellow in that case).

closes #67

### DIFF
--- a/lean4-infoview/src/lspTypes.ts
+++ b/lean4-infoview/src/lspTypes.ts
@@ -20,6 +20,8 @@ export interface PlainTermGoal {
 export interface LeanFileProgressProcessingInfo {
     /** Range which is still being processed */
     range: Range;
+    /** If the file processing was stopped after a (fatal) error. */
+    error: boolean;
 }
 
 export interface LeanFileProgressParams {
--- a/lean4-infoview/src/lspTypes.ts
+++ b/lean4-infoview/src/lspTypes.ts
@@ -17,11 +17,16 @@ export interface PlainTermGoal {
     range: Range;
 }
 
+export enum LeanFileProgressKind {
+    Processing = 1,
+    FatalError = 2
+}
+
 export interface LeanFileProgressProcessingInfo {
-    /** Range which is still being processed */
+    /** Range for which the processing info was reported. */
     range: Range;
-    /** If the file processing was stopped after a (fatal) error. */
-    error: boolean;
+    /** Kind of progress that was reported. */
+    kind?: LeanFileProgressKind;
 }
 
 export interface LeanFileProgressParams {
--- a/vscode-lean4/src/taskgutter.ts
+++ b/vscode-lean4/src/taskgutter.ts
@@ -2,6 +2,10 @@ import { Disposable, ExtensionContext, OverviewRulerLane, Range, TextEditorDecor
 import { LeanClient } from './leanclient';
 import { LeanFileProgressKind, LeanFileProgressProcessingInfo } from '@lean4/infoview';
 
+// Workaround, since LeanFileProgressKind.* gives a 'self is not defined' error
+const ProgressKind_Processing = 1
+const ProgressKind_FatalError = 2
+
 class LeanFileTaskGutter {
     private timeout?: NodeJS.Timeout
 
@@ -42,7 +46,7 @@ class LeanFileTaskGutter {
                     editor.setDecorations(
                         decoration,
                         this.processed
-                            .filter(info => (info.kind === undefined ? LeanFileProgressKind.Processing : info.kind) === kind)
+                            .filter(info => (info.kind === undefined ? ProgressKind_Processing : info.kind) === kind)
                             .map(info => ({
                                 range: new Range(info.range.start.line, 0, info.range.end.line, 0),
                                 hoverMessage: message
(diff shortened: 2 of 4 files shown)


lean4-infoview/src/lspTypes.ts
```diff
@@ -20,6 +20,8 @@ export interface PlainTermGoal {
 export interface LeanFileProgressProcessingInfo {
     /** Range which is still being processed */
     range: Range;
+    /** If the file processing was stopped after a (fatal) error. */
+    error: boolean;
```
gebner, line 24:
```suggestion
    error?: boolean;
```

gebner: Older Lean versions don't include this flag.

Vtec234:
Looks good, but I guess we should wait for adjustments to the server-side PR and the corresponding changes here before merging.

larsk21:
Works and is backwards compatible, except for one bug I cannot find: When using any value from `LeanFileProgressKind` (declarations are fine) in `taskgutter.ts`, the extension crashes during startup with the error message: `Activating extension 'leanprover.lean4' failed: self is not defined`. If I write the numeral values instead of the enum values (for example `1` instead of `LeanFileProgressKind.Processing`), everything works fine. VSCode does not show any errors or warnings and the build runs successful.

vscode-lean4/src/taskgutter.ts
```diff
@@ -51,30 +66,57 @@ class LeanFileTaskGutter {
...
         this.subscriptions.push(
             window.onDidChangeVisibleTextEditors(() => this.updateDecos()),
             client.progressChanged((progress) => {
                 for (const [uri, processing] of progress) {
-                    this.status[uri.toString()] = processing.length === 0 ? undefined :
-                        Math.min(...processing.map(p => p.range.start.line));
+                    if (processing.length === 0) {
+                        this.status[uri.toString()] = undefined
+                    } else {
+                        const newStatus = new Map<LeanFileProgressKind, number>()
```
gebner, line 111: As discussed in the other issue, please just make one decoration per LeanFileProgressProcessingInfo (i.e., make LeanFileTaskGutter take a `processing: LeanFileProgressProcessingInfo[]` argument).  No need to do `Math.min`.

vscode-lean4/src/taskgutter.ts
```diff
@@ -51,30 +66,57 @@ class LeanFileTaskGutter {
...
-            overviewRulerColor: 'rgba(255, 165, 0, 0.5)',
-            dark: {
-                gutterIconPath: context.asAbsolutePath('media/progress-dark.svg'),
-            },
-            light: {
-                gutterIconPath: context.asAbsolutePath('media/progress-light.svg'),
-            },
-            gutterIconSize: 'contain',
-        });
+        this.decorations.set(LeanFileProgressKind.Processing, [
```
gebner, line 75: Is this where you get the exception?  And it goes away if you replace LeanFileProgressKind.Processing by 1?

larsk21: Yes, here and everywhere else in this file where I write `LeanFileProgressKind.something`.

vscode-lean4/src/taskgutter.ts
larsk21: I sure can do this, but this would be a greater change than just adding the red bar for fatal errors. Should this be included here or in its own PR?

vscode-lean4/src/taskgutter.ts
gebner: Apparently `self` is a browser-only feature and does not exist on node.js: https://developer.mozilla.org/en-US/docs/Web/API/Window/self

The take-away is that we probably shouldn't import `@lean4/infoview` in the extension host at all (since it depends on browser code which may fail when run outside of a browser).  For now, please just use 1 instead of the constant.

@Vtec234 Do you have any ideas on how to fix this properly?  Should we make a third package that contains the platform-independent code?  This was all a bit easier before the lerna setup, when we could just (and did) import single files.

vscode-lean4/src/taskgutter.ts
gebner: Doing it properly won't be larger than the current change.  Please include it here.

vscode-lean4/src/taskgutter.ts
Vtec234: Ahh that is annoying, afaik the usual solution is for a package to have two entry points, one for the browser and one elsewhere. NPM provides ways to set this up in `package.json`. Please feel free to try, otherwise I can look into it later next week.

vscode-lean4/src/taskgutter.ts
larsk21: Alright, I will do so later this week.

gebner:
Works great! Thanks!

PR #82, opened 2021-12-06
doc: sync readme with our lean quickstart.

lovettchris:


lovettchris:
@gebner , I'm finding this version of the readme is easier for new folks to follow since they don't need to install Elan & Lake manually.  What do you think?

gebner:
Both approaches have downsides.  Creating a directory without a lakefile and `leanprover/lean4:nightly` in the `lean-toolchain` file is far from the recommended setup, but it does make it easier to try out lean.

PR #83, opened 2021-12-07
fix: avoid errors from folders that have no lean-toolchain file yet.

lovettchris:
When loading a folder in VS code that contains lean code but no lean-toolchain file you get a weird error about server crashing with no easy way to fix it.  This change adds a nice popup showing option to allow user to select a toolchain.

image

this fixes a mess of popups that happen right now:
https://github.com/leanprover/vscode-lean4/issues/89


gebner:
I don't think the extension should ever create a `lean-toolchain` file automatically.  It happens very often that users open the wrong directory in vscode (that is, not the one where `lakefile.lean` is, but some subdirectory or even their home directory).  Adding a `lean-toolchain` file there will allow lean to start, but their setup will still be broken.

From what I can tell, you're doing this change so that users can smoothly create a project so that they can type some lean code.  For that, I think it would be better to add a command which executes `lake init` (or rather `lake +leanprover/lean4:nightly-2021-12-07 init someprojectname`, or `lake +leanprover/lean4:nightly init someprojectname` if we can't figure out the current nightly release).

If we want to mention this command in the "Lean failed to start" error message, then we should make it very clear that you're only supposed to execute this command in new projects that you just created yourself.

lovettchris:
subdirectory is taken care of it will find the lean-toolchain in parent directories.  What I'm fixing is if you open a lean file in a folder where there is no lean-toolchain file in that directory or any of it's parents, today you get a bad error server crashed 5 times and you can't fix it from vs code.  But since you feel so strongly about it I will add a popup message "We need to create a lean-toolchain file, click here to do it..." sort of thing and then it will be the user's choice if that is what they want or not.

gebner:
> But since you feel so strongly about it I will add a popup message "We need to create a lean-toolchain file, click here to do it..." sort of thing and then it will be the user's choice if that is what they want or not.

Yes please, and please also mention that this is for new projects only!  This is a real issue and we've had lots and lots of troubles with this over the years.  The Lean 3 extension has a lot of code to guide users to open the right directory.

Relevant Zulip discussion: https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/customizing.20vscode/near/202796216

tydeu:
> When loading a folder in VS code that contains lean code but no lean-toolchain file you get a weird error about server crashing with no easy way to fix it.

This is simply not true. I have a playpen folder with a multitude of `.lean` files that I play (*heh*) with all the time and that folder has no `lean-toolchain` file works fine (and I would very much want it to continue working fine without one).

lovettchris:
you get this error from `lean --version` in that folder:
```
error: no default toolchain configured. run `elan default stable` to install & configure the latest Lean 3 community release.
```

The server crashing was a bug in the vscode extension where it tries to start the server anyway, which doesn't work in this case.

tydeu:
@lovettchris 

That is because `elan` has no default toolchain configured so there is no such thing as `lean`  outside of a project. I do have default toolchain configured so everything works fine (as I would expect it to).

This goes back to my issues with the `elan` install created by VS Code not having a default toolchain.

tydeu:
@lovettchris one way to fix this would be have VS Code prompt the user to select a toolchain no default exists.

lovettchris:
@tydeu sure, I'll add that option to the popup prompt since I can't seem to get commands.execureCOmmand("vscode.openFolder", uri) to work properly...  so I have this:

image

lovettchris:
> Relevant Zulip discussion: https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/customizing.20vscode/near/202796216

Thanks for the link @gebner ,  I think most of that thread would be already solved by the fact that I now walk the parent folder hierarchy in findLeanPkgVersionInfo but this PR will also help in cases where the folder is completely orphaned.

tydeu:
@lovettchris 

> @tydeu sure, I'll add that option to the popup prompt

Does this prompt only appear when `lean --version` fails? As I said, a `lean-toolchain` file (or setting) is unnecessary as long as default toolchain is set in `elan`. That is, I want my playpen folder to continue using the default without issues (and without this prompt appearing).

lovettchris:
> Does this prompt only appear when `lean --version` fails? As I said, a `lean-toolchain` file (or setting) is unnecessary as long as default toolchain is set in `elan`. That is, I want my playpen folder to continue using the default without issues (and without this prompt appearing).

@tydeu - yes, in fact, only when `lean --version` returns `error: no default toolchain configured. run `elan default stable` to install & configure the latest Lean 3 community release.`

Kha:
> @tydeu sure, I'll add that option to the popup prompt since I can't seem to get commands.execureCOmmand("vscode.openFolder", uri) to work properly... so I have this:
> 
> image

The button labels look a bit confusing to me - if I choose lean-toolchain, I don't get to select a toolchain?

lovettchris:
`Create lean-toolchain file` does only that which is enough to unblock the issue.  But you can always manually invoke "Select Toolchain" from the command palette if you change your mind.

semorrison:
I think the "Select lean toolchain" button should be more explicit about what actual configuration change is being made. Is it setting a user-level default with elan? If is changing something just in this VS Code session? I should be able to infer which from the button text.

lovettchris:
@semorrison  I think you are trying to solve 2 problems here that perhaps don't belong in one button label.  The problem of what does "Select Toolchain" do exactly, and where does that state live and how long does it live, perhaps belongs in the Select Toolchain UI someplace, which I think can be tackled in a separate PR, no?

lovettchris:
@semorrison , @gebner , I'd like to get this PR in and resolve the other issue with Select Lean Toolchain later, this PR fixes this mess: https://github.com/leanprover/vscode-lean4/issues/89

Are you ok with that?

tydeu:
> @Kha `Create lean-toolchain file` does only that which is enough to unblock the issue. But you can always manually invoke "Select Toolchain" from the command palette if you change your mind.

I think Sebastian's question is more along the lines of which toolchain does `Create lean-toolchain file` stick in the file? I seems that should present a select toolchain dialogue to decide that. As such, it seems more reasonable for the prompt to only have a single "Select Lean Toolchain" option for the prompt (which may or may not create a toolchain depending on where we want that state to live).

lovettchris:
@tydeu, Oh, I see, yeah that sense now thanks for explaining.  I agree creating the lean-toolchain is related to that discussion, but what I'm trying to do is eliminate part 2 of step 2 in this [readme](https://github.com/leanprover/vscode-lean4/blob/541eecfc9bee0421f3236eb37c3ca02a121165a3/README.md)

> and a file named lean-toolchain containing just this:
```
leanprover/lean4:nightly
```

but simply doing that part for them with this button click.  So yes this PR only puts `leanprover/lean4:nightly` in the file for now, perhaps later we change that to `leanprover/lean4:stable`.

tydeu:
@lovettchris 

If this is just for new users who want to play around with Lean without making a project, why not just present the "Select Lean Toolchain", have them select a Lean toolchain to use, and just avoid the `lean-toolchain` concept entirely. Or, even better, since the goal of the quick start is also install `elan` why not just install `elan` with a default toolchain `leanprover/lean4:nightly`?

The main thing here is that users (especially new ones) really should not be creating `lean-toolchain` files themselves outside of a Lean/Lake project.

lovettchris:
> Or, even better, since the goal of the quick start is also to install `elan`, why not just install `elan` with a default toolchain of `leanprover/lean4:nightly` under these circumstances?

yes that works too, I will try that approach.

lovettchris:
Ok, this PR is simplified down to just this:
https://github.com/leanprover/vscode-lean4/issues/89

With the initial bootstrap of installing elan is now modified such that if there is now lean-toolchain it install elan with a default toolchain as suggested by @tydeu .

Adding @Kha  and @gebner - hopefully this PR is done now.

vscode-lean4/package.json
```diff
@@ -379,7 +379,7 @@
 		],
 		"configurationDefaults": {
 			"[lean4]": {
-                                "editor.unicodeHighlight.ambiguousCharacters": false,
+                "editor.unicodeHighlight.ambiguousCharacters": false,
```
tydeu, line 382: spurious edit?

vscode-lean4/src/utils/leanInstaller.ts
```diff
@@ -13,12 +13,14 @@ export class LeanInstaller implements Disposable {
     private subscriptions: Disposable[] = [];
     private prompting : boolean = false;
     private pkgService : LeanpkgService;
+    private defaultToolchain : string;
 
     private installChangedEmitter = new EventEmitter<string>();
     installChanged = this.installChangedEmitter.event
 
-    constructor(outputChannel: OutputChannel, localStorage : LocalStorageService, pkgService : LeanpkgService){
+    constructor(outputChannel: OutputChannel, localStorage : LocalStorageService, pkgService : LeanpkgService, defaultToolchain : string) {
```
tydeu, line 21: As far as I can tell, the `defaultToolchain` property of the installer is never actual used (and seems to conceptually conflict with the fact that a different `defaultToolchain` is passed to many of its functions).

vscode-lean4/src/utils/leanInstaller.ts
```diff
@@ -78,6 +82,9 @@ export class LeanInstaller implements Disposable {
             defaultPath = 'lean';
         }
         const installedToolChains = await this.elanListToolChains();
+        if (installedToolChains.length === 1 && installedToolChains[0] === 'no installed toolchains') {
+            installedToolChains[0] = this.defaultToolchain
```
tydeu, line 86: Shouldn't this be `defaultToolchain` and not `this.defaultToolchain`?

vscode-lean4/src/utils/leanInstaller.ts
tydeu, line 21: While I initially thought the `defaultToolchain` was unused,  I see now that is not that case (and somewhat get what is going on). However, I feel that this a bit confusing that the installer is initialized with a `defaultToolchain` but many of its methods ask for a different `defaultToolchain`.

tydeu: Also, is there any reason why it is not just using the `defaultToolchain` from the `LeanpkgService`?

tydeu:
@lovettchris Other than those minor code comments, this looks much better (at leas to me)! 🎉 Sorry for the rough, time consuming back-and-forth to get here! :(

tydeu:
One last note: on a text level, I would also suggest capitalizing 'Lean' in the prompt (as that is its standard spelling) and maybe 'Toolchain' as well (as you previously had done).

Kha:
> And the initial bootstrap of installing elan is now modified such that if there is no default lean-toolchain version it installs elan with a default toolchain of `leanprover/lean4:nightly` as suggested by @tydeu .

So whether elan is set up with a default toolchain depends on which project the user opens first, is that correct? We already have quite diverse elan setups depending on when and how the user set it up, but I'm not sure this inconsistency won't become a headache in the future.

tydeu:
@Kha I am not sure what other option there are though? 

One would have been to also set a default toolchain, but that would install 2 in projects with a set toolchain, which was considered a no-no when this was first discussed (though I personally still think it would be the most appropriate and consistent solution). 

Alternatively, if elan is already installed (but without a default toolchain), we could prompt to install one here. That is, use the 'select toolchain' prompt to instead select a default toolchain for VSCode to install.

What do you think?

lovettchris:
I have merged the code with master branch (there were some tricky conflicts with the progress notification change).  And this PR allows me to remove this bit from the readme:
```
    and a file named `lean-toolchain` containing just this:

    leanprover/lean4:nightly
```
because this PR now chooses `leanprover/lean4:nightly` when the folder has no lean-toolchain yet.  So this is one step closer to a smoother bootstrap...

Kha:
> Sadly, I am not sure if there is another good option here?

Well, Chris has already implemented a VS Code-only toolchain setting, no? So we could have VS Code use the selected toolchain without setting a default elan toolchain *or* writing a lean-toolchain file to disk. It won't work from the cmdline, but at that point you really want to make it a Lake project first anyway.

tydeu:
@Kha 

> So we could have VS Code use the selected toolchain without setting a default elan toolchain _or_ writing a lean-toolchain file to disk. 

That's sounds like a good solution to me! However, does the toolchain setting @lovettchris implemented currently allow you to select a toolchain that has not yet been installed?

lovettchris:
> That's sounds like a good solution to me! However, does the toolchain setting @lovettchris implemented currently allow you to select a toolchain that has not yet been installed?

@tydeu, yes that is precisely what this PR does now.  I removed the creation on lean-toolchain, however, and everything works fine, here's the full flow :

image
  
The select toolchain could be avoided if I save the installed version as the "workspace override" but the problem with doing that implicitly is that if the user adds a lean-toolchain file the "workspace override" will still override that which would be confusing.

Kha:
Looks good to me, except it still seems to install a default toolchain?

tydeu:
I second that. The flow looks good, but you should need to install a default toolchain if you are using the `+<toolcahin>` syntax. And if you do install a default toolchain, you should not be getting the no toolchain prompt. (because you have one, the default).

lovettchris:
@Kha, @tydeu Oh, of course, you are right, if there is a "(default)" toolchain I now use that instead of prompting for 
image so the right hand side of my flow is no longer necessary.

lovettchris:
@Kha, @Vtec234, and here's the flow that also happens when the user opens a folder that DOES contain a lean-toolchain file already.  E.g. they cloned mathlib4 or something:

image

lovettchris:
And the final flow that is the reason for this PR is:

image

tydeu:
@lovettchris The no 'elan /w lean-toolchain file` and 'elan w/ no default toolchain or lean-toolchain file' flows look great to me! ❤️ However, I do  have a question about the 'no elan w/ no default or no toolchain' flow. That is, this one:

> image

Why not prompt for a workspace override instead of  installing a default toolchain? That way it is consistent with the 'elan w/ no default toolchain or lean-toolchain file ' flow.

Kha:
@lovettchris Does this also work with an unsaved file (which I assume has no workspace?) as in https://github.com/leanprover/lean4/pull/842#discussion_r764261106?

lovettchris:
> @lovettchris Does this also work with an unsaved file (which I assume has no workspace?) as in [leanprover/lean4#842 (comment)](https://github.com/leanprover/lean4/pull/842#discussion_r764261106)?

Not sure I understand, can you provide repro steps?  When I open vs code witthout any folder and enter File/New File, I get "Untitled-1" without any .lean extension, so how can this activate the Lean vscode extension?

lovettchris:
> Why not prompt for a workspace override instead of installing a default toolchain? That way it is consistent with the 'elan w/ no default toolchain or lean-toolchain file ' flow.

I thought folks didn't like depending on "Select Toolchain" workspace overrides, so I favor the "(default)" toolchain in this brand new case (no elan or lean toolchain installed yet).  Is there a reason you don't want to steer the user towards using a  "(default)" toolchain?

tydeu:
@lovettchris 

> Not sure I understand, can you provide repro steps? When I open vs code witthout any folder and enter File/New File, I get "Untitled-1" without any .lean extension, so how can this activate the Lean vscode extension?

Open VS Code > New File > Text File > In New File ('Untitled-1') > Open Command Palette > Change Language Mode > 
 lean4 > Viola! Extension is active on new unsaved Lean 4 file

tydeu:
@lovettchris 

> I thought folks didn't like depending on "Select Toolchain" workspace overrides, so I favor the "(default)" toolchain in this brand new case (no elan or lean toolchain installed yet). Is there a reason you don't want to steer the user towards using a "(default)" toolchain?

My understanding was the original complaint was with the concept of workspace overrides altogether. Now that it seems like we are planning to continue supporting them, we might as well use them instead of a default toolchain to keep things consistent in VS Code. I think having one path lead to an override and one path lead to a default toolchain is very dangerous and will cause a lot of confusion. VS Code should either do one or the other -- not a mixture. Having VS Code fiddle opaquely with elan seems unwise, so I think the override is a good plan.

lovettchris:
> My understanding was the original complaint was with the concept of workspace overrides altogether. Now that it seems like we are planning to continue supporting them, we might as well use them instead of a default toolchain to keep things consistent in VS Code. I think having one path lead to an override and one path lead to a default toolchain is very dangerous and will cause a lot of confusion. VS Code should either do one or the other -- not a mixture. Having VS Code fiddle opaquely with elan seems unwise, so I think the override is a good plan.

I'm not sure I agree with that.  In the flows above VS code is never setting a "workspace override" without the user choosing "Select Toolchain" button.  I think doing that opaquely is worse than installing Elan with a --default-toolchain of leanprover/lean4:nightly.  Besides when a fodler is opened with a `lean-toolchain` file then we also do not use the "workspace override".

lovettchris:
> @lovettchris
> 
> > Not sure I understand, can you provide repro steps? When I open vs code witthout any folder and enter File/New File, I get "Untitled-1" without any .lean extension, so how can this activate the Lean vscode extension?
> 
> Open VS Code > New File > Text File > In New File ('Untitled-1') > Open Command Palette > Change Language Mode > lean4 > Viola! Extension is active on new unsaved Lean 4 file

Oh, thanks, didn't know you could change the language like that.  I fixed on bug related to this scenario, but the "lean --server" does not seem to be working on this file, and when I edit it nothing shows up in the infoview and no exceptions are being thrown, so I think we have other issues here, I filed a [new bug](https://github.com/leanprover/vscode-lean4/issues/92) for this.

tydeu:
@lovettchris 

> I'm not sure I agree with that. In the flows above VS code is never setting a "workspace override" without the user choosing "Select Toolchain" button. I think doing that opaquely is worse than installing Elan with a --default-toolchain of leanprover/lean4:nightly.

What I was suggesting was to install Elean with a default toolchain of none and then present a "Select Toolchain" button for the user to set the override for the workspace. No opaque activity needed. I think hard coding the default toolchain to install in the extension is unwise. What toolchain to use should be an option left up to the user.

clovett:
> I think hard coding the default toolchain to install in the extension is unwise

Can you tell me more, why does Elan provide the default toolchain option if it is unwise to use it?

tydeu:
@clovett 

It is not that using a default toolchain is, in general, unwise. It is that VS Code opaquely configuring it that is problematic. If you have manually installed elan and gone through the process of selecting a default toolchain for it to install (like I did), great.  

Thus, if VS Code is installing it, it should provide similar configuration options. That is, either by prompting the user to select the default toolchain to install with it or by installing none and instead prompting the user to select a toolchain for the workspace override (and this behavior should be consistent with the other flow where the later is currently done). It should not, I feel, opaquely install some toolchain without asking for the user's input. This will be especially important once there are both active `stable` and `nightly` distros -- the user should not magically be assigned one without notification.

lovettchris:
Ok, I have fixed issue https://github.com/leanprover/vscode-lean4/issues/91 in this PR - it now always applies the findLeanPkgVersionInfo algorithm no matter which file the file watcher found.  I also fixed a bug in getWorkspaceLeanFolderUri so that if you open lean4 with D:\git\leanprover\lean4\src\lake\Lake\Version.lean in the editor it does NOT find the lake lean-toolchain file.  Perhaps this is what both of you were seeing above which I agree is confusing.  It now correctly uses `workspace.getWorkspaceFolder(documentUri)` to find the workspace folder and starts the search from there.  So opening lean4 with no "(default)" toolchain installed now correctly shows the select toolchain dialog.  Hopefully the PR is done now.

gebner:
> It now correctly uses workspace.getWorkspaceFolder(documentUri) to find the workspace folder and starts the search from there.

This is good.

> Hopefully the PR is done now.

I'd also like to get this PR merged, it's been sitting for a while and fixes a few bugs.  From my point of view, there's two things to be done:
1. Mac made two comments, 1) about a spurious whitespace change and 2) that some functions in LeanInstaller still have a `defaultToolchain` argument even though the class now has a `defaultToolchain` field.  Please address those comments.
2. There's now a merge conflict with master, please resolve it.

vscode-lean4/package.json
lovettchris: yes, will undo, but I'm also surprised to see a text file that is using actual tab characters for indentation.  My editor automatically converts those to spaces.

vscode-lean4/src/utils/leanInstaller.ts
lovettchris: You are right, installElan should be a private method and it no longer needs a defaultToolchain  argument.  So I removed it.

lovettchris:
> I'd also like to get this PR merged, it's been sitting for a while and fixes a few bugs. From my point of view, there's two things to be done:

@gebner, ok, I have fixed those two additional issues and retested all scenarios.  Should be good now.

lovettchris:
This is the final flow:
image

tydeu:
> This is the final flow

@lovettchris Wait, I thought that for no elan, no lean-toolchain, the desired flow was to prompt for a list of toolchains to install, and not just install the fixed toolchain of `leanprover/lean4:nightly`? Isn't that what [you said](https://leanprover.zulipchat.com/#narrow/stream/147302-lean4-dev/topic/.5BRFC.5D.20vscode.3A.20installing.20default.20toolchain/near/265975368) on Zulip?

lovettchris:
The flow I've shown is accurate, the idea of prompting the user with a list of known good toolchains by doing some sort of http query against https://github.com/leanprover/lean4-nightly/releases is not yet implemented.  I filed this as a separate issue: https://github.com/leanprover/vscode-lean4/issues/96

lovettchris:
Plus I _really_ want to call this LONG PR done already...

gebner:
Okay, from what I can tell all the obvious issues are fixed.  Let's see what this breaks in the real world and take further improvements to new PRs.

PR #86, opened 2021-12-10
fix: new lines appearing in hover tooltips.

lovettchris:
Tricky bug, the problem seems to be a bug in the Chrome browser where a position="absolute" `<div>` appearing in a string of `<spans>` can affect the layout of those spans causing them to take just a couple pixels more room which is enough to cause them to line wrap, since the outer box is not expanded to contain the larger spans.  The fix is to move the `<Tippy>` div to a different location in the document which can be done using the [appendTo](https://atomiks.github.io/tippyjs/v6/all-props/#appendto) property.

This also fixes a problem with "click pinning" of Tippy popups and also fixes a problem with tips going up bouncing off the top and being flipped down then the very next tip goes up again, instead it remembers a global direction until the next flip so you can now play a proper game of `pong` with tool tips :-)

image


lean4-infoview/src/infoview/interactiveCode.tsx
```diff
@@ -58,6 +58,14 @@ const LazyTippy = React.forwardRef<HTMLElement, TippyProps>((props, ref) => {
     computedProps.content = mounted ? props.content : undefined
   }
 
+  if (computedProps.appendTo == 'parent'){
+    // need to go some levels higher because we sometimes have nested spans and so parent is
+    // not enough to avoid weird layout bugs where Chrome does not like a <div> popping up
+    // in the middle of a string of spans, even though the div is position:absolute, it is
+    // causing newlines to appear (bug https://github.com/leanprover/vscode-lean4/issues/51)
+
```
gebner, line 66: This is an empty if?  (The `appendTo=` property below would also be a good place for the comment.)

gebner (approved):

lean4-infoview/src/infoview/interactiveCode.tsx
lovettchris: woops, yeah, dead code sorry.  removed.

gebner: Could you please add the comment back somewhere?  It's good to know what the document.body setting is for.

lovettchris: sure thing, done.

Vtec234 (approved):
Yay, thanks for fixing this! The different DOM structure resulting from not appending hovers at the parent unfortunately breaks the code keeping nested hovers alive, i.e. the ability to hover on part of a hover and then on part of that, etc, to explore nested subexpressions. With this PR they just disappear. But the newline bug is worse and the hover logic is half-broken anyway, so I'd say let's merge this. I tried hard to get Tippy to do the right thing for nested hovers and it still quite doesn't. I think the solution will have to be reverting back to Popper and then writing all the hover logic ourselves again, which will be a fair bit of work. Could you please leave a TODO comment to fix nested hovers?

lovettchris:
> I tried hard to get Tippy to do the right thing for nested hovers and it still quite doesn't.

I think I have this solved and added a globalPlacement feature that works nicely so tooltips down stick to the top of the window any more.

lean4-infoview/src/infoview/interactiveCode.tsx
```diff
@@ -61,6 +61,8 @@ const LazyTippy = React.forwardRef<HTMLElement, TippyProps>((props, ref) => {
   return <Tippy {...computedProps} ref={ref} />
 })
 
+let globalPlacement : Placement = 'top';
```
Vtec234, line 64: This should probably be a `React.useRef` inside `HoverableTypePopupSpan` to avoid multiple span interfering with each other.

lean4-infoview/src/infoview/interactiveCode.tsx
```diff
@@ -145,10 +151,39 @@ const HoverableTypePopupSpan =
...
+    if (span._tippy) {
+      return true;
+    }
+    return false;
+  }
+
   return (
     <LazyTippy
       ref={ref}
+      // To avoid weird layout bugs where Chrome does not like a <div> popping up
```
Vtec234, line 166: Are we sure that this is a Chrome bug rather than some weird edge case in the HTML layout specification, in which case it would also misbehave in other browser engines? If we're not, I'd just write we're working around a layout issue.

Vtec234 (approved):

lean4-infoview/src/infoview/interactiveCode.tsx
Vtec234: I also noticed that as soon as one popup is drawn downwards, all the new ones including a new initial popup (i.e. non-nested) are as well, even if the original tree which was growing downwards is long gone. I'm guessing we could reset to `top` once the entire tree rooted at a given span has disappeared.

gebner: I'm not sure if `useRef` is the correct mechanism here, but global variables are almost always the wrong one.  As far as I understand, the idea behind `globalPlacement` is that nested tooltips use the same placement strategy as their parent, if possible.  So if the first tooltip goes down, then its children should go down as well.  With a global variable, we use the placement from the last shown popup, instead of from the parent popup.

This seems like a good fit for contexts: make a ref inside `HoverableTypePopupSpan` initialized to the context value, and then propagate the ref to the child via the context.

lovettchris: Correct, it is trying to influence the next popup placement, and it is hard to communicate across React components but I will try and find a non-global solution...

lean4-infoview/src/infoview/interactiveCode.tsx
lovettchris: I removed the "bug" language

lean4-infoview/src/infoview/interactiveCode.tsx
lovettchris: Ok, I removed the global variable - the solution is to "pass down" the placement strategy from tip to tip using props.  Unfortunately, that results in way more code churn catching all the places tips can start and chaining it through to all the places another tip can be started.  Props seems to be the only real way to pass things down the react module chain.  The nice side effect of this fix is that the tooltips only remember this "inherited state" and once that chain of tips is discarded the next tip goes "upwards" again even if the previous chain was going "downwards", whereas the global variable did not behave that way.

lovettchris: Ah ha, and bringing this full circle, the same fix can be used to fix where the window timeout object is being stored, moving from UseRef to this inherited property.  So I renamed the property "TipChainState" and now hover tips also stay open correctly and you can really go to town like this without any mouse clicks:

image

gebner: > Unfortunately, that results in way more code churn catching all the places tips can start and chaining it through to all the places another tip can be started.

Right, that's why I thought contexts would help here because they can pass state through components without having to add extra props.  You should be able to do something like this:
```typescript
export const TipChainContext = React.createContext<TipChainState>(undefined as any);
export function WithNewTipChainState({children}: {children: React.ReactNode}) {
    const [tipChain, _] = React.useState(() => new TipChainState)
    return <TipChainContext.Provider value={tipChain}>{children}</TipChainContext.Provider>
}

// sets the context value for the child inside the <WithNewTipChainState>
<WithNewTipChainState><InteractiveCode pos={pos} fmt={goal.type} /></WithNewTipChainState>

// read the context value
const tipChainState = React.useContext(TipChainContext)
```

lovettchris: Hmmm, still have to wrap everything with <WithNewTipChainState> so it is not really any better.  The code I have now works, please try it and technically I'm on vacation now :-)

Vtec234: It seems the tooltips no longer stay open when the mouse is over a non-interactive part of the tooltip, e.g. a doc comment.

lovettchris: > It seems the tooltips no longer stay open when the mouse is over a non-interactive part of the tooltip, e.g. a doc comment.

Oh, I see what the problem is, it is clearing too much (the whole chain) rather than just up to the tip you left.

lovettchris: @Vtec234 well I might need some help on this one, I've been beating my head on it for a bit and the parent chain is working nicely now except for the but you mention above.  The thing that is confusing me is when you move the mouse out of an interactive part into the non-interactive part there is now "enter" mouse event from anywhere to tell me the mouse is still inside the parent of the interactive part.  Can you think of a place to put another mouseEnter event to catch that?

lovettchris: Interesting, HTML behaves correctly with the pointerout and pointerover events, but this is not what I'm seeing in the infoview. See attached.  If we can get the events to behave this way then my code handling those events will work perfectly.

[test.zip](https://github.com/leanprover/vscode-lean4/files/7722984/test.zip)

lovettchris: @Vtec234 , ok, I figured out how to get the events, so I'm closer now:
```html
<div
          onPointerOver={onPointerOverContent}
          onPointerOut={onPointerOut}>
        <TypePopupContents
          redrawTooltip={() => {
            tippyInstance.current?.popperInstance?.update()
          }}
          {...props}
        /></div>
```
the only problem now is that the `onPointerOverContent` is seeing the wrong `tippyInstance.current?.id` which is odd, it always has the id of the top level tip...

lovettchris: @Vtec234, ok, I think I've got it now, I had to fix out onPointerOverContent finds the parent Tip id, with a new helper function getTipId.  Please try it out now and if you agree this is good I'll remove all the debug console output stuff I needed to debug this.

lovettchris: @Vtec234, @gebner, ok please play with it now I made a couple more tweaks, I think it works really well now.

Vtec234: @lovettchris : Thank you for working on this. I will be able to look at the code later, but before that some thoughts:
- Your code is essentially a custom implementation of debouncing, which removes any need to use Tippy.js instead of Popper.js. In other words, Tippy.js is only useful for `hideWithInteractivity` which hides a lot of complexity. If we're doing that ourselves, we should use the underlying Popper.js.
- It appears to be working reasonably well, but I quickly found a few bugs. For example, the highlights "stick around" as in this [video](https://user-images.githubusercontent.com/13901751/146618531-1a7699af-aa1b-4a12-aeb2-b64ad59d55f4.mp4), and sometimes they also appear on the source span when hovering over some part of its tooltip. There are **lots** of annoying, small things like this which have to work. This is why I was hoping to avoid implementing this ourselves. But I am not sure that's possible as Tippy seemd to hardcode too much. I am happy to work with you to q/a these, but a fair warning -- I expect it to be a little tedious.
- In terms of the code, I support Gabriel's suggestion to use contexts instead of explicit state-passing.

lovettchris: Ok, but the current behavior of this PR is better than before in my opinion, so yes, I'd like to work on this incrementally, getting this PR in as a stake in the ground moving forwards.  Then we can work on the Context idea, then the Popper.js idea as separate PR.s  

Thanks for the video, but can you explain the bug in the video?  I see you pop up the tip for String.Pos, you move inside it stays, you move outside the tip it goes away, good, then you draw some circles with the mouse, and I don't know what that's about, then you move over the arrow and "String.Pos -> String.Pos : Type" pops up and you immediately move the mouse down so that popup also goes away?  That all looks good to me, what were you expecting?

Vtec234: Oh, the issue is that the blue highlight appears on all of `String.Pos -> String.Pos` when hovering over the tooltip even though I was only hovering over `String.Pos`, and that it does not disappear while the mouse is elsewhere -- I had to hover over the whole expression to make it disappear.

clovett: @Vtec234 thanks I see now thanks.  I fixed this bug, and I switched to useContext which cleaned things up a lot and fixed some bugs in root level tips so there is guaranteed to be only one TipChainState now for the entire page which works better.  See [this commit](https://github.com/leanprover/vscode-lean4/pull/86/commits/67c4dabf82ef41e6c4a81acaa43d494d5071d9e7).

Vtec234: @lovettchris : I figured out why everything was so broken. It turns out that Tippy.js somehow blocks events such as `onpointerover` from bubbling up from the tooltip into our component, but only sometimes. It's not even consistent as events still bubble up from *some* parts of tooltip contents such as links, but not from text and the background area. So I removed Tippy.js and rewrote the component with Popper [here](https://github.com/Vtec234/vscode-lean4/tree/fix_51/), and now everything including keeping clicked tooltips displayed works. Also the newline bug is fixed. With this we can just use simple event handlers and don't need global `TipChainState`. I restored the old Lean 3 widget look for which we still have the CSS around. Could you please try it out? If it works, are you okay with me pushing that to this branch?

lovettchris: @Vtec234 I cloned your branch and tried it and I get no tooltips at all; the developer console shows this warning:

```
index.js:480 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `TermGoal`. See https://reactjs.org/link/warning-keys for more information.
    at option
    at TermGoal (https://file+.vscode-resource.vscode-webview.net/d%3A/Temp/vtec234/vscode-lean4/vscode-lean4/media/webview.js:2301:21)
    at div
    at details
    at Details (https://file+.vscode-resource.vscode-webview.net/d%3A/Temp/vtec234/vscode-lean4/vscode-lean4/media/webview.js:2103:20)
    at div
    at details
    at Details (https://file+.vscode-resource.vscode-webview.net/d%3A/Temp/vtec234/vscode-lean4/vscode-lean4/media/webview.js:2103:20)
    at InfoDisplay (https://file+.vscode-resource.vscode-webview.net/d%3A/Temp/vtec234/vscode-lean4/vscode-lean4/media/webview.js:2366:82)
    at Info (https://file+.vscode-resource.vscode-webview.net/d%3A/Temp/vtec234/vscode-lean4/vscode-lean4/media/webview.js:2434:51)
    at div
    at Infos (https://file+.vscode-resource.vscode-webview.net/d%3A/Temp/vtec234/vscode-lean4/vscode-lean4/media/webview.js:2523:51)
    at div
    at Main
```

Vtec234: @lovettchris Are you sure you are using the correct branch? If memory serves me right, the `TermGoal` component has not been in use since [this commit](https://github.com/leanprover/vscode-lean4/commit/73a23763d6d0963fdb6b1245aa486f073e9bdbd1).

lovettchris: my bad I got it working, it works well, I like the clever `{pinParent: () => {setState('pin'); tipChainCtx.pinParent()}}` trick, the mouse click 'pinning' of the whole open chain works better with your code.  Thanks for also keeping the screen bounce logic so chain opens downwards when you bounce off the top of the window.

Scrolling while a large chain is pinned open gets interesting, see https://popper.js.org/docs/v2/faq/#why-is-my-popper-jittering-while-scrolling

lovettchris: I'm going to just close this pull request once you are done with yours.

Vtec234: Interestingly I am seeing more jitter on scrolling with the recommended value of `strategy: 'fixed'` compared to the default.

> I'm going to just close this pull request once you are done with yours.

I see, I don't think I am able after all to push directly to your branch, so instead I did open #94.

clovett:
replaced by PR #94.

PR #87, opened 2021-12-10
fix: use progress notification instead of OutputChannel for lean installation

lovettchris:
image

image

image


gebner:
Great! This looks much better in real life, since the notifications don't stack and disappear when they're done.

PR #93, opened 2021-12-28
chore: split infoview API into separate package

Vtec234:
Related to the [discussion](https://github.com/leanprover/vscode-lean4/pull/80#discussion_r767208732) on #80. We should not import all of `@lean4/infoview` from the extension host (a Node.js environment) since it references browser APIs, but we need part of it. According to [here](https://github.com/Microsoft/TypeScript/issues/29128), this split is best achieved with an actual split into two packages rather than the `browser`/`main` fields of `package.json`. So this introduces `@lean4/infoview-api` with just a few types and interfaces.


gebner:
Thanks!

PR #94, opened 2022-01-05
fix: new lines appearing in hover tooltips [continued]

Vtec234:
Continues the work from #86. It should hopefully be ready to merge as most of the review was already done there. CC @lovettchris

### DIFF
--- a/lean4-infoview/package.json
+++ b/lean4-infoview/package.json
@@ -25,7 +25,6 @@
   },
   "dependencies": {
     "@lean4/infoview-api": "^0.1.0",
-    "@tippyjs/react": "^4.2.6",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "react-fast-compare": "^3.2.0",
--- a/lean4-infoview/src/infoview/tooltips.tsx
+++ b/lean4-infoview/src/infoview/tooltips.tsx
@@ -76,12 +76,12 @@ export const Tooltip = forwardAndUseRef<HTMLDivElement,
 export const HighlightOnHoverSpan = forwardAndUseRef<HTMLSpanElement, React.HTMLProps<HTMLSpanElement>>((props, ref, setRef) => {
   const [isPointerOver, setIsPointerOver] = React.useState<boolean>(false)
   const onPointerOver = (e: React.PointerEvent<HTMLSpanElement>) => {
-    if (ref.current && e.target == ref.current)
+    if (ref.current && e.target === ref.current)
       setIsPointerOver(true)
   }
 
   const onPointerOut = (e: React.PointerEvent<HTMLSpanElement>) => {
-    if (ref.current && e.target == ref.current)
+    if (ref.current && e.target === ref.current)
       setIsPointerOver(false)
   }
 
@@ -187,14 +187,14 @@ export const WithTooltipOnHover =
     // event was handled with a property.
     if ('_WithTooltipOnHoverSeen' in e) return
     if (!isWithinHoverable(e.target)) return
-    (e as any)['_WithTooltipOnHoverSeen'] = {}
+    (e as any)._WithTooltipOnHoverSeen = {}
     startShowTimeout()
   }
 
   const onPointerOut = (e: React.PointerEvent<HTMLSpanElement>) => {
     if ('_WithTooltipOnHoverSeen' in e) return
     if (!isWithinHoverable(e.target)) return
-    (e as any)['_WithTooltipOnHoverSeen'] = {}
+    (e as any)._WithTooltipOnHoverSeen = {}
     startHideTimeout()
   }
 
--- a/lean4-infoview/src/infoview/tooltips.tsx
+++ b/lean4-infoview/src/infoview/tooltips.tsx
@@ -1,30 +1,85 @@
 import * as React from 'react'
+import * as ReactDOM from 'react-dom'
 
-import { Instance as TippyInstance, Props as RawTippyProps, Placement as TippyPlacement } from 'tippy.js'
-import { default as Tippy, TippyProps } from '@tippyjs/react'
-import 'tippy.js/dist/tippy.css'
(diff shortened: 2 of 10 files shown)


lean4-infoview/package.json
```diff
@@ -25,7 +25,6 @@
   },
   "dependencies": {
     "@lean4/infoview-api": "^0.1.0",
-    "@tippyjs/react": "^4.2.6",
```
lovettchris, line 28: The package-lock.json also needs updating with this.

lovettchris (approved):

lean4-infoview/package.json
Vtec234: Indeed, thanks!

lean4-infoview/src/infoview/tooltips.tsx
```diff
@@ -0,0 +1,221 @@
...
+  const onClick = (e: React.PointerEvent<HTMLSpanElement>) => {
+    if (!isWithinHoverable(e.target)) return
+    e.stopPropagation()
+    clearTimeout()
+    setState(state => state === 'pin' ? 'hide' : 'pin')
+  }
+
+  React.useEffect(() => {
+    const onClickAnywhere = (e: Event) => {
+      if (!logicalDom.contains(e.target as Node)) {
```
gebner, line 147:
```suggestion
      if (e.target instanceof Node && !logicalDom.contains(e.target)) {
```

I'm not 100% sure if this can happen, but it seems good practice regardless.

lean4-infoview/src/infoview/util.ts
```diff
@@ -180,10 +180,73 @@ export type Keyed<T> = T & { key: string };
...
+ * register it in the `LogicalDomContext`.
+ *
+ * To carry out checks, `useLogicalDom` with a ref to the logical DOM root and wrap the root
+ * in a `LogicalDomContext` using the resulting `LogicalDomStorage`. */
+export function useLogicalDom(ref: React.RefObject<HTMLElement>): [LogicalDomTraverser, LogicalDomStorage] {
+  const parentCtx = React.useContext(LogicalDomContext)
+  React.useEffect(() => {
+    if (ref.current)
+      parentCtx.registerDescendant(ref.current)
+  }, [ref])
```
gebner, line 235: What if parentCtx changes?

lean4-infoview/src/infoview/tooltips.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  // We are pinned when clicked, shown when hovered over, and otherwise hidden.
+  type TooltipState = 'pin' | 'show' | 'hide'
+  const [state, setState] = React.useState<TooltipState>('hide')
+  const shouldShow = state !== 'hide'
+
+  const tipChainCtx = React.useContext(TipChainContext)
+  React.useEffect(() => {
+    if (state === 'pin') tipChainCtx.pinParent()
+  }, [state])
```
gebner, line 119: What if tipChainCtx changes?

lean4-infoview/src/infoview/tooltips.tsx
```diff
@@ -0,0 +1,221 @@
...
+  return ReactDOM.createPortal(popper, document.body)
+})
+
+/** A `<span>` element which gets highlighted when hovered over. It is implemented with JS rather
+ * than CSS in order to allow nesting of these elements. When nested, only the smallest nested
+ * element is highlighted. */
+export const HighlightOnHoverSpan = forwardAndUseRef<HTMLSpanElement, React.HTMLProps<HTMLSpanElement>>((props, ref, setRef) => {
+  const [isPointerOver, setIsPointerOver] = React.useState<boolean>(false)
+  const onPointerOver = (e: React.PointerEvent<HTMLSpanElement>) => {
+    if (ref.current && e.target === ref.current)
```
gebner, line 79: Will this break if we add additional elements (such as spans for syntax highlighting)?

Vtec234: It will break in some cases with more deeply nested children, but this is intentional. In the case you mention this is what we want, since we only want the innermost span to have a highlight. Note also that React propagates `onPointerOver` events from children in the React tree even if they are not children in the DOM, so hovering over a tooltip will bubble up an event up to the hoverable span that generated the tolltip. Without the `=== ref.current` check this would highlight the span which is not what we want. If we want to make something other than the directly nested text or span highlightable, I think the only reliable way to achieve it is by passing an explicit ref.

lovettchris: more comments would help in the code too explaining the tricky parts of the implementation, like the info you had about how React propagates onPointerOver events from children in the React tree even if they are not children in the DOM and how the ref.current check works and so on...

lovettchris:
@Vtec234 , your pinning is very strong such that I can pin one chain open and then start another.  I'm not sure you could do that with my version.  Is this intentional?

image

lovettchris:
@Vtec234 your "placement" memory and going downwards doesn't always work:  This chain went up, bounced off the tip but then the next tip went upwards again creating quite the mess...

image

Vtec234:
>  I can pin one chain open and then start another. I'm not sure you could do that with my version. Is this intentional?

You are right, and yes this is intentional -- I thought it could be useful to have multiple subexpressions open.

> This chain went up, bounced off the top of the window, but then the next tip went upwards again creating quite the mess...

Hm, I am unable to reproduce this but looking at the video it seems like something has gone wrong with Popper rather than our extension. In the chain created at 0:14, the third popup isn't quite located below the second, more like on top. So I'm assuming its Popper-determined placement was still 'top', and thus the fourth one also goes up.

Vtec234:
@lovettchris are you happy with the explanation? If so let's merge.

lovettchris:
> @lovettchris are you happy with the explanation? If so let's merge.

Yes, let's merge this.

PR #95, opened 2022-01-05
fix: support for running lean server on 'untitiled' files.

lovettchris:
image

### DIFF
--- a/vscode-lean4/src/taskgutter.ts
+++ b/vscode-lean4/src/taskgutter.ts
@@ -14,6 +14,7 @@ class LeanFileTaskGutter {
         const oldProcessed = this.processed;
         this.processed = processed;
         if (processed === undefined) {
+            this.processed = []
             this.clearTimeout();
             this.updateDecos();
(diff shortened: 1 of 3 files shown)


vscode-lean4/src/taskgutter.ts
```diff
@@ -14,6 +14,7 @@ class LeanFileTaskGutter {
         const oldProcessed = this.processed;
         this.processed = processed;
         if (processed === undefined) {
```
gebner, line 16: Mmmh, the type annotations are a bit suspect here... `processed` is not supposed to be `undefined`.  I wonder why the call site in `updateDecos` even type checks.  (Ah, I guess we should eventually enable another typescript option for this: `strictNullChecks`)

lovettchris: I actually hit an unhandled exception in updateDecos where this.processed was undefined.  But it was very hard to repro, so I figured since line 16 already checked for undefined and calls updateDecos we'd better set this.processed to something we can call .filter on.

PR #97, opened 2022-01-08
feat: use lake serve

lovettchris:
See [Zulip Discussion](https://leanprover.zulipchat.com/#narrow/stream/147302-lean4-dev/topic/Expose.20--load-dynlib.20to.20Lean.3F)

Proposal, replace the following setting:
image

With:
image

Then the rules for using `lake serve` would be:
1. is 'lake' or 'lake.exe' found in the toolchainPath, If not don't use `lake`
2. is there a `lakefile.lean` in the current workspace root?  If not don't use `lake`
3. If the above are both good, use `lake serve` with optional +version if there's a workspace override in effect.
4. If `lake serve -- args` fails fall back on `lean --server`.


vscode-lean4/src/utils/batch.ts
```diff
@@ -39,3 +39,55 @@ export async function batchExecute(
     });
 
 }
+
+export async function testExecute(
+    toolchainPath: string,
+    args: any[],
+    workingDirectory: string,
+    delay: number = 1000): Promise<number> {
```
gebner, line 47: If you start `lake serve` on a new lean version, then elan will certainly take more than one second to download it.  Also `lake serve` seems to do some initialization, which we probably shouldn't interrupt.  Do we ensure the download happens before trying `lake serve`?  Do we need the timeout at all?

lovettchris: Yep this is the part I'm most unsure about, what I really want is to be able to ask Lake "can you run a server here" and have it answer "yes".  The timeout is needed because "lake serve" runs forever :-) I think the install should not happen because I have already called "lean --version" in the extension activate `const versionInfo = await installer.testLeanVersion();` and any required install should have happened there (with nice notification progress, etc).

lovettchris: Ooo, I wonder if this can be the test that I do instead: `lake env lean --version` this one terminates and fails when there is no lakefile.lean and it fails when the lakefile.lean  contains a syntax error...

lovettchris: It works, so I'm moving to that approach, much better than having to kill the "lake serve" process.

gebner: That's not going to be better than checking for the existence of `lakefile.lean`.  What I'm worried about is versions of lake that don't have `lake serve` (it's a somewhat recent addition).

It should be enough to just close the standard input of `lake serve`.  At least locally, `lake server </dev/null` terminates promptly.

lovettchris: Ok, I'll see if I can figure out how to do that in typescript...

lovettchris: Rats, lake returns an error code in that case, so now I have to check for the english message 'Watchdog error: Cannot read LSP request: Stream was closed\n'.  Yuk.

lovettchris: I pushed the changes.

Kha: > That's not going to be better than checking for the existence of `lakefile.lean`. What I'm worried about is versions of lake that don't have `lake serve` (it's a somewhat recent addition).

Just to mention an alternative, we could also ask @tydeu to tag a new version for `serve` and check for that.

gebner: > we could also ask @tydeu to tag a new version for serve and check for that.

I think that would be the best solution.  It's also more future-proof.  (Mac just needs to make sure the lake version is bumped when the `lake serve` API changes.)

lovettchris: I'm not following?  I went with the idea of closing the stdin stream, which terminates `lake serve`.  The only thing I'd like to improve is make `lake` or `lean` not spit out out error when stdin is closed.  It should just quietly terminate because there's nothing more to read and return 0.

lovettchris: @gebner , I'm calling this done for now, please reopen if you disagree, thanks.

gebner: > I'm not following?

I think we should parse the output of `lake version`.  This also allows us to adapt if in the future the command-line syntax of `lake serve` changes (e.g. dropping the `--`, or allowing other arguments, etc.).

lovettchris: Sorry I'm confused, are you asking me to not use `lake serve` and go back to my previous code using `lake env lean --version` ?  This is what is currently checked into this PR branch:
```typescript
 const expectedError = 'Watchdog error: Cannot read LSP request: Stream was closed\n';
            const rc = await testExecute(cmd, ['serve'], folder.fsPath, this.outputChannel, true, expectedError);
            if (rc !== 0) {
                const failover = 'Lake failed, using lean instead.'
                console.log(failover);
                if (this.outputChannel) this.outputChannel.appendLine(failover);
                useLake = false;
            }
```

gebner: Sorry, what I meant was neither of the two.  My last suggestion was to parse the output of `lake --version`.  If it's >=4.0.0 (which Mac will hopefully release soon), then use `lake serve`, otherwise fall back to `lean --server`.

gebner: Note that `lake env lean --version` is different from `lake --version`.

lovettchris: Ok, but why? The code I have now works...

lovettchris: Ok, I pushed this change.  See [this commit](https://github.com/lovettchris/vscode-lean4/commit/aaf10145197c2495ea6651e14ff210a353fde745)

tydeu: @lovettchris you should take into account the suffix for version checking `3.0.0` =/= `3.0.0-pre`. My plan is to bump the officially bump the version to `3.0.0` this weekend (which will be the first version with official `lake serve` support).

lovettchris: Ok, I added parsing of the pre-release label in [Version.ts](https://github.com/leanprover/vscode-lean4/blob/735e54b18267ca87d8a8fffeb409e612b30de9f2/vscode-lean4/src/utils/version.ts) and switch the check in leanclient.ts to "3.0.0" and tested that it skips "lake serve" on my machine due to the fact that it finds version "3.0.0-pre".

vscode-lean4/src/leanclient.ts
```diff
@@ -94,33 +100,96 @@ export class LeanClient implements Disposable {
...
+        // current workspace.  The LanguageClient is much slower because it does
+        // 5 retries and everything... but this testExecute implementation is
+        // not perfect either... would be nice to have a "canExecute" command line
+        // option or something...
+        if (useLake) {
+            // First check we have a version of lake that supports "lake serve"
+            const lakeVersion = await batchExecute(cmd, ['--version'], folder.fsPath, null);
+            const v4 = new Version('3.0.0')
+            const actual = new Version(this.removePrefix(lakeVersion, 'Lake version '))
+            if (actual.compare(v4) >= 0) {
```
gebner, line 163: No need to reinvent the wheel here.

```suggestion
            const actual = this.removePrefix(lakeVersion, 'Lake version ')
            if (semver.gte(actual, '3.0.0')) {
```

gebner (commented):
Otherwise lgtm.  Let's wait for the lake version bump to land in lean and check that the integration works as expected, then merge.

vscode-lean4/src/leanclient.ts
```diff
@@ -94,33 +100,96 @@ export class LeanClient implements Disposable {
...
+                const expectedError = 'Watchdog error: Cannot read LSP request: Stream was closed\n';
+                const rc = await testExecute(cmd, ['serve'], folder.fsPath, this.outputChannel, true, expectedError);
+                if (rc !== 0) {
+                    const failover = 'Lake failed, using lean instead.'
+                    console.log(failover);
+                    if (this.outputChannel) this.outputChannel.appendLine(failover);
+                    useLake = false;
+                }
```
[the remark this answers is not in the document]
lovettchris: No it can still fail if there is a syntax error in the `lakefile.lean` file.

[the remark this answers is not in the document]
gebner: Ah, reported at https://github.com/leanprover/lake/issues/49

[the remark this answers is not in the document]
lovettchris: Ok, cool, I'll remove this code when that issue is resolved.  So can we merge this one now?

lovettchris:
> Otherwise lgtm. Let's wait for the lake version bump to land in lean and check that the integration works as expected, then merge.

When will that be, I don't see any Lake PR that is removing the "pre" specialDesc label...

tydeu:
@lovettchris my hope is to get it done this weekend, now that I have finished with other obligations that had been occupying my time.

lovettchris:
@gebner I merged this with master, but it was a nasty merge, so will need thorough testing.

lovettchris:
@gebner - another tricky merge with master.  We both changed ensureClient to take a Uri, but the new version also returns a boolean telling caller if this was a new client or a cached client, needed in the implementation of installChanged.  So anyway, I tested it on a multi-workspace situation and it seems to work, the latest build of Lean now has "Lake 3.0.0" and so lake is now showing up in the ps trace, and you can mix and match, here's a multi-workspace with one using older version of lean so it's running "lean --server" directly, while the other is using Lake.  Nice.

```
Code.exe: "Code.exe" .
  Code.exe: "Code.exe" --ms-enable-electron-run-as-node --nolazy --inspect-brk=53727 "resources\app\out\bootstrap-fork" --type=extensionHost --skipWorkspaceStorageLock
    cmd.exe: C:\WINDOWS\system32\cmd.exe /d /s /c "lean --server d:\Temp\lean_examples\test"
      lean.exe: lean  --server d:\Temp\lean_examples\test
        lean.exe: "leanprover--lean4---nightly-2022-01-20\bin\lean.exe" --server d:\Temp\lean_examples\test
          lean.exe: leanprover--lean4---nightly-2022-01-20\bin\lean.exe "--worker" "file:///d%3A/Temp/lean_examples/test/Test.lean"
    cmd.exe: C:\WINDOWS\system32\cmd.exe /d /s /c "lake serve -- d:\Temp\lean_examples\Foo"
      lake.exe: lake  serve -- d:\Temp\lean_examples\Foo
        lake.exe: "leanprover--lean4---nightly\bin\lake.exe" serve -- d:\Temp\lean_examples\Foo
          lean.exe: leanprover--lean4---nightly\bin\lean.exe "--server" "d:\Temp\lean_examples\Foo"
            lean.exe: leanprover--lean4---nightly\bin\lean.exe "--worker" "file:///d%3A/Temp/lean_examples/Foo/Main.lean"
```

lovettchris:
I am seeing one interesting warning when I remove a folder from my multi-folder workspace:

```
Watchdog error: Got `shutdown` request, expected an `exit` notification
```
Do you know what LSP is talking about?

Vtec234:
@lovettchris : according to the LSP spec, after the client sends a `shutdown` request it should follow that up with an `exit` notification but not anything else inbetween. So I'm guessing the client actually does send something else, which a spec violation. It would be better to print out the actual message received [here](https://github.com/leanprover/lean4/blob/c30380e2fa0fcaed60ef79c97e6724b8763633a0/src/Lean/Server/Watchdog.lean#L714). We could also make the server ignore this, but I rather keep it spec-compliant and fix the client.

vscode-lean4/src/leanclient.ts
```diff
@@ -91,34 +97,99 @@ export class LeanClient implements Disposable {
...
+            }
+        }
+        if (!rootPath) {
+            const workspaceFolders = workspace.workspaceFolders;
+            if (workspaceFolders && workspaceFolders.length > 0) {
+                rootPath = workspaceFolders[0].uri;
+            }
+        }
+        if (!rootPath) {
+            return Uri.parse('file:///' + cwd());
```
gebner, line 118:
```suggestion
            return Uri.file(cwd());
```

vscode-lean4/src/utils/leanpkg.ts
```diff
@@ -95,13 +123,29 @@ export class LeanpkgService implements Disposable {
...
 
     dispose(): void {
         for (const s of this.subscriptions) { s.dispose(); }
     }
 
+    private async handleLakeFileChanged(uri : Uri) {
+        // Note: just opening the file fires this event sometimes which is annoying, so
+        // we compare the contents just to be sure and normalize whitespace so that
+        // just adding a new line doesn't trigger the prompt.
+        const contents = await this.readLakeFile(uri);
```
gebner, line 141: This will read the wrong file if you have a lakefile.lean in some subdirectory (e.g. a dependency or an example or a test case).

tydeu:
@lovettchris A general note: In `elan` paralance, the toolchain path is to the root directory (sysroot) of the toolchain (e.g., `leanprover--lean4---nightly`). So the executables should be located at `<toolchainPath>/bin`.

vscode-lean4/src/utils/leanpkg.ts
lovettchris: Fixed.

lovettchris:
> @lovettchris A general note: In `elan` paralance, the toolchain path is to the root directory (sysroot) of the toolchain (e.g., `leanprover--lean4---nightly`), not the `bin` directory of the toolchain. So VSCode should ideally be using the executables at `<toolchainPath>/bin`.

Done.  But in testing this things are going a bit wonky, I tried debugging it but got stuck, so I would love your help.  It seems now that when I use SelectToolchain to select a different version, or a custom path, the lean server restarts no problem, I get "running" state and the pstree shows the right version of lean.exe is running, but any change to a .lean source file returns an error saying:

```
Watchdog error: cannot find open document 'file:///d%3A/Temp/lean_examples/Foo/Main.lean'
[Info  - 8:57:59 PM] Connection to server got closed. Server will restart.
Watchdog error: cannot find open document 'file:///d%3A/Temp/lean_examples/Foo/Main.lean'
[Info  - 8:58:00 PM] Connection to server got closed. Server will restart.
```

Here are the log files for the newly restarted lean server (started with lake serve):

[logs.zip](https://github.com/leanprover/vscode-lean4/files/8020942/logs.zip)

I have no idea why a newly created lean client things the file is closed.  It is not closed...

lovettchris:
`Update`: there's huge merge conflicts between this PR and  `fix: #115 : Infoview not working for external file` so I'm going to pre-empt that mess, and re-apply the changes in this PR on top of that PR carefully checking every line and perhaps that will solve this weird problem I'm running into here.

lovettchris:
Indeed, re-applying the changes on top of #117 solved the problem.

PR #100, opened 2022-01-11
fix: copyToComment sometimes fails to do anything.

### LINKED ISSUE #99: Sometimes the copy to comment feature from infoview doesn't work
Repro is not easy, but some combination of window activations causes window.visibleTextEditors[0] to fail.  The solution seems to be to use `window.visibleTextEditors.find(e => e.document.languageId === 'lean4');`

lovettchris:
fixes #99

### DIFF
--- a/vscode-lean4/src/infoview.ts
+++ b/vscode-lean4/src/infoview.ts
@@ -389,10 +389,15 @@ export class InfoProvider implements Disposable {
         } else {
             editor = window.activeTextEditor;
             if (!editor) { // sometimes activeTextEditor is null.
-                editor = window.visibleTextEditors[0];
+                editor = window.visibleTextEditors.find(e => e.document.languageId === 'lean4');
             }
         }
-        if (!editor) return;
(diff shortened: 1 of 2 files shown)


vscode-lean4/src/infoview.ts
```diff
@@ -389,10 +389,15 @@ export class InfoProvider implements Disposable {
...
             if (!editor) { // sometimes activeTextEditor is null.
-                editor = window.visibleTextEditors[0];
+                editor = window.visibleTextEditors.find(e => e.document.languageId === 'lean4');
             }
         }
-        if (!editor) return;
+        if (!editor) {
+            // user must have switch away from any lean source file in which case we don't know
+            // what to do here.  TODO: show a popup error?  Or should we use the last uri used in
+            // sendPosition and automatically activate that editor?
```
gebner, line 398: The other question is what should happen if the user has multiple visible editors with Lean files (like I often do).

PR #103, opened 2022-01-14
Provide definition at word position

larsk21:
Provide definition(, declaration, type definition, and references) at word position instead of cursor position, if available.

This is consistent with `provideDocumentHighlights`, i.e. the provided definition will be for the word that received a highlight based on the current cursor position. It is also consistent with the error reporting of VSCode regarding definitions, i.e. if no definition could be found, VSCode will show `No definition found for 'x'` with `x` being the word at the word range of the cursor position.

The reason why `wordRange.start` is used for the position, is that the Lean server provides the definition based on the given position with an inclusive start and exclusive end position for words.

resolves https://github.com/leanprover/lean4/issues/767


gebner:
I think the cure is worse than the disease here.  Silently changing the position of the LSP request only leads to even more intransparent error modes:
```
macro "f#" x:term: term => `("f" ++ toString $x)
def abc := 42
#check f#abc
        --^ go-to-definition goes to f# and not abc
```

I agree that the vscode "no definitions found" message is horrible, but that should be reported to the vscode bug tracker, not patched around here.

Kha:
Hmm, isn't VS Code already changing the position of the request when I Ctrl+click in the middle of an ident? Using its own definition of a word presumably?

Kha:
Ah no, I suppose it isn't :smile:

larsk21:
The found definition in the example is also wrong because VSCode doesn't know that `#` splits words (`#` is not excluded in the `wordPattern` in `languageConfiguration.json`).

gebner:
> Hmm, isn't VS Code already changing the position of the request when I Ctrl+click in the middle of an ident?

To the best of my knowledge, none of the editors we support change the position of definition queries.

> The found definition in the example is also wrong because VSCode doesn't know that # splits words

I really don't want to commit to perfectly emulating the identifier parser in vscode (`#` isn't the only difference).  Besides, with a custom parser you can break the word pattern in any way you want.

> But I agree that there are situations in which the client doesn't have enough information to correctly filter out words in macros.

I'm not following.  This issue is not about filtering, it's about a misleading message from vscode?

larsk21:
> To the best of my knowledge, none of the editors we support change the position of definition queries.

I don't know an example either. However, Microsoft suggests that a client may use a different position instead of the cursor position when the user selects a range (https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams). While not the same, it's a use case for changing the position.

> I really don't want to commit to perfectly emulating the identifier parser in vscode (# isn't the only difference). Besides, with a custom parser you can break the word pattern in any way you want.

It's not about re-implementing the parser, vscode calculates the word range based on the `wordPattern` defined in this extension. Simply adding `#` to it would make your example work.

> I'm not following. This issue is not about filtering, it's about a misleading message from vscode?

Initially this issue was about double-clicking a word and using "go to definition", which currently does not work, because vscode places the cursor at the end of the word and the Lean server uses an exclusive end position for finding the word at the given position. I then thought reusing the code responsible for the highlights (part of this extension, uses the word pattern) would be a more consistent behavior, because the found definition would be for the part of the code that received the highlight.

gebner:
> It's not about re-implementing the parser, [...] Simply adding # to it would make your example work.

And then I just come back with the next example.  This is a cat-and-mouse game until the regex matches what Lean does internally.  And even then it only works for term parsers.

One of the cool features in Lean 4 is that you can extend the syntax.  I really don't want to hamstring that by hardcoding an incomplete part of the lexical syntax of one syntax category in a specific Lean version.

> I then thought reusing the code responsible for the highlights (part of this extension, uses the word range) would be a more consistent behavior, because the found definition would be for the part of the code that received the highlight.

Making the behavior more consistent is certainly a good thing.  However that part of the code will eventually move to the server anyhow so that it can show the actual references and not just variables with the same name.  I've only put in that fallback for now because the server claims it supports highlights, but then doesn't return them (except for do<->return).

larsk21:
> Making the behavior more consistent is certainly a good thing. However that part of the code will eventually move to the server anyhow so that it can show the actual references and not just variables with the same name. I've only put in that fallback for now because the server claims it supports highlights, but then doesn't return them (except for do<->return).

Okay, that makes sense. I guess I will implement it in the server then.

PR #104, opened 2022-01-14
Add multi folder support

lovettchris:
See in this [vscode multi-server sample](https://github.com/microsoft/vscode-extension-samples/tree/main/lsp-multi-server-sample) and see: https://github.com/microsoft/vscode/wiki/Adopting-Multi-Root-Workspace-APIs

Here's a screenshot showing a multi-folder workspace open with 2 different versions of lean specified in the lean-toolchain of each folder.   It fires up 2 different `lean --server` instances, one for each folder, and seamlessly switches between them as you edit files from each folder.   The InfoView expects an "active client changed" event and "reconnects" that client to the infoview as the user switches between active workspaces.  

image

Here's the process tree you will see for serving both these workspaces, 2 lean --servers with their separate workers.

image


lovettchris:
@gebner , thanks for the bug report, I suspect I have to filter out sendNotification calls for files opened in a different workspace now, see [this commit](https://github.com/leanprover/vscode-lean4/pull/104/commits/2336a6475e6c9497a1cd794a8369b2c0d75369b8).  With this change however, I'm getting some strange LSP errors, like this one:
image

lovettchris:
> @gebner Instead we should route the LSP requests (and notification subscriptions) depending on the document URI. That is, if the infoview makes an LSP request, we look at the the URI (which should maybe become an explicit argument of sendClientNotification and co.) and then send it to the responsible LSP server. And the serverRestarted event should probably send all connected servers then.

Done.  See [InfoView is now mult-client aware](https://github.com/leanprover/vscode-lean4/pull/104/commits/50b301a92dfb9ce9d8612c6a48e8ade2a9f574d4).   `lean4.restartServer` still only restarts the client associated with the active document you are editing, but that's easy to change if we want.

gebner:
Great!

vscode-lean4/src/leanclient.ts
```diff
@@ -127,9 +128,22 @@ export class LeanClient implements Disposable {
...
             }
         }
+
+        const documentSelector: DocumentFilter = {
+            language: 'lean4'
+        }
+
+        if (this.workspaceFolder){
+            documentSelector.scheme = 'file'
+            documentSelector.pattern = `${this.workspaceFolder.uri.toString()}/**/*`
```
gebner, line 138: BTW, this was the wrong pattern. :cry: 

https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/VS.20Code.200.2E0.2E60.20-.20Manual.20refresh.20necessary

lovettchris: Interesting.  Weird that the toString() version seemed to be working fine on my machine.   On Windows your change to 'fsPath' produces a weird windows path of "d:\temp\lean_examples\foo/**/*" which seemed wrong to me, but I guess it still works.  Thanks for fixing it.

PR #108, opened 2022-01-26
fix: reveal file location should move the keyboard focus to the text editor

### LINKED ISSUE #71: "reveal file location" should move the keyboard focus to the text editor
Would be nice if when you click the `reveal file location` icon if it also switch to active editor back to the lean program.

lovettchris:
fixes #71 

A simple window.showTextDocument does the trick.


Vtec234 (approved):
LGTM!

PR #109, opened 2022-02-01
fix: some more -32801 errors are reported in the console log.

lovettchris:
Minor cleanup of console.log errors.  Do we want to implement a retry in the lazy fetch of updated diagnostics or just wait for the next edit?

### DIFF
--- a/lean4-infoview/src/infoview/messages.tsx
+++ b/lean4-infoview/src/infoview/messages.tsx
@@ -100,7 +100,8 @@ export function AllMessages({uri: uri0}: { uri: DocumentUri }) {
                 return await getInteractiveDiagnostics(rs, { uri: uri0, line: 0, character: 0 }) || [];
             } catch (err: any) {
                 if (err?.code === -32801) {
-                    // Document has been changed since we made the request, try again later?
+                    // Document has been changed since we made the request.
+                    // This can happen while typing quickly, so server will catch up on next edit.
                 } else {
                     console.log('getInteractiveDiagnostics error ', err)
                 }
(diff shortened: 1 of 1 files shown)


Vtec234 (approved):
Looks good. If I am imagining the codepath right, a retry is not needed since if the document has changed, the server will eventually send updated diagnostics and the fetch of interactive diagnostics will be executed again in the infoview in response to that. But if you can, please do test that this works.

lovettchris:
Ok, I added a print statement like this:
```
if (err?.code === -32801) {
                    // Document has been changed since we made the request, try again later?
                    console.log("retry?");
                } else {
                    console.log('getInteractiveDiagnostics error ', err)
                }
```
then look at it, and when typing quickly in the editor I see a whole bunch of them come through, so yeah, I think this is the right fix.

lean4-infoview/src/infoview/messages.tsx
```diff
@@ -99,7 +99,12 @@ export function AllMessages({uri: uri0}: { uri: DocumentUri }) {
             try {
                 return await getInteractiveDiagnostics(rs, { uri: uri0, line: 0, character: 0 }) || [];
             } catch (err: any) {
-                console.log('getInteractiveDiagnostics error ', err)
+                if (err?.code === -32801) {
+                    // Document has been changed since we made the request.
+                    // This can happen while typing quickly, so server will catch up on next edit.
```
Vtec234, line 104: I would also explain the infoview codepath:
```suggestion
                    // Document has been changed since we made the request. This can happen
                    // while typing quickly. When the server catches up on next edit, it will
                    // send new diagnostics to which the infoview responds by calling
                    // `getInteractiveDiagnostics` again.
```

PR #117, opened 2022-02-08
fix: #115 : Infoview not working for external file

### LINKED ISSUE #115: Infoview not working for external file.
Infoview doesn't show the context in files located outside the working directory and sub directories.

### How to reproduce
```sh
$ lake new minimal
$ cd minimal
$ code .
```
open Main.lean
Infoview shows the context
open ~/.elan/toolchains/leanprover--lean4---nightly-2022-02-05/src/lean/Init/System/IO.lean
Lean infoview doesn't show the context

### Versions
lean version : leanprover/lean4:nightly-2022-02-05
vscode-lean4 version : 0.0.63
last  version without the bug : 0.0.59

### Additional informations
Runtime status shows an uncaught error when opening IO.lean : Cannot read property 'filter' of undefined

lovettchris:
Fixes #115

Turns out the work to support "untitled" files broke the "ad hoc" files.   Now I've tested the following scenarios:
- start vs code from start menu (not from a folder), File/Open file..., works (let's call this an "adhoc" file).
- start vs code from start menu (not from a folder), File/New file, set Lean4... works (let's call this an "untitled" file).
- start vs code from folder, open files in that folder, works!
   - now File/Open file... from somewhere else, also works!
   - now File/New file & set Lean4, also works!  
 
This last option now has 3 LeanClients running, one for the folder, and one for the adhoc package root, and one for 'untitled' files.  

For "adhoc" files the extension searches upwards in the directory hierarchy for a Lean package root and creates a new LeanClient for that root.

### DIFF
--- a/vscode-lean4/src/utils/clientProvider.ts
+++ b/vscode-lean4/src/utils/clientProvider.ts
@@ -167,13 +167,13 @@ export class LeanClientProvider implements Disposable {
     }
 
     async getLeanVersion(uri: Uri) : Promise<LeanVersion> {
-        let [workspaceFolder, folderUri, packageFileUri] = findLeanPackageRoot(uri);
+        const [workspaceFolder, folderUri, packageFileUri] = findLeanPackageRoot(uri);
         const path = folderUri?.toString();
         if (this.versions.has(path)){
             return this.versions.get(path);
         }
         let versionInfo : LeanVersion = null;
-        if (uri.scheme == 'untitled'){
+        if (uri.scheme === 'untitled'){
             versionInfo = { version: '4', error: null };
         } else {
             versionInfo = await this.installer.testLeanVersion(folderUri);
@@ -183,7 +183,7 @@ export class LeanClientProvider implements Disposable {
     }
 
     async ensureClient(uri: Uri, versionInfo: LeanVersion | null): Promise<LeanClient> {
-        let [workspaceFolder, folderUri, packageFileUri] = findLeanPackageRoot(uri);
+        const [workspaceFolder, folderUri, packageFileUri] = findLeanPackageRoot(uri);
 
         const path = folderUri?.toString();
         let  client: LeanClient = null;
--- a/vscode-lean4/src/utils/clientProvider.ts
+++ b/vscode-lean4/src/utils/clientProvider.ts
@@ -99,7 +99,7 @@ export class LeanClientProvider implements Disposable {
     }
 
     private getVisibleEditor(uri: Uri) : TextEditor | null {
-        var path = uri.toString();
+        const path = uri.toString();
         for (const editor of window.visibleTextEditors) {
             if (editor.document.uri.toString() === path){
                 return editor;
@@ -135,7 +135,7 @@ export class LeanClientProvider implements Disposable {
             return;
         }
 
-        let client = await this.ensureClient(document.uri, null);
+        const client = await this.ensureClient(document.uri, null);
         await client.openLean4Document(document)
     }
 
--- a/vscode-lean4/src/utils/clientProvider.ts
+++ b/vscode-lean4/src/utils/clientProvider.ts
@@ -1,8 +1,10 @@
-import { Disposable, OutputChannel, workspace, TextDocument, commands, window, EventEmitter, Uri, languages } from 'vscode';
+import { Disposable, OutputChannel, workspace, TextDocument, commands, window, EventEmitter, Uri, languages, TextEditor } from 'vscode';
 import { LocalStorageService} from './localStorage'
 import { LeanInstaller, LeanVersion } from './leanInstaller'
 import { LeanClient } from '../leanclient'
 import { LeanFileProgressProcessingInfo, RpcConnectParams, RpcKeepAliveParams } from '@lean4/infoview-api';
+import * as path from 'path';
+import { urlToHttpOptions } from 'url';
 
 // This class ensures we have one LeanClient per workspace folder.
 export class LeanClientProvider implements Disposable {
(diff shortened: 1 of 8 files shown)


vscode-lean4/src/utils/clientProvider.ts
```diff
@@ -103,29 +116,68 @@ export class LeanClientProvider implements Disposable {
...
+            if (document.languageId === 'lean') {
+                // Only change the document language for *visible* documents,
+                // because this closes and then reopens the document.
+                await languages.setTextDocumentLanguage(document, 'lean4');
+
+                // setTextDocumentLanguage triggers another didOpenEditor event,
+                // and we want that event callback to be the one that calls
+                // ensureClient, so we bail here so we don't try and do it twice
+                // on the same document.
+                return;
```
gebner, line 145: I vaguely remember that I didn't get the second didOpenEditor event when I tried it last time.  (That's why the condition was fall-through.)

vscode-lean4/src/utils/clientProvider.ts
```diff
@@ -156,12 +219,10 @@ export class LeanClientProvider implements Disposable {
                 // ignore workspaces that belong to a different version of Lean.
                 console.log(`Lean4 extension ignoring workspace '${folderUri}' because it is not a Lean 4 workspace.`);
                 this.pending.delete(path);
+                this.clients.delete(path);
```
gebner, line 222: Should we dispose the client object in this case?

gebner:
> In this case should we walk up the directory structure and find the Leanpkg.lean in lib\lean\src

If the user opens that `IO.lean` file directly, then this would be the ideal behavior.  (Maybe look for something else than `Leanpkg.lean` though, because that file no longer exists in recent nightlies.)

When the user opens that file via go-to-definition, it would also be reasonable to associate it with the lean server from the other file.  An argument for reusing servers is that find-references only works across the same server.

lovettchris:
> Maybe look for something else than `Leanpkg.lean` though

Right I meant to say I would be looking for the usual package files `leanpkg.toml` or `lean-toolchain`..

vscode-lean4/src/utils/clientProvider.ts
lovettchris: seems reliable now, but we can keep an eye on if, if you find a repro for when the second open event does not happen let me know...

vscode-lean4/src/utils/clientProvider.ts
lovettchris: Good catch, fixed.

lovettchris:
@gebner ok, I did a big cleanup on when and how we walk directory looking for Lean package roots, they are now stateless helper functions.  This also cleaned up some redundant calls to testLeanVersion that happen on first Lean extension activation - so the first lean file now gets to running LSP state a bit faster than before. I also implemented the adhoc directory walk which seems to be working nicely.  For example, I do an adhoc open of a file deep in mathlib4 repo and it correctly finds the math4 root and opens only one LeanClient for all of mathlib4 which is the right thing to do.  So I think this PR is in good state now and ready to merge.

vscode-lean4/src/utils/projectInfo.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+async function readLeanVersionFile(packageFileUri : Uri) : Promise<string> {
+    const url = new URL(packageFileUri.toString());
+    const tomlFileName = 'leanpkg.toml';
+    if (packageFileUri.path.endsWith(tomlFileName))
+    {
+        return new Promise<string>((resolve, reject) => {
+            if (fs.existsSync(url)) {
+                fs.readFile(url, { encoding: 'utf-8' }, (err, data) =>{
```
gebner, line 115: Please stick to either async or sync.  Reading a small local file is fast enough that sync would be okay, but mixing them is not really helpful (since it could still block).  The `exists` operation is also deprecated (and shouldn't be used together with readFile due to the race condition where the file is deleted between the exists and readFile calls).

So either:
```typescript
try {
  fs.readFileSync(url, {encoding: 'utf-8'}).toString()
  // ...
} catch (err) {
  // could check for err.code === 'ENOENT'
  return null;
}
```

Or async, which is just as easy:
```typescript
try {
  (await fs.promises.readFile(url, {encoding: 'utf-8'})).toString()
  // ...
} catch (err) {
  // could check for err.code === 'ENOENT'
  return null;
}
```

I would also try to drop the duplicate `existsSync` check in `readLeanVersion`.

vscode-lean4/src/utils/projectInfo.ts
```diff
@@ -0,0 +1,146 @@
+import * as fs from 'fs';
+import { URL } from 'url';
+import { Uri, workspace, WorkspaceFolder } from 'vscode';
+
+// Find the root of a Lean project and return the Uri for the package root and the Uri
+// for the 'leanpkg.toml' or 'lean-toolchain' file found there.
+export function findLeanPackageRoot(uri: Uri) : [WorkspaceFolder, Uri,Uri] {
+    if (!uri) return [null, null, null];
```
gebner, line 8:
```suggestion
export function findLeanPackageRoot(uri: Uri) : [WorkspaceFolder | null, Uri | null, Uri | null] {
    if (!uri) return [null, null, null];
```

I'm surprised that this passes the type checker.  Another more self-explanatory return type might be `{workspaceFolder?: WorkspaceFolder, leanpkgToml?: Uri, leanToolchain?: Uri}`.

vscode-lean4/src/utils/projectInfo.ts
lovettchris: Generally I agree with you but the only reason I didn't use exists() here is because it is deprecated for some reason.  In general I prefer all IO to be async as hard drives and network drives can have random very long delays depending on what's going on and I hate apps that force me to wait for them to do synchronous IO... but if you feel strongly about it I can change this to readFileSync - it will have ripple effects all over removing async behavior up the chain.

PR #118, opened 2022-02-10
feat: use lake serve, take 2

lovettchris:
See [Zulip Discussion](https://leanprover.zulipchat.com/#narrow/stream/147302-lean4-dev/topic/Expose.20--load-dynlib.20to.20Lean.3F)

This PR is dependent on #117 

Proposal, replace the following setting:
image

With:
image

Then the rules for using `lake serve` would be:
1. is 'lake' or 'lake.exe' found in the toolchainPath, If not don't use `lake`
2. is there a `lakefile.lean` in the current workspace root?  If not don't use `lake`
3. If the above are both good, use `lake serve` with optional +version if there's a workspace override in effect.
4. If `lake serve` fails fall back on `lean --server`.

This PR also fixes a bug introduced by support for "multi-workspace" namely that the "currentVersion" maintained in LeanpkgService needs to be "per-workspace".

### DIFF
--- a/vscode-lean4/src/leanclient.ts
+++ b/vscode-lean4/src/leanclient.ts
@@ -17,10 +17,16 @@ import {
     State
 } from 'vscode-languageclient/node'
 import * as ls from 'vscode-languageserver-protocol'
-import { executablePath, addServerEnvPaths, serverArgs, serverLoggingEnabled, serverLoggingPath, getElaborationDelay } from './config'
+import { toolchainPath, addServerEnvPaths, serverArgs, serverLoggingEnabled, serverLoggingPath, getElaborationDelay, lakeEnabled } from './config'
 import { assert } from './utils/assert'
 import { LeanFileProgressParams, LeanFileProgressProcessingInfo } from '@lean4/infoview-api';
 import { LocalStorageService} from './utils/localStorage'
+import { batchExecute, testExecute } from './utils/batch'
+import { cwd } from 'process'
+import * as fs from 'fs';
+import { URL } from 'url';
+import { join } from 'path';
+import { SemVer } from 'semver';
 
 const escapeRegExp = (s: string) => s.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 
@@ -37,7 +43,7 @@ export function getFullRange(diag: Diagnostic): Range {
 export class LeanClient implements Disposable {
     running: boolean
 	private client: LanguageClient | undefined
-    private executable: string
+    private toolchainPath: string
     private outputChannel: OutputChannel;
     private storageManager : LocalStorageService;
     private workspaceFolder: WorkspaceFolder;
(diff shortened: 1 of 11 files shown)


lovettchris:
@gebner, Rats this PR is now exhibiting the problem I thought I fixed by manually re-applying all those changes.  Select Toolchain restart is not working right, the newly created LanguageClient that comes up to "Running" state without any errors, dies and restarts each time I edit the file.  Doing a full "developer reload" solves the problem...  I thought [this important commit](https://github.com/leanprover/vscode-lean4/pull/118/commits/c9da2eb786038125e84057718685c2a00899bcfe) would fix it but it doesn't.  Can you help debug this? Seems like a weird LSP issue...

lovettchris:
@Gabriel, oh, interesting, I just confirmed this bug is also in `master` !  So it's not caused by this PR.

vscode-lean4/src/leanclient.ts
```diff
@@ -98,30 +104,71 @@ export class LeanClient implements Disposable {
...
+            if (actual.compare('3.0.0') >= 0) {
+                const expectedError = 'Watchdog error: Cannot read LSP request: Stream was closed\n';
+                const serveOptions =  version ? ['+' + version, 'serve'] : ['serve'];
+                const rc = await testExecute(executable, serveOptions, this.folderUri?.fsPath, this.outputChannel, true, expectedError);
+                if (rc !== 0) {
+                    const failover = 'Lake failed, using lean instead.'
+                    console.log(failover);
+                    if (this.outputChannel) this.outputChannel.appendLine(failover);
+                    useLake = false;
+                }
```
gebner, line 142: I think we can drop this check now.  `lake serve` falls back to `lean --server` now if the `lakefile.lean` doesn't compile.

lovettchris: Do you know what version of lake does that?  I tried this version: leanprover/lean4:nightly-2022-02-10, with a syntax error in the lakefile.lean and I get these retry errors and no lean server running:
```
client starting
vscode-lean4/src/leanclient.ts:257
client has stopped or it failed to start
vscode-lean4/src/leanclient.ts:262
client starting
vscode-lean4/src/leanclient.ts:257
client has stopped or it failed to start
vscode-lean4/src/leanclient.ts:262
client starting
vscode-lean4/src/leanclient.ts:257
client has stopped or it failed to start
vscode-lean4/src/leanclient.ts:262
client starting
vscode-lean4/src/leanclient.ts:257
client has stopped or it failed to start
vscode-lean4/src/leanclient.ts:262
client starting
vscode-lean4/src/leanclient.ts:257
client has stopped or it failed to start
vscode-lean4/src/leanclient.ts:262
```

gebner: Ohh, apparently the bundled lake hasn't been updated since https://github.com/leanprover/lake/pull/52 was merged.  Which means we should probably keep this check for now (and gate it on the lake version in the future).

Kha: Since this was the only commit not merged upstream, it felt safe to do so now. /cc @tydeu

lovettchris:
@gebner  I copied your info to a new issue https://github.com/leanprover/vscode-lean4/issues/125 since this thread is now closed.

PR #123, opened 2022-02-12
fix #107 : fix wrong pre-wrap value

### LINKED ISSUE #107: Better line wrapping of long snippets in the InfoView
The long horizontal scrolling in this example is annoying, we should line wrap:
image

In Lean 3 we had some fancy CSS that did the formatting (i.e., indenting and inserting newlines at the appropriate places). We also used it for the docs: https://leanprover-community.github.io/mathlib_docs/measure_theory/covering/besicovitch_vector_space.html#besicovitch.exists_good_%CE%B4 (try resizing the browser window).

This webpage is perfect, as it provides client side live resizing (rather than formatting to some fixed width in the LSP). Notice the block `{τ : ℝ}` wraps as a chunk, and even much bigger things wrap as a block like `[normed_group E]` and so on. We should copy this nice CSS implementation to the InfoView.

ydewit:


Vtec234:
Oops! Thanks.

PR #127, opened 2022-02-14
fix: adhoc and untitled files were hitting some null reference exceptions

lovettchris:
and not iterable exceptions in the results from findLeanPackageRoot and findLeanPackageVersionInfo.  I also turned on strictNullChecks and noImplicitAny compilerOptions which revealed & fixed a few more potential bugs.

I tested the following "open file/open folder" scenarios:
1. open random mathlib file (D:\git\leanprover\mathlib4\Mathlib\Data\Array\Basic.lean) and not open folder.  It installs the mathlib version of lean (currently leanprover/lean4:nightly-2022-01-23) and works.
1. Goto definition from mathlib to Array.push jumps to (C:\users\clovett\.elan\toolchains\leanprover--lean4---nightly-2022-01-23\lib\lean\src\init\prelude.lean) and creates a new LeanClient in (c:\users\clovett\.elan\toolchains\leanprover--lean4---nightly-2022-01-23").
1. Open folder of a local lake package folder, everything works as expected.
1. Open multi-folder workspace from a file named `workspace.code-workspace`.  2 separate LeanClients are created for each folder each running different versions of lean.
1. Open adhoc file that has no inherited lean-toolchain.  LeanClient starts in that folder and works fine with default toolchain.
1. An old leanpkg.toml project pointing to lean version "master" where that is a linked toolchain to a local build (C:\msys64\home\clovett\git\lean4\build\release\stage1) all works, and lean.versionString prints `4.0.0, commit 4fd2eef9f80b5124269dc885317b015c72925bc5`
1. File/New File, set language to lean4 (uses default toolchain)
1. File/New file... with no default toolchain installed, it auto-installs leanprover/lean4:nightly and works.

### Select Toolchain
and I tested the following select toolchain options:
1. open adhoc file with no inherited lean_toolchain, you get the default toolchain.  Then do Select Toolchain and switch to the "master" linked toolchain, and lean server is restarted and the master bits work fine.
1. open a folder with lean-toolchain, and it comes up with that toolchain.  Select Toolchain to switch to "master" works.
1. Now "Reset workspace override..." via the select toolchain menu and lean server is restarted and goes back to the lean-toolchain specified version.

## Edit lean-toolchain file
1. change the version from `leanprover/lean4:nightly-2022-02-13` to `leanprover/lean4:nightly-2022-02-08`.  You get a prompt saying "Lake file configuration changed" with a "Restart lean" button, click the button and it works, version changes to 2022-02-08.
1. delete `lean-toolchain` and reload folder.  It should use your default toolchain.  Then run `echo leanprover/lean4:nightly-2022-02-08 > lean-toolchain` and you will get a prompt saying "Lake file configuration changed" with a "Restart lean" button, click the button and it works, version changes to 2022-02-08.

### DIFF
--- a/vscode-lean4/src/utils/localStorage.ts
+++ b/vscode-lean4/src/utils/localStorage.ts
@@ -11,7 +11,7 @@ export class LocalStorageService {
 
     getLeanPath() : string
     {
-        return this.storage.get<string>('LeanPath', null);
+        return this.storage.get<string>('LeanPath', '');
     }
 
     setLeanPath(path : string) : void
@@ -21,7 +21,7 @@ export class LocalStorageService {
 
     getLeanVersion() : string
     {
-        return this.storage.get<string>('LeanVersion', null);
+        return this.storage.get<string>('LeanVersion', '');
     }
 
     setLeanVersion(path : string) : void
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -42,6 +42,7 @@ module.exports = {
                 "default": "array"
             }
         ],
+        "@typescript-eslint/ban-ts-comment": "off",
         "@typescript-eslint/ban-types": "off",
         "@typescript-eslint/consistent-type-assertions": "error",
         "@typescript-eslint/explicit-member-accessibility": [
--- a/vscode-lean4/src/config.ts
+++ b/vscode-lean4/src/config.ts
@@ -7,9 +7,9 @@ import * as path from 'path';
 
 export function getEnvPath() : string {
     if (process.platform === 'win32') {
-        return process.env.Path
+        return process.env.Path ?? ''
     } else {
-        return process.env.PATH
+        return process.env.PATH ?? ''
     }
 }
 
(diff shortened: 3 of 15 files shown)


vscode-lean4/src/utils/localStorage.ts
```diff
@@ -11,7 +11,7 @@ export class LocalStorageService {
 
     getLeanPath() : string
     {
-        return this.storage.get<string>('LeanPath', null);
+        return this.storage.get<string>('LeanPath', '');
```
gebner, line 14: This took me a second to appreciate.  `''` is falsy in javascript, so all the `if (leanPath) ...` checks take the same path as before.

I'm not sure if `''` is the best value to indicate that no version has been set; but at least it typechecks now.

gebner:
> and not iterable exceptions in the results from findLeanPackageRoot and findLeanPackageVersionInfo. I also turned on strictNullChecks and noImplicitAny compilerOptions which revealed & fixed a few more potential bugs.

Awesome!

The testing plan looks good, I can't think of anything else off the top of my head.

vscode-lean4/src/utils/localStorage.ts
lovettchris: yeah, I think at least it is consistent now, all string properties we get this way default to empty string.

PR #130, opened 2022-02-15
Beginning of a unit test for VS code extension

lovettchris:
This adds some `integration tests` that run during CI build to ensure there are no obvious regressions in critical behavior.  See [readme.md](https://github.com/lovettchris/vscode-lean4/blob/clovett/unittest/vscode-lean4/test/README.md) for summary of the new tests.

Test seems table now passing 3 consecutive runs:
image


test/.vscode/launch.json
```diff
@@ -0,0 +1,30 @@
+// A launch configuration that compiles the extension and then opens it inside a new window
+// Use IntelliSense to learn about possible attributes.
+// Hover to view descriptions of existing attributes.
+// For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+{
```
gebner, line 5: Does the `.vscode` directory do anything if it's not in the repository root?  At least I couldn't find the launch configurations in vscode.

gebner:
> Let me know if you think this is the right direction...

Some general thoughts:
- Infoview tests are finicky.  We have them in the neovim plugin as well and they're a constant source of issues.  You need to wait until the Lean server settles down, then you need to wait until the infoview has finished updated, etc.  This is hard to get right so they often fail due to timeouts.
- We've had a lot of issues recently related to server startup.  If we had automatic tests for the items you checked in #127, then this could be a good start.  Fallback to Lean 3 would also be good to check.
- The tests need to be run in CI.

test/.vscode/launch.json
lovettchris: it works for me if I cd into the "test" folder and run "code .", to run the unit tests I'm also doing "cd test" then "npm run test"

lovettchris: this is mostly just a clone of [helloworld-test-sample](https://github.com/microsoft/vscode-extension-samples/tree/main/helloworld-test-sample) from [working-with-extensions/testing-extension](https://code.visualstudio.com/api/working-with-extensions/testing-extension).

lovettchris:
> * Infoview tests are finicky.  We have them in the neovim plugin as well and they're a constant source of issues.  You need to wait until the Lean server settles down, then you need to wait until the infoview has finished updated, etc.  This is hard to get right so they often fail due to timeouts.

yes indeed, to tackly this I'm adding some "TestApi" machinery that gives a predictable way to scrape the contents of the infoview and some `utils.ts` code to do retries with timeouts to ensure we wait long enough for different events to occur.  Each of my 2 new tests run in 5 and 3 seconds on github CI, so with a test timeout of 60 seconds it should be ok, but we shall see how many timeouts happen in practice.

> * We've had a lot of issues recently related to server startup.  If we had automatic tests for the items you checked in [fix: adhoc and untitled files were hitting some null reference exceptions #127](https://github.com/leanprover/vscode-lean4/pull/127), then this could be a good start.  Fallback to Lean 3 would also be good to check.

Indeed, this was exactly the reason I started this effort.

> * The tests need to be run in CI.

Indeed, which is now is, see [CI test run results](https://github.com/leanprover/vscode-lean4/runs/5222569399?check_suite_focus=true)

test/.vscode/launch.json
Vtec234: It's a little awkward to have to open the `test/` directory as its own workspace. Why not put the launch configs in the top-level `.vscode` configuration?

.github/workflows/on-push.yml
```diff
@@ -3,20 +3,41 @@ name: vscode-lean4 build
...
-      - run: npx lerna bootstrap --ci
-      - run: npm run build
-      - run: npx lerna run --scope=lean4 package
-      - uses: actions/upload-artifact@v2
+      - name: Build
+        run: |
+          npm ci
+          npx lerna bootstrap --ci
+          npm run build
+          npx lerna run --scope=lean4 package
```
Vtec234, line 21: I would put this in its own "Package" step.

vscode-lean4/src/infoview.ts
```diff
@@ -399,13 +417,14 @@ export class InfoProvider implements Disposable {
...
-            webviewPanel.rpc = new Rpc(m => webviewPanel.webview.postMessage(m));
+            webviewPanel.rpc = new Rpc(m => { if (!closed) void webviewPanel.webview.postMessage(m) });
             webviewPanel.rpc.register(this.editorApi);
 
             // Similarly, we can received data from the webview by listening to onDidReceiveMessage.
-            webviewPanel.webview.onDidReceiveMessage(m => webviewPanel.rpc.messageReceived(m))
+            webviewPanel.webview.onDidReceiveMessage(m => { if (!closed) webviewPanel.rpc.messageReceived(m) })
             webviewPanel.api = webviewPanel.rpc.getApi();
             webviewPanel.onDidDispose(() => {
+                closed = true;
```
Vtec234, line 427: I am not familiar with how `let` is scoped in JS. Are you sure setting this will update the state of the closures above? A class member might be clearer.

lean4-infoview-api/src/infoviewApi.ts
```diff
@@ -128,3 +129,9 @@ export interface InfoviewApi {
    */
   requestedAction(action: InfoviewAction): Promise<void>;
 }
+
+export interface TestApi {
```
Vtec234, line 133: Could you please explain what `TestApi` is for? Is this functionality that the infoview itself provides for test purposes? If so, how could `isInfoViewOpen` possibly return `false`? On the other hand if the `TestApi` isn't so much methods on the infoview as much as it is something for extension<->test suite communication, then it should live in the `vscode-lean4` package.

Copying the infoview HTML to clipboard seems like quite wierd functionality, but if it's needed for something in testing, fine by me. `toggleAllMessages` seems to duplicate `requestedAction('toggleAllMessages')` so probably isn't needed?

test/.vscode/launch.json
Vtec234: Sounds resolved, thanks!

lean4-infoview-api/src/infoviewApi.ts
Vtec234: Judging from the `messages.tsx` diff, only `copyHtmlToClipboard` is functionality that actually needs to be added here, so I would put it directly in `InfoviewApi` and put a comment explaining what kind of test it's needed for.

lovettchris: Sure, it is an Api that allows the "test" package to be strongly typed on an interface.  We could program it all using `any` but that's not very nice.  So we have this handy infoview api package which the test can take a dependency on.  The other approach is to have the test depend on vscode-lean4 entirely then we could access all the types defined in there and not need to "funnel" everything through an explicit API, but then again I kinda like being explicit about what the test code needs.  

The Copy HTML is the most generic thing I could think of that allows the tests do look for any thing they want to in the info view so I don't have to keep adding more and more TestApi methods.  toggleAllMessages  is indeed a wrapper on requestedAction('toggleAllMessages'), again in the spirit of being strongly typed and explicit about what the tests are depending on.  Conversely if the test depended on vscode-lean4 it could depend on `InfoProvider` and call that guy's toggleAllMessages method...  So if you want me to drop the TestApi idea and just expose all of vscode-lean4 to the test package then let me know and I can do that.  It probably means eventually the object returned from `activate` will return also return the InfoProvider, DocViewProvider, AbbreviationFeature, LeanClientProvider, etc, and then the tests can go to town with those objects...

Vtec234: Having an explicit `TestApi` sounds like a good idea, I am just not sure it should live in `lean4-infoview` since that package is meant to be independent of VSCode and extensions for other editors may want to test differently. The crux of the issue seems to be that the `test` package does not depend on `vscode-lean4`, otherwise `TestApi` could live there. This is indeed like the setup of `helloworld-test-sample`. On the other hand the Yeoman extension scaffold  (referenced [here](https://code.visualstudio.com/api/working-with-extensions/testing-extension#overview)) places integration tests directly in the extension package and does not bother to create a separate test package. Maybe just doing that would be simpler? The Haskell extension [does it](https://github.com/haskell/vscode-haskell/tree/79456aaaa98b98f9e7b9f690bb18ab2b2838b6e2/test). In particular, I can't see anyone wanting to depend on `vscode-lean4` but not the tests for some reason.

vscode-lean4/src/infoview.ts
lovettchris: Sure that works, I pushed the change.

lean4-infoview-api/src/infoviewApi.ts
lovettchris: But we could rename the `lean4-infoview-api` and call it `vscode-lean4-api` then it wouldn't look so weird...

test/.vscode/launch.json
lovettchris: Well not quite, the debugger doesn't seem to launch the test correctly, the test runs, but the setup in runtests.ts is not working, I get weird errors like this:
```
[d:\git\leanprover\vscode-lean4]: property `version` is mandatory and must be of type `string`
```
and the runtest should be opening the `src\lean\lean3` folder but it is failing to do so... so I could use some help debugging this one...

lean4-infoview-api/src/infoviewApi.ts
Vtec234: Well, what do you think about moving `test` to the `vscode-lean4` package like Haskell's VSC extension? Note that `lean4-infoview-api` is also intended not to depend on VSCode.

lovettchris: I'm not against it, @gabriel, is this what you recommend also?

lean4-infoview-api/src/infoviewApi.ts
```diff
@@ -128,3 +129,9 @@ export interface InfoviewApi {
...
+  isInfoViewOpen() : Promise<boolean>;
+  copyHtmlToClipboard(): Promise<boolean>;
```
gebner, line 135:
```suggestion
  getHtmlContents(): Promise<string>;
```
I think that this is the operation that we actually want, and then we don't need to do the awkward `sleep(500)` in the test.

lean4-infoview-api/src/infoviewApi.ts
gebner: > Having an explicit TestApi sounds like a good idea, I am just not sure it should live in lean4-infoview since that package is meant to be independent of VSCode and extensions for other editors may want to test differently.

We can also rename it `VscodeTestApi`.  I agree that it doesn't really fit in that package, but there's no technical downside to putting it there either.

> In particular, I can't see anyone wanting to depend on vscode-lean4 but not the tests for some reason.

Nobody is *depending* on the `vscode-lean4` package.  The issue here is whether the tests (and their dependencies) are included in the packaged vscode extension.  But I think webpack does tree shaking already (that's why you don't see the test code in the generated extension.js file).

> `import { InfoProvider } from '../../../../vscode-lean4/src/infoview';`

This works just fine once you remove the extra `tsconfig.json` file in the `test` directory.

If we can have the tests in the `test` subdirectory of the `vscode-lean4` package, then that's the most straightforward solution.  I'd preferably avoid adding new packages if possible; the monorepo setup is already complicated enough as it is.

Vtec234: > The issue here is whether the tests (and their dependencies) are included in the packaged vscode extension. 

They wouldn't be -- this is configurable in `.vscodeignore`.

> If we can have the tests in the test subdirectory of the vscode-lean4 package, then that's the most straightforward solution. I'd preferably avoid adding new packages if possible

I agree and would be happy with this also -- as long as tests aren't in a separate package, I am agnostic as to whether they reside under `vscode-lean4/src/test`, or `vscode-lean4/test`, or `test/`. Maybe either of the first two is still preferable since ultimately only `vscode-lean4` is being integration-tested.

lean4-infoview-api/src/infoviewApi.ts
lovettchris: Sure, this is cleaner.  Note this just pushes the retry loop lower, see https://github.com/lovettchris/vscode-lean4/blob/clovett/unittest/vscode-lean4/src/infoview.ts#L346

lean4-infoview-api/src/infoviewApi.ts
lovettchris: So I'm still messing with this but the "webpack" job is getting in the way.  If I'm going to merge the test in with vscode_lean4, then I need to figure out how to break the compile into 2 stages, a "webpack" job for "src/**/*" and a normal "npm run compile" for "test/**/*" with a different "test/out" folder.  I'm messing with https://stackoverflow.com/questions/37579969/how-to-use-multiple-tsconfig-files-in-vs-code but no luck yet.  Do either of you have experience with this sort of thing?  If I can't get it to work, I will have to use the separate test project.

lovettchris: IF you want to help, perhaps you can experiment with https://github.com/lovettchris/vscode-lean4/tree/clovett/merge_unittest

lovettchris: Oh, hey, I figured it out, the tests are working, they are still in a separate project but it does not need a TestApi.  on reading https://www.typescriptlang.org/docs/handbook/project-references.html more closely I discovered a new build option `--build` is needed on the tsc build jobs instead of `tsc -p ./` it is now `tsc --build` and `tsc --build --watch`, so now the command line tests pass and the vs debugging of tests also works.

lovettchris: But rats, the same `tsc --build` that works on my machine is not working in the CI build...

gebner: If you put the tests in a separate package, then you can just list the extension as a dependency in test/package.json.  There's no need to bother with project references.

lovettchris: Ok, Wojciech helped me out, and we have something working without a separate test package.

test/.vscode/launch.json
lovettchris: this is now resolved and tests are passing and are debuggable.

lean4-infoview-api/src/infoviewApi.ts
lovettchris: Wojciech showed me how to make this synchronous so no retry loop necessary.

lovettchris:
@gebner and @Vtec234  I now have most of https://github.com/leanprover/vscode-lean4/pull/127 implemented.  The only path not tested is the one where user creates a "linked toolchain" to some lean4 bits they compiled on their machine.  I have an idea on how to implement that using the same `Download Lean4 build artifact` trick I use in the [TPIL workflow](https://github.com/leanprover/theorem_proving_in_lean4/blob/master/.github/workflows/ci.yml).

gebner:
1. Please revert the change which converts `abbreviations.json` to `abbreviations.ts`.  The `abbreviations.json` file is used in a number of other projects, including `lean.nvim`, `m17n-lean`, web editors, etc.  These require JSON.  I believe you can import json files directly in webpack.
2. Please remove the debugging logging statements.

lovettchris:
> * Please revert the change which converts `abbreviations.json` to `abbreviations.ts`.  The `abbreviations.json` file is used in a number of other projects, including `lean.nvim`, `m17n-lean`, web editors, etc.  These require JSON.  I believe you can import json files directly in webpack.

Done and I removed the debug output.

PR #131, opened 2022-02-16
fix: improve startup performance of LeanClient

lovettchris:
1. add optimization that skips the lake test on well known versions because it is too slow where the optimization is based on well known lean-toolchain version strings.  
1. Cache `lean --version` check results because we were doing 2 checks on each reload.

### Results

I put a console output that measures full lean server "restart" time following a `Select Toolchain` choice, so I can run release bits to get good measurements, and with this PR it is:

| Version      | Full restart time| Lake? |
| ----------- | ----------- | -------------|
| leanprover/lean4:nightly-2022-02-01| 658 ms | no |
| leanprover/lean4:nightly-2022-02-16 | 823 ms | yes |
| leanprover/lean4:4.0.0-m3 | 656 ms | no |

Previously without this PR the times are:

| Version      | Full restart time | Lake? |
| ----------- | ----------- |-------------|
| leanprover/lean4:nightly-2022-02-01| 1594 ms | no |
| leanprover/lean4:nightly-2022-02-16 | 1526 ms | yes |
| leanprover/lean4:4.0.0-m3 | 1637 ms | no |

So I think this shows a pretty healthy speed up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,3 +308,22 @@ the packages in tandem using Lerna.
 
 **Note:** if breakpoints are not working, try changing one line of code in `function activate` in `extension.ts`, even
 adding a newline seems to work, then press F5.
+
+## Packaging
+
+To publish the extension on the VS Code marketplace you
+should follow [these instructions](https://code.visualstudio.com/api/working-with-extensions/publishing-extension).
+But for a quick summary you need to install this tool
+
+```
+npm install -g vsce
+```
+
+Then run this command
+```
+cd <vscode-lean4-repo>\vscode-lean4
+vsce package
+```
+
+This gives you a .vsix which you can install using the VS Code
+command `Extensions: Install from VSIX...`.
--- a/vscode-lean4/src/leanclient.ts
+++ b/vscode-lean4/src/leanclient.ts
@@ -101,6 +101,7 @@ export class LeanClient implements Disposable {
 
     async restart(): Promise<void> {
         this.restartingEmitter.fire(undefined)
+        const startTime = Date.now()
 
         if (this.isStarted()) {
             await this.stop()
@@ -260,7 +261,8 @@ export class LeanClient implements Disposable {
                 if (s.newState === State.Starting) {
                     console.log('client starting');
                 } else if (s.newState === State.Running) {
-                    console.log('client running');
+                    const end = Date.now()
+                    console.log('client running, started in ', end - startTime, 'ms');
                     this.running = true; // may have been auto restarted after it failed.
                 } else if (s.newState === State.Stopped) {
                     console.log('client has stopped or it failed to start');
@@ -518,8 +520,7 @@ export class LeanClient implements Disposable {
         const versionOptions = version ? ['+' + version, '--version'] : ['--version']
         const start = Date.now()
         const lakeVersion = await batchExecute(executable, versionOptions, this.folderUri?.fsPath, undefined);
-        const end = Date.now()
-        console.log("Ran '" + executable + " " + versionOptions.join(' ') + "' in " + (end - start) + "ms");
+        console.log(`Ran '${executable} ${versionOptions.join(' ')}' in ${Date.now() - start} ms`);
         const actual = this.extractVersion(lakeVersion)
         if (actual.compare('3.0.0') > 0) {
             return true;
--- a/vscode-lean4/src/leanclient.ts
+++ b/vscode-lean4/src/leanclient.ts
@@ -22,6 +22,7 @@ import { assert } from './utils/assert'
 import { LeanFileProgressParams, LeanFileProgressProcessingInfo } from '@lean4/infoview-api';
 import { LocalStorageService} from './utils/localStorage'
 import { batchExecute, testExecute } from './utils/batch'
+import { readLeanVersion } from './utils/projectInfo';
 import { cwd } from 'process'
 import * as fs from 'fs';
 import { URL } from 'url';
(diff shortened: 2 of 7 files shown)


vscode-lean4/src/leanclient.ts
```diff
@@ -117,31 +118,22 @@ export class LeanClient implements Disposable {
...
-                    if (this.outputChannel) this.outputChannel.appendLine(failover);
+        if (useLake && this.folderUri) {
+            if (this.folderUri.scheme === 'file') {
+                const lakefile = Uri.joinPath(this.folderUri, 'lakefile.lean').toString()
+                if (!fs.existsSync(new URL(lakefile))) {
                     useLake = false;
+                } else {
+                    const toolchainVersion = await readLeanVersion(this.folderUri);
+                    if (toolchainVersion) {
+                        const date = this.extractToolchainDate(toolchainVersion);
```
gebner, line 129: This is unfortunately not a solution, since not every version is a nightly.  There version can also be `leanprover/lean4:4.0.0-m3` and you can also make custom releases like `gebner/lean4:vrpcrace` (which we sometimes use for testing).  From these version numbers you can't deduce the lake version at all.

README.md
```diff
@@ -308,3 +308,22 @@ the packages in tandem using Lerna.
...
+But for a quick summary you need to install this tool
+
+```
+npm install -g vsce
+```
+
+Then run this command
+```
+cd <vscode-lean4-repo>\vscode-lean4
+vsce package
```
gebner, line 325:
````suggestion
But for a quick summary, run this command:
```
cd <vscode-lean4-repo>/vscode-lean4
npx vsce package
````

We already recommend `npx` in other places in the readme.  (Also forward slashes should be cross-platform.)

**Review by gebner — CHANGES_REQUESTED** [review]

**Comment by gebner** [comment]
> remove the lake test because it is too slow and instead switch to a test based on the lean-toolchain version string instead which is a lot faster.
> The Lean 4: Restart Server command goes from 2.7 seconds to about 1.6 seconds.

I agree that we should improve the startup performance, I've also noticed that there's a noticeable delay.  But I don't think it's *that* bad for me.

Can you post the runtime of the following commands on your machine?
```bash
lean --version             #  31ms
lake --version             #  58ms
lake serve </dev/null      # 645ms
lean --server </dev/null   # 378ms
```

If the numbers are similar on your machine, then we should just get rid of the `lake serve` check.  At the moment, this would cause a small regression where the server fails to start if the `lakefile.lean` contains syntax errors.  But we can just bug @tydeu to release a new lake version with https://github.com/leanprover/lake/pull/52 to fix this regression.

tydeu:
> But we can just bug @tydeu to release a new lake version with [leanprover/lake#52](https://github.com/leanprover/lake/pull/52) to fix this regression.

Note that one can just check for a version greater than `3.0.0`, since any future version will include that fix.

gebner:
> Note that one can just check for a version greater than 3.0.0, since any future version will include that fix.

But then find-references won't work until you release 3.0.1. :smile:

tydeu:
@gebner what do you mean? The current version in the Lean repo is `3.1.0-pre`, so if the extension checks for versions greater than `3.0.0` it should activate with the current master but not for older nightlies using `3.0.0`. Right?

gebner:
Thanks, I missed the lake update yesterday. In that case everything is fine on the lake side and we just need to bump the version check here.

lovettchris:
> Can you post the runtime of the following commands on your machine?
```
lean --version             # 78ms
lake --version             #  119ms
lake serve </dev/null      # 500ms
lean --server </dev/null   # 431ms
```

lovettchris:
> Thanks, I missed the lake update yesterday. In that case everything is fine on the lake side and we just need to bump the version check here.

If I understand correctly you guys want me to change it from:
```typescript
 if (actual.compare('3.0.0') >= 0) {
```
to:
```typescript
 if (actual.compare('3.0.0') > 0) {
```
if so I optimistically pushed this change.

README.md
lovettchris: thanks.

vscode-lean4/src/leanclient.ts
lovettchris: Ok, I've re-instated the lake version check, but I've keep some "optimizations" for explicit date checking that I can do when the toolchain version is something we can recognize, like `leanprover/lean4:nightly-2022-02-01` so we get the best of both.  Optimized code path on known versions, and fall back to slower `lean --version` when necessary.

vscode-lean4/src/leanclient.ts
```diff
@@ -489,6 +481,53 @@ export class LeanClient implements Disposable {
...
 
+    private async checkToolchainVersion(folderUri: Uri) : Promise<Date | undefined> {
+        const lakefile = Uri.joinPath(folderUri, 'lakefile.lean').toString()
+        if (!fs.existsSync(new URL(lakefile))) {
+            return undefined;
+        }
+
+        // see if we have a well known toolchain label that corresponds
+        // to a known date like 'leanprover/lean4:nightly-2022-02-01'
+        const toolchainVersion = await readLeanVersion(folderUri);
```
gebner, line 492: I don't think this version check is necessary anymore (since we got rid of the extremely expensive `lake serve` check).  But if you want to keep it, then at least use a regexp please to parse the toolchain version:
```javascript
const match = /^leanprover\/lean4:nightly-(\d+)-(\d+)-(\d+)$/.exec(toolchainVersion);
if (match) {
  return new Date(match[1], match[2] - 1, match[3]);
}
```

PR #133, opened 2022-02-21
Add intelligent wrapping for theorems with parentheses

marianaalanis93:
Adding CSS word-wrap Property in order to allow intelligent wrapping for theorems with parentheses (including nested ones)

### DIFF
--- a/lean4-infoview/src/infoview/index.css
+++ b/lean4-infoview/src/infoview/index.css
@@ -31,7 +31,8 @@ html,body {
 }
 
 .pre-wrap {
-    white-space: 'pre-wrap';
+    white-space: pre-wrap;
+    word-wrap: break-word;
 }
 
 .error {


lean4-infoview/src/infoview/index.css
```diff
@@ -31,7 +31,8 @@ html,body {
 }
 
 .pre-wrap {
-    white-space: 'pre-wrap';
+    white-space: pre-wrap;
+    word-wrap: break-word;
```
lovettchris, line 35: Did you find a case that needed [break-work](https://www.w3schools.com/cssref/playdemo.asp?filename=playcss_word-wrap&preval=break-word). I don't recall seeing very long words...?

marianaalanis93: Yes! I'm still WIP checking a best solution for the parenthesis case (break-word resolve the issue with the parenthesis but break "big words" as you said)

marianaalanis93:
Closing because of some conflicts with fork

PR #134, opened 2022-02-21
Cleanup use of fs.existsSync()

marianaalanis93:
Cleanup use of fs.existsSync() adding try-catch in some read files statements and replacing deprecated "fs.existsSync()" to "fs.statSync()"


gebner (commented):
Thanks for looking into this!  Replacing `existsSync` by `statSync` is not the solution though.  (Also note that `statSync` throws an exception when the file doesn't exist.)

There are two reasons why we want to get rid of existsSync:
1. It's synchronous and can block.  We'd rather use async functions.
2. We often used existsSync to check the existence of a file before doing a readFile.  This is a problem if the file is deleted after existsSync but before readFile.  (The nodejs documentation explains this in the entry for fs.exists: https://nodejs.org/api/fs.html#fsexistspath-callback)

So in cases where we only do existsSync and then don't read the file afterwards, the replacement should be `await fs.promises.access(...).then(() => true, () => false)` (maybe add a helper function for that).

If the existsSync is followed by a readFile, then we should get rid of the existsSync entirely and catch the exception returned by readFile (as recommended in the nodejs docs).  Here it's probably also better to use the modern `fs.promises.readFile` instead of the old callback-style version or the synchronous readFileSync.

vscode-lean4/src/docview.ts
```diff
@@ -25,11 +25,11 @@ function findProjectDocumentation(): string | null {
     const rootPath = findActiveEditorRootPath();
     if (rootPath) {
         let html = join(rootPath, 'html', 'index.html');
-        if (fs.existsSync(html)) {
```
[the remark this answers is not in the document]
marianaalanis93: mmmmm good point. We can add fs.statSync inside the 'else' and just show in a message the stat of those file. WHat do you think?

vscode-lean4/src/utils/leanpkg.ts
```diff
@@ -93,13 +93,13 @@ export class LeanpkgService implements Disposable {
             // search parent folders for a leanpkg.toml file, or a Lake lean-toolchain file.
             while (true) {
                 const leanToolchain = Uri.joinPath(uri, this.toolchainFileName);
-                if (fs.existsSync(new URL(leanToolchain.toString()))) {
```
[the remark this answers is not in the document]
marianaalanis93: Ok will do it :)

vscode-lean4/src/utils/leanpkg.ts
```diff
@@ -172,7 +172,7 @@ export class LeanpkgService implements Disposable {
             // must be a lean-toolchain file, these are much simpler they only contain a version.
             const url = new URL(this.leanVersionFile.toString());
             return new Promise<string>((resolve, reject) => {
-                if (fs.existsSync(url)) {
+                if (fs.statSync(url)) {
```
marianaalanis93, line 175: @lovettchris just checking, in this case the "statSync" could be valid?

vscode-lean4/src/utils/fsExistHelper.ts
```diff
@@ -0,0 +1,18 @@
...
+export async function fsExistHelper(pathFile: string): Promise<boolean> {
+    /*
+    Function that helps to replace fs.existsSync (using existsSync to check for the existence
+    of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended.
+    Doing so introduces a race condition, since other processes may change the file's state between the two calls.
+    Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.)
+    */
+
+    const fileExists = await promises.access(this._temporaryStorage).then(() => true).catch(() => false);
+    if (!fileExists) {
```
Vtec234, line 14: Why not `return !fileExists`?

vscode-lean4/src/utils/fsExistHelper.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+import { promises } from 'fs';
+
+
+export async function fsExistHelper(pathFile: string): Promise<boolean> {
```
Vtec234, line 5: It looks like this function doesn't use its argument. What is `this._temporaryStorage`?

vscode-lean4/src/utils/fsExistHelper.ts
lovettchris: Hi Wojciech ,  this PR is a work in progress and I am working with Mariana on it, I'll let you know when it's ready for review, thanks.

Vtec234:
@marianaalanis93 did you mean to merge this? The protocol we usually follow is that we only merge after the reviewers state they have no more comments, or give at least one approving review and there have been no further comments for about a full day or so. In particular, PR authors don't usually merge their own PRs before the review period is over unless the patch is very small. I will revert for now.

vscode-lean4/src/utils/fsExistHelper.ts
Vtec234: I see, I started the review now because it has been merged. Please see #136, and I guess the discussion can continue in another PR from the same branch?

lovettchris: Yes, the merge was accidental, and I have undone the merge.

lovettchris: Yes, let's create a new PR since I don't know how to change the state of this PR back to unmerged.

Vtec234: Ah okay, sorry about creating even more confusion. I closed the above, and yeah it is not possible to go back to not-merged so a new PR is needed.

marianaalanis93:
@Vtec234 Thank you for your comments, I merged this PR by mistake (working with my fork) and the commit has already been reverted. As Chris said it is only a draft so I will let you now when it's ready. Thanks. (btw I just see your new PR about this so we can continue the review there :))

PR #135, opened 2022-02-22
feat: tag path settings as machine-specific

Vtec234:
Prevents remote machine settings from inheriting paths set up for the local machine. I also include `lean4.serverEnv` because that contains `LEAN_SRC_PATH` and possibly `LEAN_PATH`.


gebner:
I usually just set these per-workspace btw.

PR #137, opened 2022-02-23
[issue-119] Cleanup use of fs.existsSync()

marianaalanis93:
This PR contains the next changes:

- Add a helper for checking for the existence of a file before calling fs.open(), fs.readFile() or fs.writeFile() (is not recommended).
- Add some try-catch statements in order to avoid possible issues when a file is read or accessed.


vscode-lean4/src/utils/leanpkg.ts
```diff
@@ -119,11 +120,16 @@ export class LeanpkgService implements Disposable {
 
     // Return file contents with whitespace normalized.
     private readWhitespaceNormalized(fileUri: Uri) : string {
-        const contents = fs.readFileSync(fileUri.fsPath).toString();
-        // ignore whitespace changes by normalizing whitespace.
-        const re = /[ \t\r\n]+/g
-        const result = contents.replace(re, ' ');
-        return result.trim();
+        if(fsExistHelper(fileUri.fsPath)){
```
Vtec234, line 123: This change *adds* an exists-read race condition, right? Regarding what to do in case the `readFile` here fails, I think we should either handle the possible exception wherever `readWhitespaceNormalized` is normalized, or change the return type to something like `string || undefined`. Returning an empty string does not distinguish between an empty file and one that doesn't exist.

vscode-lean4/src/utils/fsExistHelper.ts
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+import { open, close, PathLike } from 'fs';
+
+
+export function fsExistHelper(pathFile: PathLike): boolean {
```
Vtec234, line 5: What is the purpose of this helper? It is not asynchronous, so it is just a reimplementation of `fs.existsSync`, no? In the previous PR version there was an `async` variant.

marianaalanis93: I will change this into an async function :)

vscode-lean4/old/src/librarynote.ts
```diff
@@ -17,17 +17,24 @@ export class LibraryNoteLinkProvider implements DocumentLinkProvider {
...
+                        link.target = leanFile.with({ fragment: `L${lineNo}` });
+                        return link;
+                    }
                 }
             }
+            await window.showErrorMessage(`Library note "${noteName}" not found.`);
+        }
+        catch (ex){
+            // Throw an exception in case any other issue is encountered
+            window.showErrorMessage('Exception thrown: .' + ex);
```
lovettchris, line 37: Probably don't need the full stop "." at the end of this message.

vscode-lean4/src/docview.ts
```diff
@@ -83,7 +84,7 @@ export class DocViewProvider implements Disposable {
     }
 
     private async offerToOpenProjectDocumentation() {
```
lovettchris, line 86: This appears to be dead code, please remove `offerToOpenProjectDocumentation`, thanks.

vscode-lean4/src/utils/fsHelper.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+import { readFile, promises, PathLike } from 'fs';
+
+
+export async function fsExistHelper(pathFile: PathLike): Promise<boolean> {
```
lovettchris, line 5: Ok, I think we can now give this a nicer name, how about "fileExists"

vscode-lean4/src/utils/fsHelper.ts
```diff
@@ -0,0 +1,39 @@
...
+    Doing so introduces a race condition, since other processes may change the file's state between the two calls.
+    Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.)
+    param: pathFile - A string representing a PathLike
+
+    returns Promise<boolean> a boolean that represents if a file exist
+    */
+    return await promises.access(pathFile).then(() => true, () => false);
+}
+
+export async function fsReadHelper(url: PathLike): Promise<string | null>{
```
lovettchris, line 18: and how about "fileRead" or "readFile"

vscode-lean4/src/utils/fsHelper.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    return await new Promise<string | null>((resolve, reject) => {
+        readFile(url, { encoding: 'utf-8' }, (err, data) =>{
+            if (err) {
+                reject(err);
+            } else {
+                if (data) {
+                    resolve(data.trim());
+                } else {
+                    resolve(null);
```
lovettchris, line 34: I wonder if in most cases an empty string here would be fine ('') then the promise can just be Promise<string>.

vscode-lean4/src/utils/leanpkg.ts
```diff
@@ -3,6 +3,7 @@ import { LocalStorageService} from './localStorage'
 import { findLeanPackageRoot, findLeanPackageVersionInfo } from './projectInfo';
 import * as fs from 'fs';
```
lovettchris, line 4: You can now remove `import * as fs from 'fs';` -- notice in VS code how this line is "dimmed" - this is a que that the import is no longer needed.  I think the same is true in some other files you changed.

vscode-lean4/src/utils/leanpkg.ts
```diff
@@ -67,11 +68,11 @@ export class LeanpkgService implements Disposable {
         // Note: just opening the file fires this event sometimes which is annoying, so
         // we compare the contents just to be sure and normalize whitespace so that
         // just adding a new line doesn't trigger the prompt.
-        const [workspaceFolder, packageUri, packageFileUri] = findLeanPackageRoot(uri);
+        const [workspaceFolder, packageUri, packageFileUri] = await findLeanPackageRoot(uri);
         if (packageUri) {
-            const fileUri = this.findLakeFile(packageUri);
+            const fileUri = await this.findLakeFile(packageUri);
             if (fileUri) {
-                const contents = this.readWhitespaceNormalized(fileUri);
```
lovettchris, line 74: this can throw which means it would be an unhandled exception in this function so handleLakeFileChanged needs a try catch.  I think the right semantic if an exception is thrown would be to pretend the file no longer exists so the `existing` content should go to empty string and the `this.lakeFileChangedEmitter.fire` even should be raised.

vscode-lean4/src/utils/leanpkg.ts
```diff
@@ -108,22 +109,28 @@ export class LeanpkgService implements Disposable {
...
+        try{
+            const contents = (await workspace.fs.readFile(fileUri)).toString();
+            // ignore whitespace changes by normalizing whitespace.
+            const re = /[ \t\r\n]+/g
+            const result = contents.replace(re, ' ');
+            return result.trim();
+        }
+        catch(ex) {
+            // In case there is an error in the read
+            throw ex;
```
lovettchris, line 132: There is no point to this try/catch block, it has the same semantics as if you had no try catch, but perhaps the right semantic is to NOT throw and return empty string, then handleLakeFileChanged can continue as normal and not have to worry about exceptions.

vscode-lean4/src/utils/projectInfo.ts
```diff
@@ -2,15 +2,16 @@ import * as fs from 'fs';
 import { URL } from 'url';
 import { Uri, workspace, WorkspaceFolder } from 'vscode';
 import { addDefaultElanPath } from '../config';
+import { fsExistHelper, fsReadHelper } from './fsHelper';
 
 // Detect lean4 root directory (works for both lean4 repo and nightly distribution)
-export function isCoreLean4Directory(path: Uri): boolean {
-    return fs.existsSync(Uri.joinPath(path, 'LICENSE').fsPath) && fs.existsSync(Uri.joinPath(path, 'LICENSES').fsPath);
+export async function isCoreLean4Directory(path: Uri): Promise<boolean> {
```
lovettchris, line 8: There is one more use of `isCoreLean4Directory` in `testLeanVersion` that you did not add an `await` keyword to.

vscode-lean4/src/utils/projectInfo.ts
```diff
@@ -123,39 +124,20 @@ async function readLeanVersionFile(packageFileUri : Uri) : Promise<string | null
...
-        });
+        const data = await fsReadHelper(url);
+        if (data) {
+            const match = /lean_version\s*=\s*"([^"]*)"/.exec(data);
+            if (match) return match[1].trim();
+        }
+
     } else {
         // must be a lean-toolchain file, these are much simpler they only contain a version.
-        return await new Promise<string | null>((resolve, reject) => {
```
lovettchris, line 145: Nice!  I always love to see lots of red during a code review.  One of the most useful tools in programming is the garbage can :-)

lovettchris (changes requested):
Looking good, I have a few comments to resolve.

lovettchris:
Can you also update this PR override where it says: "Add a helper for replacing existsSync to check for the existence
of a file before calling fs.open(), fs.readFile() or fs.writeFile() (is not recommended)."  Since we are now recommending you do NOT  use the exists help if you are about to open or read or write the file.

lovettchris:
I also merged the PR with master so we can see how the new unit tests perform, and hey, they are failing, so we may have a bug here somewhere... perhaps a missing await?

vscode-lean4/src/utils/projectInfo.ts
marianaalanis93: Of course :) thanks!

vscode-lean4/old/src/librarynote.ts
marianaalanis93: Oh good catch! thanks

vscode-lean4/src/utils/fsHelper.ts
marianaalanis93: Sure!!

vscode-lean4/src/utils/fsHelper.ts
marianaalanis93: Sure, that sounds cool

vscode-lean4/src/utils/leanpkg.ts
marianaalanis93: Sure! good catch!

vscode-lean4/src/utils/fsHelper.ts
marianaalanis93: I think that for the case of reading files, a ("") is enough. Since it was verified that the file exists but maybe it is simply empty (and null and "" for this case is the same)

vscode-lean4/src/utils/fsHelper.ts
lovettchris: sure fileRead sounds good.

vscode-lean4/src/utils/projectInfo.ts
marianaalanis93: Ok good catch

lovettchris:
@marianaalanis93 note that the test failure looks like a real regression in this PR, notice this error message in the test output log:
```
### error opening document: Error: [UriError]: cannot call joinPath on URI without path
```

vscode-lean4/old/src/librarynote.ts
```diff
@@ -17,17 +17,24 @@ export class LibraryNoteLinkProvider implements DocumentLinkProvider {
...
         const noteName = link.tooltip;
-        for (const leanFile of await workspace.findFiles('**/*.lean')) {
-            const content = (await workspace.fs.readFile(leanFile)).toString();
-            for (const m of content.matchAll(libraryNoteRegex)) {
-                console.log(m[1]);
-                if (m[1] === noteName) {
-                    const lineNo = content.substr(0, m.index).split(/\r\n|\r|\n/).length;
-                    link.target = leanFile.with({ fragment: `L${lineNo}` });
-                    return link;
+        // Adding try-catch statement for read file possible exceptions
```
gebner, line 20:
```suggestion
```
Comments are not the place to explain *changes* in a PR; that's what's the PR message is for.  Comments should explain why the code does what it does.

vscode-lean4/old/src/librarynote.ts
```diff
@@ -17,17 +17,24 @@ export class LibraryNoteLinkProvider implements DocumentLinkProvider {
...
+                        link.target = leanFile.with({ fragment: `L${lineNo}` });
+                        return link;
+                    }
                 }
             }
+            await window.showErrorMessage(`Library note "${noteName}" not found.`);
+        }
+        catch (ex){
+            // Throw an exception in case any other issue is encountered
+            await window.showErrorMessage('Exception thrown: ' + ex);
```
gebner, line 37:
```suggestion
            await window.showErrorMessage('Exception while searching for library note: ' + ex);
```
No exception is being thrown here (quite the opposite, we're catching one).

vscode-lean4/src/utils/clientProvider.ts
```diff
@@ -211,7 +211,8 @@ export class LeanClientProvider implements Disposable {
     // Returns a boolean "true" if the LeanClient was already created.
     // Returns a null client if it turns out the new workspace is a lean3 workspace.
     async ensureClient(uri : Uri, versionInfo: LeanVersion | undefined) : Promise<[boolean,LeanClient | undefined]> {
-        const [workspaceFolder, folder, packageFileUri] = findLeanPackageRoot(uri);
+        console.log('### Uri: ${uri}');
```
gebner, line 214:
```suggestion
```
Please remove the debugging statement.

vscode-lean4/src/utils/fsHelper.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+import { readFile, promises, PathLike } from 'fs';
+
+
+export async function fileExists(pathFile: PathLike): Promise<boolean> {
+    /*
+    Helper used to replace fs.existsSync (using existsSync to check for the existence
+    of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended.
+    Doing so introduces a race condition, since other processes may change the file's state between the two calls.
+    Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.)
```
gebner, line 10: Please turn this into a documentation comment (`/**`).

vscode-lean4/src/utils/fsHelper.ts
```diff
@@ -0,0 +1,39 @@
...
+    param: pathFile - A string representing a PathLike
+
+    returns Promise<boolean> a boolean that represents if a file exist
```
gebner, line 13:
```suggestion
```

No need to duplicate the type annotation.

vscode-lean4/src/utils/fsHelper.ts
```diff
@@ -0,0 +1,39 @@
...
+    */
+    return await promises.access(pathFile).then(() => true, () => false);
+}
+
+export async function fileRead(url: PathLike): Promise<string>{
```
gebner, line 18: This is just `(await fsPromises.readFile(url, {encoding: 'utf-8'})).trim()`.  I don't think we need a helper function for this.  (The name is also slightly misleading, because I wouldn't expect that a generic `fileRead` function trims the file contents).

vscode-lean4/src/utils/projectInfo.ts
```diff
@@ -47,27 +51,29 @@ export function findLeanPackageRoot(uri: Uri) : [WorkspaceFolder | undefined, Ur
...
                 }
-                else if (isCoreLean4Directory(path)) {
+                else if (await isCoreLean4Directory(path)) {
                     return [wsFolder, path, null];
                 }
                 else if (searchUpwards) {
                     const parent = Uri.joinPath(path, '..');
                     if (parent === path) {
                         // no project file found.
+                        console.log('### no project file found');
```
gebner, line 69: Please remove the debugging statement.

vscode-lean4/src/utils/projectInfo.ts
```diff
@@ -47,27 +51,29 @@ export function findLeanPackageRoot(uri: Uri) : [WorkspaceFolder | undefined, Ur
...
                         break;
                     }
                     path = parent;
                 }
                 else {
                     // don't search above a WorkspaceFolder barrier.
+                    console.log('### dont search above a WorkspaceFolder barrier.');
```
gebner, line 76: Please remove the debugging statement.

vscode-lean4/src/utils/projectInfo.ts
```diff
@@ -99,63 +105,44 @@ export async function findLeanPackageVersionInfo(uri: Uri) : Promise<[Uri | null
...
-export async function readLeanVersion(packageUri: Uri) : Promise<string | null> {
+export async function readLeanVersion(packageUri: Uri) : Promise<string> {
     const toolchainFileName = 'lean-toolchain';
     const tomlFileName = 'leanpkg.toml';
     if (packageUri.scheme === 'file') {
         const leanToolchain = Uri.joinPath(packageUri, toolchainFileName);
-        if (fs.existsSync(new URL(leanToolchain.toString()))) {
+        if (await fileExists(new URL(leanToolchain.toString()))) {
             return await readLeanVersionFile(leanToolchain);
         }
```
gebner, line 115:
```suggestion
        try {
            return await readLeanVersionFile(leanToolchain);
        } catch (e) {} // file does not exist
```
(assuming that `readLeanVersionFile` throws an exception if it can't open the file)

It's a bad idea to check for the existence of a file before reading it, because it introduces a race condition: the file could be deleted between the existence check and reading it.  So it's better to just handle the exception from the read operation.

gebner (commented):
BTW, the CI failure has nothing to do with this PR and is already fixed on master.

vscode-lean4/src/utils/clientProvider.ts
marianaalanis93: Sure!

vscode-lean4/src/utils/fsHelper.ts
marianaalanis93: Ok, resolved! Thanks!!

vscode-lean4/src/utils/fsHelper.ts
marianaalanis93: Ok, resolved! Thanks!!

vscode-lean4/src/utils/projectInfo.ts
marianaalanis93: Ok, thanks! :) resolved

vscode-lean4/src/utils/projectInfo.ts
marianaalanis93: Thanks!,, resolved!

vscode-lean4/src/utils/fsHelper.ts
marianaalanis93: Ok :) Resolved!

vscode-lean4/src/utils/projectInfo.ts
```diff
@@ -99,63 +103,38 @@ export async function findLeanPackageVersionInfo(uri: Uri) : Promise<[Uri | null
...
     const toolchainFileName = 'lean-toolchain';
     const tomlFileName = 'leanpkg.toml';
     if (packageUri.scheme === 'file') {
         const leanToolchain = Uri.joinPath(packageUri, toolchainFileName);
-        if (fs.existsSync(new URL(leanToolchain.toString()))) {
+        try {
             return await readLeanVersionFile(leanToolchain);
-        }
-        else {
-            const leanPkg = Uri.joinPath(packageUri, tomlFileName);
```
gebner, line 111: This might be a bit subtle, but we're actually trying to read two different files here: `lean-toolchain` and `leanpkg.toml`.  You've just deleted the `leanpkg.toml` part, please put it back.

vscode-lean4/src/utils/fsHelper.ts
gebner: The documentation comments needs to go *above* the function like this:
```typescript
/** Documentation goes here. */
function frobnicate() {}
```

lovettchris (changes requested):
Gabriel found an excellent bug involving realLeanVersion, once that is fixed I'm good with this PR.

vscode-lean4/src/utils/fsHelper.ts
marianaalanis93: Oh thanks! sorry I was thinking on python hehe

vscode-lean4/src/utils/projectInfo.ts
marianaalanis93: Sure! thanks guys!! :)

clovett (approved):
Looks good now, thanks!

gebner:
Thanks!

PR #142, opened 2022-02-28
feat: distinguish when messages are being loaded

Vtec234:


lovettchris (approved):

gebner:
Looks good to me.  As far as I can tell, this will only affect the behavior during startup though.  Once the infoview is loaded and messages have been shown at least once, the loading messages won't be shown again.

Vtec234:
The message also shows whenever you toggle "All Messages" back to being visible, if the server had sent in new diagnostics in the meantime.

PR #145, opened 2022-03-01
fix flakey test

lovettchris:
fix: sometimes depending on how we get to this function the Uri has different case on some folders, like I see "d:\temp\..." or "d:\Temp\..." and that caused this function to fail to see the file was in the same workspace and then it failed to create a new `lean --worker` at all for that file.  This was causing some test flakiness on Windows machines.

Also found and fixed some other obvious bugs in the tests and test utils.

### DIFF
--- a/vscode-lean4/src/leanclient.ts
+++ b/vscode-lean4/src/leanclient.ts
@@ -377,7 +377,7 @@ export class LeanClient implements Disposable {
 
     isSameWorkspace(uri: Uri){
         if (this.folderUri) {
-            if (this.folderUri.scheme != uri.scheme) return false;
+            if (this.folderUri.scheme !== uri.scheme) return false;
             if (this.folderUri.scheme === 'file') {
                 const realPath1 = fs.realpathSync(this.folderUri.fsPath).toLowerCase();
                 const realPath2 = fs.realpathSync(uri.fsPath).toLowerCase();
--- a/vscode-lean4/src/leanclient.ts
+++ b/vscode-lean4/src/leanclient.ts
@@ -1,5 +1,5 @@
 import { TextDocument, EventEmitter, Diagnostic,
-    languages, DocumentHighlight, Range, DocumentHighlightKind, window, workspace,
+    DocumentHighlight, Range, DocumentHighlightKind, workspace,
     Disposable, Uri, ConfigurationChangeEvent, OutputChannel, DiagnosticCollection,
     Position, WorkspaceFolder } from 'vscode'
 import {
@@ -23,7 +23,6 @@ import { LeanFileProgressParams, LeanFileProgressProcessingInfo } from '@lean4/i
 import { LocalStorageService} from './utils/localStorage'
 import { batchExecute, testExecute } from './utils/batch'
 import { readLeanVersion } from './utils/projectInfo';
-import { cwd } from 'process'
 import * as fs from 'fs';
 import { URL } from 'url';
 import { join } from 'path';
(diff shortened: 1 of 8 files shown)


vscode-lean4/src/leanclient.ts
```diff
@@ -378,9 +377,19 @@ export class LeanClient implements Disposable {
...
-                return true;
+            if (this.folderUri.scheme !== uri.scheme) return false;
+            if (this.folderUri.scheme === 'file') {
+                const realPath1 = fs.realpathSync(this.folderUri.fsPath).toLowerCase();
+                const realPath2 = fs.realpathSync(uri.fsPath).toLowerCase();
+                if (process.platform === 'win32') {
+                    // windows paths are case insensitive.
+                    return realPath2.startsWith(realPath1);
+                } else {
+                    return realPath2.toString().startsWith(realPath1);
```
gebner, line 388: This `toString` is a no-op since `realPath2` is already a string.

vscode-lean4/src/leanclient.ts
```diff
@@ -378,9 +377,19 @@ export class LeanClient implements Disposable {
 
     isSameWorkspace(uri: Uri){
         if (this.folderUri) {
-            if (uri.toString().startsWith(this.folderUri.toString())) {
-                // skip it, this file belongs to a different workspace...
-                return true;
+            if (this.folderUri.scheme !== uri.scheme) return false;
+            if (this.folderUri.scheme === 'file') {
+                const realPath1 = fs.realpathSync(this.folderUri.fsPath).toLowerCase();
+                const realPath2 = fs.realpathSync(uri.fsPath).toLowerCase();
```
gebner, line 383: Lower-casing should only be performed on windows.  (And probably on macos as well?)

gebner:
> But there is still a deeper issue here, gotoDefinition still fails sometimes, and it always fails when I see lower case letters in the file name tab in VS Code like this:

This is crazy, but I'm not at all surprised since this isn't well tested.  Can you make a reproducible test case for this?  If it's an upstream bug in the Lean 4 LSP server, then we should fix it there.

vscode-lean4/src/leanclient.ts
```diff
@@ -378,9 +377,19 @@ export class LeanClient implements Disposable {
 
     isSameWorkspace(uri: Uri){
         if (this.folderUri) {
-            if (uri.toString().startsWith(this.folderUri.toString())) {
-                // skip it, this file belongs to a different workspace...
-                return true;
+            if (this.folderUri.scheme !== uri.scheme) return false;
+            if (this.folderUri.scheme === 'file') {
+                const realPath1 = fs.realpathSync(this.folderUri.fsPath).toLowerCase();
```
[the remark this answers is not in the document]
lovettchris: Fixed

vscode-lean4/src/leanclient.ts
lovettchris: Fixed

vscode-lean4/src/leanclient.ts
lovettchris: fixed

vscode-lean4/src/utils/clientProvider.ts
```diff
@@ -170,14 +169,22 @@ export class LeanClientProvider implements Disposable {
         return Array.from(this.clients.values());
     }
 
-    getClientForFolder(folder: Uri) : LeanClient | undefined {
+    getKeyForFolder(folder: Uri | null) : string{
```
Vtec234, line 172: What is a key for a folder? Please put some explanation in a comment.

vscode-lean4/test/suite/utils/helpers.ts
```diff
@@ -172,10 +172,9 @@ export async function restartLeanServer(client: LeanClient, retries=10, delay=10
     return false;
 }
 
-export async function assertLeanVersion(infoView: InfoProvider, version: string) : Promise<string> {
```
Vtec234, line 175: This is more like "assertStringInInfoview" or something, right? Since e.g. `Lake Version:` is not itself a version, just a string that we look for.

Vtec234 (changes requested):

vscode-lean4/src/leanclient.ts
```diff
@@ -376,11 +377,33 @@ export class LeanClient implements Disposable {
         });
     }
 
-    isSameWorkspace(uri: Uri){
+    async getRealPath(fsPath : string) : Promise<string> {
```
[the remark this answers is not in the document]
lovettchris: I forgot about fsPromises, thanks.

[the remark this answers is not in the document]
lovettchris: fixed

vscode-lean4/test/suite/utils/helpers.ts
lovettchris: fixed

vscode-lean4/src/utils/clientProvider.ts
lovettchris: fixed

Vtec234 (approved):

PR #147, opened 2022-03-04
Add docview test

lovettchris:
Depends on https://github.com/leanprover/vscode-lean4/pull/145


vscode-lean4/src/docview.ts
```diff
@@ -262,6 +262,10 @@ export class DocViewProvider implements Disposable {
         }
     }
 
+    async getHtmlContents() : Promise<string> {
```
gebner, line 265: The `Promise` here is a bit misleading, since the function only returns the current state and doesn't wait until the docview is initialized.

gebner (commented):
These tests take quite a bit of code.  I'm also concerned that this will become a maintenance burden as we get more tests.  So I think it would be a good idea to concentrate the tests on key features that are likely to break without anyone noticing.

vscode-lean4/src/docview.ts
lovettchris: agreed, but the docview doesn't have an "api" like infoview has and I didn't think this test warranted adding one.  But at least it is is testing that the html generated can be parsed by cheerio and that a link to the example code exists and works when invoked.  I don't want to do "UI style testing" with mouse clicks and keystrokes and so on - that kind of test is a royal pain to maintain.  

In general I agree with you that we have to be fanatic about minimizing test code size.  This is why I want to move as much test code as possible into the utils.ts helpers.  The test code should read almost like a "spec" outline of the feature.  There is more work to be done to boil the test code down to it's core essence.  

In terms of what to test, I agree things most likely to break are highest priority, but one cannot always predict what that is going to be.  Coverage is also important in any test strategy.

lovettchris: Speaking of coverage, I was able to use [istanbuljs nyc](https://github.com/istanbuljs/nyc) to get the following coverage information for our current tests:

image

lovettchris: PS: @gebner if you approve this PR I will also promise to improve the test, cover all of docview, the abbreviations cheat sheet, clicking 'TryIt' buttons in the TPIL book and so on, so we get a bigger bang for the buck in terms of code coverage per line of test code.

gebner: > agreed, but the docview doesn't have an "api" like infoview has and I didn't think this test warranted adding one. 

I didn't take issue with the function or what it does, just the return type. :smile: 

> In general I agree with you that we have to be fanatic about minimizing test code size. This is why I want to move as much test code as possible into the utils.ts helpers. The test code should read almost like a "spec" outline of the feature. There is more work to be done to boil the test code down to it's core essence.

Great that we're on the same page here.

> cover all of [...] the abbreviations cheat sheet

Before testing the abbreviations cheat sheet, it might be worth it to test the actual abbreviations feature itself.

PR #150, opened 2022-03-08
Development config fixes

Vtec234:
(Regarding first commit.) I semi-frequently inspect the actual build artifacts and imported libraries, so it is more useful to me to keep them visible. They would still be excluded from search results with this change. I think you introduced some of these @gebner , are you okay with this change?

The other commits hopefully fix the ESLint setup.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -30,6 +30,7 @@ module.exports = {
         ],
         "sourceType": "module"
     },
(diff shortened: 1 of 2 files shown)


gebner:
Certainly agreed about the imported library, I've been annoyed by the ignore setting there as well.  Not sure how I feel about showing `out` again.  Let's try it out!

PR #151, opened 2022-03-09
Move infoview to JavaScript modules

Vtec234:
We move the infoview to be a JavaScript [module](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) rather than a UMD library. This will allow us to [dynamically import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports) widget modules provided by the Lean server and provide them with the same version of React that we use, as well as core UI components such as interactive expressions.
- [x] Clarify infoview documentation. `@lean4/infoview` is now becoming something that is to be imported both by the webview in order to call `renderInfoview`, as well as by custom widgets in order to use interactive expressions, messages, etc. We may want to split these functionalities into two packages eventually, but for now I think this is okay as long as it's explained.
- [x] The webpack `CopyPlugin` in `vscode-lean4` doesn't seem to propagate changes to the infoview on `npm run watch` properly.
- [ ] @lovettchris Could you please check whether the infoview debugging setup still works? This PR changes the code generation quite a bit.

/CC @EdAyers


Vtec234:
@lovettchris Nope, they are still `npm install; npx lerna bootstrap; npx lerna run build` at the top level. I just tried again and it works on Linux. Could you please try with a fresh clone? If it fails, it might is a Windows-specific problem.

lovettchris:
Perhaps your rollup calls are missing `--configPlugin typescript` ?

Vtec234:
Nope, the `rollup.config.js` is JavaScript, it does not need the TS plugin.

lovettchris:
I tried again starting with `git clean -dfx` on your branch, followed by `npm install`, then `npx lerna bootstrap` and then `npx lerna run build` and the build fails:

image

Looks like we need to setup a CI build on a Windows machine.

gebner:
AFAICT, you need to `import {React} from '@lean4/infoview'` now to get react.  I guess this won't work with libraries like react-d3-library, etc.  Is there a way to make `import * as React from 'react'` work, so that you can use it as a peer dependency?

vscode-lean4/webpack.config.js
```diff
@@ -4,10 +4,13 @@ const path = require('path');
...
 const CopyPlugin = require('copy-webpack-plugin');
 
+const prodOrDev = (env) => env.production ? 'production' : 'development';
+const minIfProd = (env) => env.production ? '.min' : '';
+
 function getWebviewConfig(env) {
 	let webview = {
 		name: 'webview',
-		mode: env.production ? 'production' : 'development',
+		mode: prodOrDev(env),
```
gebner, line 13: Since you're already cleaning up the prod/dev switch, should we use separate files for prod/dev build outputs?  It's a bit confusing that the files have a different content depending on whether you last compiled them from vscode or on the command-line.

Vtec234: I made this change in `@lean4/infoview` and then started in `vscode-lean4`, but making it there seems to complicate things more than it helps since it introduces lots of `production/development` switches in code that uses the bundle (I mean even more than [these ones](https://github.com/leanprover/vscode-lean4/pull/151/commits/7bd74683a36fca23fde454c72885c8fb906198db#diff-ece3fe1d415234d756ba9dd601f6701bb51c6bed1e5d07ecc0ac4ec4796e6e57L626)), so maybe it is best to leave as-is.

Vtec234:
This is ready to review. I still can't get `CopyPlugin` to update the files when `lean4-infoview` is recompiled in watch mode. It might be a bug in the plugin but I do not have the energy to investigate that.

lean4-infoview/package.json
```diff
@@ -3,34 +3,35 @@
...
   "dependencies": {
     "@lean4/infoview-api": "^0.1.0",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "@vscode/codicons": "0.0.28",
     "react-fast-compare": "^3.2.0",
-    "react-popper": "^2.2.5",
     "tachyons": "^4.12.0",
-    "vscode-codicons": "0.0.17"
+    "vscode-languageserver-protocol": "^3.17.0-next"
```
gebner, line 35: AFAICT we only use this for type-imports now?  Is there a reason it moved to `dependencies`?

Vtec234: I marked the type-imports to distinguish runtime-imports, of which there are two, [here](https://github.com/leanprover/vscode-lean4/blob/0abfd048b8937446a8ff7677a9c75b7432bc7cee/lean4-infoview/src/infoview/infos.tsx#L28) and [here](https://github.com/leanprover/vscode-lean4/blob/0abfd048b8937446a8ff7677a9c75b7432bc7cee/lean4-infoview/src/infoview/messages.tsx#L25). So currently it does get bundled in.

vscode-lean4/webpack.config.js
Vtec234: (In particular, we might want to put `extension.production.min.js` in the `.vsix` but use `extension.development.js` for extension development, and maybe also for the test runners, which adds the extra switches. Similarly for the webview bundle.)

lean4-infoview/README.md
```diff
@@ -1,11 +1,23 @@
...
+## Components
+
+A TypeScript API for use in widgets is exported in [`components.ts`](src/components.ts).
+
+⚠️ WARNING: The API is experimental and subject to breaking changes at any point.
 
 ## Hosting
 
-A hosting editor needs to be able to display a minimal webpage (e.g. via a Webkit panel, or in an external browser). The webpage should set up a connection to the editor as defined in [`src/infoviewApi.ts`](src/infoviewApi.ts) and invoke the [`renderInfoview`](src/index.ts) entry point. The `lean4-infoview` library is distributed as an UMD module with external dependencies specified in [`webpack.config.js`](webpack.config.js) `externals` .
+The infoview can be hosted within any LSP-compatible editor capable of displaying web content (e.g. via a WebKit panel, or in an external browser) and communicating with its context (e.g. via something like [`Window.postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage)). The hosting editor must also be able to send duplicates of received and emitted LSP notifications to the infoview, as well as relay requests between the infoview and the LSP server.
```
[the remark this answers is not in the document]
Vtec234: So it is! But the main functionality should mostly work as long as what's described here is implemented. I added a warning about VSCode-specific code.

gebner:
> The webpack CopyPlugin in vscode-lean4 doesn't seem to propagate changes to the infoview on npm run watch properly.

Possibly related to this: https://github.com/webpack-contrib/copy-webpack-plugin/issues/674

This is moderately annoying, can we add a hack to fix this like directly load `node_modules/@lean4/infoview/...` in dev mode?

gebner:
There's also this new warning which we should suppress (preferably by turning `eval` into `new Function` I guess).
```
@lean4/infoview: (!) Use of eval is strongly discouraged
@lean4/infoview: https://rollupjs.org/guide/en/#avoiding-eval
@lean4/infoview: src/infoview/main.tsx
@lean4/infoview: 78:         requestedAction: async (action) => editorEvents.requestedAction.fire(action),
@lean4/infoview: 79:         // eslint-disable-next-line no-eval
@lean4/infoview: 80:         runTestScript: async (script) => eval(script),
@lean4/infoview:                                              ^
@lean4/infoview: 81:         getInfoviewHtml: async () => document.body.innerHTML,
@lean4/infoview: 82:     };
```

gebner:
I'm also getting another (new?) warning:
```json
[{
	"resource": "/home/gebner/vscode-lean4/lean4-infoview/src/infoview/main.tsx",
	"owner": "eslint",
	"code": {
		"value": "@typescript-eslint/no-unsafe-argument",
		"target": "..."
	},
	"severity": 8,
	"message": "Unsafe argument of type `any` assigned to a parameter of type `string`.",
	"source": "eslint",
	"startLineNumber": 118,
	"startColumn": 45,
	"endLineNumber": 118,
	"endColumn": 51
}]
```

gebner:
>

Vtec234:
All should work now, `CopyPlugin` turned out to be a documented bad interaction with monorepos.

Vtec234:
> I'm also getting another (new?) warning:

I am not seeing this.

gebner:
> I am not seeing this.

Can't reproduce it either now.

gebner:
Did you test whether installing and running the extension from the packaged vsix still works?

lovettchris:
@Vtec234 - the CI failure you see here is due to a change I'm making to how windows elan-init.ps1 works (so there is no prompt when all goes well). See https://github.com/leanprover/elan/pull/63.  So we need to get my matching PR https://github.com/leanprover/vscode-lean4/pull/157 in ASAP and then when you merge that fix your PR will be good.

Kha:
Wait, so did we break the interface of the installer file used on e.g. https://leanprover.github.io/lean4/doc/quickstart.html? Could we please not do that?

lovettchris:
Good catch, yes, I need to update the quickstart, "--default-toolchain" should now be "-DefaultToolchain" (since it is a proper powershell parameter now).

Kha:
Weird, I'll have to check why it didn't update automatically. But I'm still not okay with simply breaking the interface of previous versions of the script *and* on other platforms, *especially* if internally they are translated back to the original interface anyway. If all this was for `PromptOnError`, surely this can be done in vscode-lean4 as well just like on Unix instead of making the platforms drift further apart?

lovettchris:
> Weird, I'll have to check why it didn't update automatically. But I'm still not okay with simply breaking the interface of previous versions of the script _and_ on other platforms, _especially_ if internally they are translated back to the original interface anyway. If all this was for `PromptOnError`, surely this can be done in vscode-lean4 as well just like on Unix instead of making the platforms drift further apart?

I would agree if all platforms were using the same scripting language.  But PowerShell is different, it has a different parameter parsing rules and I claim my change is more consistent with that.  But I can remove the "-PromptOnError" since the linux script does not support that concept, and move it back to the vscode extension if this is what you prefer.  But the syntax differences of powershell parameters "-DefaultToolchain" versus "--default-toolchain" should remain IMHO.

lovettchris:
> Weird, I'll have to check why it didn't update automatically. 

Currently Leo has to run `deploy.sh` in TPIL to push to gh-pages.  I asked Leo if he wanted an automated CI workflow to do that and he said no.

Vtec234:
> @Vtec234 - the CI failure you see here is due to a change I'm making to how windows elan-init.ps1 works (so there is no prompt when all goes well). See [leanprover/elan#63](https://github.com/leanprover/elan/pull/63). So we need to get my matching PR #157 in ASAP and then when you merge that fix your PR will be good.

Since the CI failure is unrelated to the content of this PR, I think it would be okay to just merge it now assuming there are no further comments?

lovettchris:
> Since the CI failure is unrelated to the content of this PR, I think it would be okay to just merge it now assuming there are no further comments?
> 
> **EDIT**: @lovettchris I think you accidentally put your response in my message by editing it :) Here is the response:
> 
> Please don't I want to see your PR pass on Windows first. So please merge #157 first, it's ready now, thanks.

Sorry about the edit, weird that github let me do that.  The re-run passed:
image

lovettchris:
Cool, the windows tests pass now!

lovettchris:
Hmmm they passed on the CI build machine, but they still fail on my devbox with this error.  Can you run the tests manually in vscode using the "Extension Tests - *" launch configurations?

```
TypeError: Cannot read property 'describe' of undefined
```

Vtec234:
> Can you run the tests manually in vscode using the "Extension Tests - *" launch configurations?

Yes, they run and pass locally for me.

Kha:
> > But I'm still not okay with simply breaking the interface of previous versions of the script _and_ on other platforms, _especially_ if internally they are translated back to the original interface anyway. If all this was for `PromptOnError`, surely this can be done in vscode-lean4 as well just like on Unix instead of making the platforms drift further apart?
> 
> I would agree if all platforms were using the same scripting language. But PowerShell is different, it has a different parameter parsing rules and I claim my change is more consistent with that. But I can remove the "-PromptOnError" since the linux script does not support that concept, and move it back to the vscode extension if this is what you prefer. But the syntax differences of powershell parameters "-DefaultToolchain" versus "--default-toolchain" should remain IMHO.

The parameters come from the `elan-init` binary, that's not a scripting language. It still looks like unnecessary churn to me given that it worked fine before, but if we're changing that, let's do it the right way: there is an unexposed flag `--no-modify-path`, and the default for `default-toolchain` is `stable`, not `none` https://github.com/leanprover/elan/blob/master/src/elan-cli/setup_mode.rs

lovettchris:
Ok @Kha, I'll move this discussion to Zulip since it is unrelated to this PR

lovettchris:
> > Can you run the tests manually in vscode using the "Extension Tests - *" launch configurations?
> 
> Yes, they run and pass locally for me.

On Linux?  And that's with vscode and not with "npm run tests" ?

Vtec234:
> On Linux? And that's with vscode and not with "npm run tests" ?

Yup to both. `npm run test` also works for me.

I'll merge this now as much of the discussion has lost its relevance to the PR itself, if there are remaining dev setup issues let's fix them in followup PRs.

Issue #152, opened 2022-03-11
`insertText` is broken when no position is specified.


Vtec234:
Per the [API](https://github.com/leanprover/vscode-lean4/blob/e0870d3eba00985f17563bdc1ce990964cb0952e/lean4-infoview-api/src/infoviewApi.ts#L52), `insertText` should write at the current cursor location when `pos === undefined`. This seems to have been broken since #104. /CC @lovettchris

lovettchris:
Thanks, see https://github.com/leanprover/vscode-lean4/pull/153

Issue #27, opened 2021-06-08
Support "untrusted workspaces"


gebner:
https://github.com/microsoft/vscode/issues/106488

leodemoura:
@gebner What should we do to address this issue?

gebner:
First we need to figure out if we need to take any action here.  Basically we should not start the Lean server if the user clicks "No, I don't trust the authors", but it would be nice if syntax highlighting and the input mode still worked.

PR #153, opened 2022-03-11
fix: issue 152, copyToComment was broken

lovettchris:
fixes: #152 
Copy to comment feature in the infoview was broken.
Also added a unit test for this feature.
The unit test found a new bug, in the previous code when the cursor is on line zero it blew up so I fixed that also.


vscode-lean4/webview/index.ts
```diff
@@ -5,7 +5,7 @@ import { Rpc } from '../src/rpc';
 
 const vscodeApi = acquireVsCodeApi();
 
-const rpc = new Rpc((m) => vscodeApi.postMessage(m));
+const rpc = new Rpc((m) => { try { vscodeApi.postMessage(m) } catch { /* ignore errors */} });
```
Vtec234, line 8: Did you run into an error appearing here? Usually it is an indication of a bug elsewhere in the application. If there is one specific kind of error that is okay to ignore, we should catch only that and explain why it is okay to do so, but in general I would strongly prefer we avoid empty `catch` blocks because they tend to make it difficult to find out what has gone wrong later on.

Vtec234 (changes requested):
Thanks for looking into this! I reviewed all changes except for the test code, which I will trust you to handle.

vscode-lean4/webview/index.ts
lovettchris: I'm trying to fix some random shutdown stack traces in this area that seem to happen while the tests are running after close all windows and vs is shutting down.  It is not something a user would ever see, but it does mess up the test logs.

lean4-infoview/src/infoview/info.tsx
```diff
@@ -50,20 +50,20 @@ export function InfoStatusBar(props: InfoStatusBarProps) {
         {isPinned && isPaused && ' (pinned and paused)'}
         <span className="fr">
             {copyGoalToComment &&
-                <a className="link pointer mh2 dim codicon codicon-quote"
+                <a id="copyGoalToComment" className="link pointer mh2 dim codicon codicon-quote"
```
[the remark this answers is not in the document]
lovettchris: Technically it works because the HTML DOM defines getElementById such that it returns the "first" element with that id, but I will change it to a unique CSS class so I can use `getElementsByClassName('copytoComment')[0]` instead, would that be better?

lean4-infoview-api/src/infoviewApi.ts
```diff
@@ -128,6 +128,11 @@ export interface InfoviewApi {
    */
   requestedAction(action: InfoviewAction): Promise<void>;
 
+  /**
+   * Interface for running test scripts inside the infoview webview.
+   */
+  runTestScript(javaScript: string) : Promise<void>;
```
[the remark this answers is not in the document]
lovettchris: `requestedAction('copyToComment')` only seems to work on goals, and not on Messages (see copyGoalToComment).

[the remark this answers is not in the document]
lovettchris: And again I wanted the most general hook for testing so I never have to change it.  Why do I need to avoid eval exactly?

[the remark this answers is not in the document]
Vtec234: The extension guide [recommends](https://code.visualstudio.com/api/extension-guides/webview#content-security-policy) setting a CSP with `script-src: ${webview.cspSource};` which disables inline scripts and `eval`. On second thought, though, with widgets we are about to give the Lean server the ability to execute arbitrary code *anyway* which we should handle in #27. So okay, let's go with `eval`.

lean4-infoview-api/src/infoviewApi.ts
```diff
@@ -128,6 +128,11 @@ export interface InfoviewApi {
    */
   requestedAction(action: InfoviewAction): Promise<void>;
 
+  /**
+   * Interface for running test scripts inside the infoview webview.
```
Vtec234, line 132:
```suggestion
   * Execute the given JavaScript code inside the infoview. Must not be used
   * for anything other than testing.
```
To clarify that the argument is source code rather than e.g. the name of a test script.

lean4-infoview/src/infoview/info.tsx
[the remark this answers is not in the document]
Vtec234: That does sound better, but then we are almost guaranteed to eventually forget these are for testing and use these classes to style things. How about `<a data-button="copyGoalToComment"` and then using a [dataset selector](https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelectorAll#dataset_selector_attribute_selectors)?

vscode-lean4/webview/index.ts
Vtec234: What do you mean by "mess up the test logs"? Do the tests fail, or is there just some extra output? In either case it sounds like we are maybe not shutting down properly which would be the real bug. I still think we shouldn't avoid empty `catch` to the extent possible.

lean4-infoview-api/src/infoviewApi.ts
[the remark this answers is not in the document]
gebner: :+1: for `eval`.  I'm not sure what threat model a CSP would address.  If you're running Lean locally then you're already fully trusting the extension, Lean, and the code you're editing.  The only advantage I can see of restricting the infoview would be if we wanted to insert untrusted HTML from the internet or something (but I don't think we plan to do something like that).

Maybe there would also be some advantage with the SSH extension, but I have no idea if it's supposed to be some sort of security boundary.  The container extension is certainly completely useless for that purpose.

vscode-lean4/src/infoview.ts
```diff
@@ -568,23 +578,29 @@ export class InfoProvider implements Disposable {
...
-            new_command = `\n${margin_str}${new_command}`;
+            new_command = `${margin_str}${new_command}\n`;
+            const insertPosition = current_line.range.start;
 
             await editor.edit((builder) => {
-                builder.insert(prev_line.range.end, new_command);
+                builder.insert(insertPosition, new_command);
             });
-            editor.selection = new Selection(pos.line, spaces, pos.line, spaces);
+            // select the whole insert so it is easy for user to delete it.
```
gebner, line 594: This is quite annoying (at least with the extensions that I use) because I first need to clear the selection before I can type something again.  Immediately deleting the inserted text is certainly not the intended use case.

vscode-lean4/src/infoview.ts
```diff
@@ -568,23 +578,29 @@ export class InfoProvider implements Disposable {
...
+            editor.selection = new Selection(pos.line, spaces, pos.line + last_line, spaces + length_of_last_line);
         } else {
             await editor.edit((builder) => {
                 if (pos) builder.insert(pos, text);
             });
             editor.selection = new Selection(pos, pos)
         }
+        // ensure the text document has the keyboard focus.
+        void window.showTextDocument(editor.document, { viewColumn: editor.viewColumn, preserveFocus: false });
```
gebner, line 603:
```suggestion
        await window.showTextDocument(editor.document, { viewColumn: editor.viewColumn });
```
According to the docs, `false` is the default.

I'm not sure if it's better to leave the focus in the infoview or to move it to the editor (potentially also switching to a different file!).  But I can't see any obvious problems with focusing the modified file.

vscode-lean4/test/suite/info/info.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+        console.log('=================== Copy to Comment ===================');
+
+        const lean = await initLean4Untitled('#eval Lean.versionString');
+        const info = lean.exports.infoProvider as InfoProvider;
+
+        await assertStringInInfoview(info, '4.0.0-nightly-');
+
+        console.log('Clicking copyToComment button in InfoView');
+        await info.runTestScript('document.getElementsByClassName(\'codicon-quote\')[0].click()');
```
gebner, line 19: Please just give the button its own class name, and then access it using that name.  I don't want this test to fail just because we change the button icon.

vscode-lean4/test/suite/info/info.test.ts
```diff
@@ -0,0 +1,32 @@
...
+import { initLean4Untitled, assertStringInInfoview, findWord } from '../utils/helpers';
+import { InfoProvider } from '../../../src/infoview';
+
+suite('InfoView Test Suite', () => {
+
+    test('Copy to Comment', async () => {
+
+        console.log('=================== Copy to Comment ===================');
+
+        const lean = await initLean4Untitled('#eval Lean.versionString');
```
gebner, line 13:
```suggestion
        const lean = await initLean4Untitled('#eval "copyToCommentTest"');
```
We don't want this test to fail immediately when switching to (or just testing with) release versions.

vscode-lean4/src/infoview.ts
lovettchris: You'd typically have to move the cursor for your next edit anyway, and that would clear the selection.

gebner:
BTW, it looks like the PR contains both a fix for #152 as well as some unrelated refactorings.  Could you please split these two changes into different PRs?

vscode-lean4/src/infoview.ts
lovettchris: It can't have been the default because I see a very real change with this.  Previously the infoview remains the active window (it is has the white close box in the tab).  Whereas with my code the active window moves back to my `Main.lean` file making it clear I can just start typing again...

vscode-lean4/test/suite/multi/multi.test.ts
```diff
@@ -17,27 +17,15 @@ suite('Multi-Folder Test Suite', () => {
...
-
-		await waitForActiveEditor('Main.lean');
-
-		const info = lean.exports.infoProvider as InfoProvider;
-        assert(await waitForInfoViewOpen(info, 60),
-			'Info view did not open after 20 seconds');
+		const lean = await initLean4(path.join(testsRoot, 'test', 'Main.lean'));
 
 		// verify we have a nightly build running in this folder.
+		const info = lean.exports.infoProvider as InfoProvider;
```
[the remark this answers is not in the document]
lovettchris: Sure, how do I do that?

lovettchris:
> BTW, it looks like the PR contains both a fix for #152 as well as some unrelated refactorings. Could you please split these two changes into different PRs?

How about next time.  The refactoring is used by the new test I added to prove the bug was fixed.

vscode-lean4/test/suite/info/info.test.ts
lovettchris: Sure I used numbers instead because I want the string coming from infoview copy to comment to not already be in the text editor.

vscode-lean4/test/suite/info/info.test.ts
lovettchris: Done.

vscode-lean4/webview/index.ts
lovettchris: But your point about hiding other future errors is a good one, so I removed the try catch.

lean4-infoview/src/infoview/main.tsx
```diff
@@ -95,6 +95,7 @@ export function renderInfoview(editorApi: EditorApi, uiElement: HTMLElement): In
         serverRestarted: new Event(),
         changedCursorLocation: new Event(),
         changedInfoviewConfig: new Event(),
+        runTestScript: new Event(),
```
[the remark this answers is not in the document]
gebner: Oh well, let's leave it like this then.

vscode-lean4/src/infoview.ts
gebner: Not with the vim extension at least.

vscode-lean4/src/infoview.ts
gebner: Previously = without `showTextDocument`, right?  I'm only suggesting to drop the `preserveFocus` argument, if merely dropping this argument results in the behavior that you describe then the docs are wrong and feel free to ignore my comment.

vscode-lean4/test/suite/multi/multi.test.ts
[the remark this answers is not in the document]
gebner: Right now the `activate` function returns a `Promise<any>` and the `waitForActiveExtension` returns a `Promise<Extension<any>>`.  All you need to do is define an `interface Exports { ... }` and then replace the `any` by `Exports`.

gebner:
> The PR is dependent on #147, so when that goes in this PR will get simpler.

Please mention that in the PR body next time.  That PR is merged now.

lean4-infoview/src/infoview/info.tsx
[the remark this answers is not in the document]
lovettchris: Done, I switched to dataset selector.  Thanks for the excellent suggestion.

vscode-lean4/src/infoview.ts
lovettchris: I also think it is good manners when inserting new text to make it easy to delete it if the user decides that's not what they wanted although I suppose undo would also do that.

vscode-lean4/test/suite/multi/multi.test.ts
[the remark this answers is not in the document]
lovettchris: will do, and I assume this code in the lean3 package will still work with that object?
```
return !(await lean4.activate()).isLean4Project;
```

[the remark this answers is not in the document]
lovettchris: Done, weird that typescript can't auto-initialize undefined fields...
```
            return { isLean4Project: false, version: toolchainVersion,
                infoProvider: undefined, clientProvider: undefined, installer: undefined, docView: undefined };
```

vscode-lean4/src/infoview.ts
lovettchris: The selection also makes it more visible what we just did int the copy to comment command.

lovettchris: and it makes it easy to move the pasted comment :-) all that for a tiny inconvenience of having to hit right arrow... :-)

lovettchris:
> Please mention that in the PR body next time. That PR is merged now.

Will do, I merged the changes from #147 so this should be a bit smaller now.

vscode-lean4/test/suite/multi/multi.test.ts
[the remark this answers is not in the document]
lovettchris: I tested the lean3 extension and the `return !(await lean4.activate()).isLean4Project;` still works.

vscode-lean4/src/infoview.ts
Vtec234: After having just tried it, I am *personally* slightly in favour of not adding the selection as the window switch with `preserveFocus: false` seems quite sufficient to indicate what the extension did. But! The "personally" is important ─ it seems hard to decide this kind of thing without doing a user study which we certainly don't want to do for such a tiny feature. So ultimately I would be okay with either option.

vscode-lean4/src/infoview.ts
Vtec234: It seems to default to `preserveFocus: false` for me, but I cannot see a guarantee of this in the docs (they just say what happens on `true`). It might even differ across platforms? Anyway I see no harm in being explicit, but we *do* need the `await` instead of the `void`.

Vtec234 (approved):
LGTM (thanks!!).

vscode-lean4/src/infoview.ts
lovettchris: I'd also be happy to to get more user feedback on this before deciding the issue.

vscode-lean4/src/infoview.ts
lovettchris: oh, can you educate me on why `await` is needed?  There's another `void window.showTextDocument` on line 557 that will also need fixing if this is required...

Vtec234: In this case I can't see things going terribly wrong without it, but it's good style IMO to use `await` instead of `void` when making `async` calls in `async` contexts since it guarantees a kind of atomicity, i.e. by the time `handleInsertText` finishes we are guaranteed that the document has been shown. As far as I know, `void` just launches the action and leaves it to finish whenever (including after `handleInsertText` has returned), which can sometimes end up with hard-to-reproduce bugs due to interleaved execution.

lovettchris: That's convincing, I added 2 missing awaits.  I've been using a lot of void on async calls for "fire and forget" semantics we so probably need to review all of these for consistency now and in future PR's.  Thanks for the education.

vscode-lean4/src/infoview.ts
gebner: Anecdotally, I don't think anybody has ever complained about the current behavior either.  I'd really prefer to keep the behavior change out of this bug fix.

> all that for a tiny inconvenience of having to hit right arrow... :-)

It's escape with the vim extension.  And you don't expect to land in visual mode since pretty much no other text insertion (such as paste) changes the selection.

And you found the right word: it's an inconvenience.  The change is a trade-off.  At best, we're making one workflow easier and another one more complicated.  At worst, users need to relearn how to use the extension just for change's sake.

lovettchris: Ok, I removed the selection change.  In fact, I removed all selection movement to so your cursor stays exactly where it was.

PR #154, opened 2022-03-15
fix: add Windows build & test to our github workflow

lovettchris:
Add Windows CI workflow, the macOs run is commented out for now because the Lake test is failing.


Kha:
Might be a silly question, but what does all that churn of downloading a lean4 artifact buy you over just testing against a nightly/stable from elan?

lovettchris:
> Might be a silly question, but what does all that churn of downloading a lean4 artifact buy you over just testing against a nightly/stable from elan?

Oh, one of the vscode tests wants to test a "linked toolchain", selected by "Select Toolchain" so I create a elan toolchain link called "master" pointing to a raw build of lean which I get from the lean4 artifact (which is quicker and easier than building lean4).

lovettchris:
This is interesting the macOS test seems to be crashing with no useful error message, see [failed run](https://github.com/leanprover/vscode-lean4/runs/5562416002?check_suite_focus=true).  Earlier tests pass, but the `Untitled Lean File` seems to get off to a good start, finding the `4.0.0-nightly` version message but then quits with `child proc::error::The process '/Users/runner/hostedtoolcache/node/16.14.0/x64/bin/npx' failed with exit code 1`.

Anyone have a macOS machine handy who can run the tests and let me know if they pass?

Kha:
> > Might be a silly question, but what does all that churn of downloading a lean4 artifact buy you over just testing against a nightly/stable from elan?
> 
> Oh, one of the vscode tests wants to test a "linked toolchain", selected by "Select Toolchain" so I create a elan toolchain link called "master" pointing to a raw build of lean which I get from the lean4 artifact (which is quicker and easier than building lean4).

This should work just as well with two non-linked toolchains as vscode-lean4 shouldn't be able to tell the difference. But if you want to keep the tests as is, you could do
```
elan toolchain link master ~/.elan/toolchains/leanprover--lean4---nightly
```

gebner:
I think the idea with the linked-toolchain test (or at least what I expected it to do) was to check whether everything works fine with a local development checkout, which has a different directory layout than the release tarballs and is not located in a well-known location like `.elan/toolchains/foo`.

Using the build artifact partly defeats that purpose because it's literally a release tarball, and thus has the same directory structure as the release tarball.  So if the choice is between nightly release and build artifact, we should just take the nightly release to keep it simple.  Maybe copy it somewhere else than `.elan/toolchains` though.

lovettchris:
> Using the build artifact partly defeats that purpose because it's literally a release tarball, and thus has the same directory structure as the release tarball. So if the choice is between nightly release and build artifact, we should just take the nightly release to keep it simple. Maybe copy it somewhere else than `.elan/toolchains` though.

This PR is not putting it in .elan, it is installing it in a local folder named "lean4", so since the "local folder" is probably the vscode-lean4 repo, then it's going into vscode-lean4/lean4.    I agree it's not quite as good as using a real `build/release/stage1` folder, but I don't want this CI to require building lean4, that would make it too big and slow.  The Windows test also needs the work that lean4 does to find and package the dependent libraries like ` libgcc_s_seh-1.dll  libgmp-10.dll  libstdc++-6.dll  libwinpthread-1.dll`.

lovettchris:
> This should work just as well with two non-linked toolchains as vscode-lean4 shouldn't be able to tell the difference. But if you want to keep the tests as is, you could do

Technically this is true, I suppose I could `cp -r` the contents of the `~/.elan/toolchains/leanprover--lean4---nightly`  to some other location and get most of the benefit, and the test would run faster.  One tiny potential advantage of using the github artifact is that we could catch a new problem sooner than waiting for nightly build.  But if guys don't care about that I can stop doing the artifact download...

.github/workflows/on-push.yml
```diff
@@ -21,50 +34,56 @@ jobs:
...
-          elan toolchain link master "$(pwd)/lean4"
+          mkdir temp
+          cp -r $HOME/.elan//toolchains/leanprover--lean4---nightly/* temp
+          elan toolchain link master temp
+
+      - name: Set path to elan on Windows
+        shell: pwsh
+        if: matrix.os == 'windows-latest'
+        run: |
+          echo "$HOME\.elan\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```
gebner, line 68: The $GITHUB_PATH doesn't work on windows?

gebner (commented):
I think the title is a bit misleading now (since it only adds windows support but not macos).

Most of the changes seem be tab->spaces and renaming waitForHtmlString; is there something I should be looking at that I have missed?

Otherwise lgtm.

.github/workflows/on-push.yml
lovettchris: This is appending to the GITHUB_PATH file, it's just in powershell you access an environment variable with the `$env:` prefix.  I probably can do `echo `$HOME\.elan\bin" >> $env:GITHUB_PATH` here.

lovettchris:
> I think the title is a bit misleading now (since it only adds windows support but not macos).

I fixed the PR title to say "fix: add Windows build & test to our github workflow"

.github/workflows/on-push.yml
gebner: Ah, makes sense.

gebner:
Please also fix the merge conflicts.

lovettchris:
> Please also fix the merge conflicts.

Ok, I fixed the merge conflicts, things are green again, so this PR is now ready.

gebner:
Hmm, the windows CI fails after merging into master now.

lovettchris:
Looks like it hit the infamous non-updating info view:

image

lovettchris:
And should be fixed by https://github.com/leanprover/vscode-lean4/pull/157

Issue #114, opened 2022-02-04
Infoview should indicate when server is not started


gebner:
Right now it shows "no info found" even if the server did not start.

lovettchris:
Really? Mine says "Waiting for Lean server to start..."  was this fixed recently perhaps?

gebner:
I think I hit that issue back when we didn't start a Lean server for some directories that didn't contain lakefiles.  I still see a similar issue when the server is terminated manually: #155

Vtec234:
It seems like another instance of this issue occurs when there are errors in `lakefile.lean`. In this case `lake serve` enters a mode in which `lakefile.lean` can be edited, but every other file should immediately fail (see leanprover/lake#52). However the infoview gets stuck displaying "Waiting for server to start...", even when `lakefile.lean` is the active editor. See also [Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/VSCode.20.22Waiting.20for.20lean.20server.20to.20start.2E.2E.2E.22).

lovettchris:
Specifically: "part of the problem seems to be that when lakefile.lean contains errors, lake serve prints them e.g. like so

```
./lakefile.lean:5:18: error: unknown constant 'Lake.PackageFacet.olean'
error: package configuration `./lakefile.lean` has errors
```
but then continues to execute the LSP server anyway. From the vscode extension's perspective, since it received no error code the server started up okay, and then this somehow leads to showing "Waiting" forever. Perhaps Lake should just exit with 1 at this point. /CC @Mac

Mac: @Wojciech Nawrocki this is actually a new feature, not a bug. See Lake [#52](https://github.com/leanprover/lake/pull/52). :stuck_out_tongue_wink:

Wojciech Nawrocki: I see! Then vscode-lean4 should detect and handle this case.

marianaalanis93:
Fixed and merged on https://github.com/leanprover/vscode-lean4/pull/159

PR #159, opened 2022-03-25
[issue-114] Infoview should indicate when server is not started

marianaalanis93:


Vtec234:
Hi! Thank you for looking into this! This is still a draft so I am a bit early to comment, but I just wanted to check-in to see where the changes are headed. I was surprised to see the changes in `info.tsx`, as my initial guess for how to fix #114 would be to only change the top-level component in `main.tsx`. Specifically, we already have [logic](https://github.com/leanprover/vscode-lean4/blob/49a2e1bcc8e21c372c5a904913c1a7616bb20950/lean4-infoview/src/infoview/main.tsx#L58) there to handle a server which isn't yet running, so it seems this change would naturally fit there as well. What is the motivation for changing components deeper in the UI hierarchy?

Vtec234:
(If the reason is to support multi-server workspaces, I think something like this makes sense and if it works, great. We will however need more changes to really support that. For example there is only one `serverInitializeResult` at the moment.)

marianaalanis93:
> Hi! Thank you for looking into this! This is still a draft so I am a bit early to comment, but I just wanted to check-in to see where the changes are headed. I was surprised to see the changes in `info.tsx`, as my initial guess for how to fix #114 would be to only change the top-level component in `main.tsx`. Specifically, we already have [logic](https://github.com/leanprover/vscode-lean4/blob/49a2e1bcc8e21c372c5a904913c1a7616bb20950/lean4-infoview/src/infoview/main.tsx#L58) there to handle a server which isn't yet running, so it seems this change would naturally fit there as well. What is the motivation for changing components deeper in the UI hierarchy?

Thanks @Vtec234 for your comments! (the more the better so no worries). Yeah, at the beginning I just wanted to add the message inside main.tsx but even though the message is displayed correctly ("server unavailable...") the section containing "All messages" remains static (and maybe it's a bit confusing for the user to keep seeing this section if the server is dead), that's why I was checking to change info.tsx logic, until now I'm blocked because I need to catch the exception related to https://github.com/microsoft/vscode-languageserver-node/issues/825. Still investigating another alternatives :)

marianaalanis93:
> Hi! Thank you for looking into this! This is still a draft so I am a bit early to comment, but I just wanted to check-in to see where the changes are headed. I was surprised to see the changes in `info.tsx`, as my initial guess for how to fix #114 would be to only change the top-level component in `main.tsx`. Specifically, we already have [logic](https://github.com/leanprover/vscode-lean4/blob/49a2e1bcc8e21c372c5a904913c1a7616bb20950/lean4-infoview/src/infoview/main.tsx#L58) there to handle a server which isn't yet running, so it seems this change would naturally fit there as well. What is the motivation for changing components deeper in the UI hierarchy?

Also, if you terminate the server manually the main.tsx is not updated anymore, so maybe we need to add more logic for re-render this part

lovettchris:
Update: Mariana and I figured out a solution by adding a new `serverStopped` event on the InfoViewApi.

lean4-infoview/src/infoview/messages.tsx
```diff
@@ -143,7 +142,7 @@ export function AllMessages({uri: uri0}: { uri: DocumentUri }) {
 }
 
 /** We factor out the body of {@link AllMessages} which lazily fetches its contents only when expanded. */
-function AllMessagesBody({uri, messages}: {uri: DocumentUri, messages: () => Promise<InteractiveDiagnostic[]>}) {
+function AllMessagesBody({uri, messages}: {uri: DocumentUri, messages: () => Promise<InteractiveDiagnostic[] | undefined>}) {
```
lovettchris, line 145: I don't see how it can be undefined.  If we look at this line `<AllMessagesBody uri={uri} messages={iDiags} />` and backtrack to the initialization of `iDiags` we can see it comes from `usePausableState` which is initialized to `iDiags0` and `iDiags0` has type Promise<InteractiveDiagnostic[]>.  And while `getInteractiveDiagnostics` can return undefined, it if does do that then this little `or` statement at the end of the line turns that into an empty array: ` || [];` so it is no longer undefined.

lean4-infoview/src/infoview/main.tsx
```diff
@@ -58,6 +59,8 @@ function Main(props: {}) {
         ret = <p>Waiting for Lean server to start...</p>
     } else if (!curUri) {
         ret = <p>Click somewhere in the Lean file to enable the infoview.</p>
+    } else if (serverStoppedResult){
```
lovettchris, line 62: I think this else case should come before the `(!curUri)` case.

vscode-lean4/package-lock.json
```diff
@@ -15,7 +15,7 @@
 				"cheerio": "^1.0.0-rc.10",
 				"mobx": "5.15.7",
 				"semver": "=7.3.5",
-				"vscode-languageclient": "=7.0.0"
+				"vscode-languageclient": "^8.0.0-next.14"
```
lovettchris, line 18: Please revert this language client version change

vscode-lean4/package-lock.json
```diff
@@ -24,7 +24,7 @@
 				"@types/node": "^17.0.7",
 				"@types/ps-node": "^0.1.1",
 				"@types/semver": "^5.3.30",
-				"@types/vscode": "^1.57.0",
+				"@types/vscode": "^1.61.0",
```
lovettchris, line 27: Please revert this vscode version change

vscode-lean4/package-lock.json
```diff
@@ -39,7 +39,7 @@
 				"webpack-cli": "^4.9.1"
 			},
 			"engines": {
-				"vscode": "^1.57.0"
+				"vscode": "^1.61.0"
```
lovettchris, line 42: Please revert this  version change

vscode-lean4/package-lock.json
```diff
@@ -4275,39 +4275,39 @@
...
-			"integrity": "sha512-P9AXdAPlsCgslpP9pRxYPqkNYV7Xq8300/aZDpO35j1fJm/ncize8iGswzYlcvFw5DQUx4eVk+KvfXdL0rehNg==",
+			"version": "8.0.0-next.14",
+			"resolved": "https://registry.npmjs.org/vscode-languageclient/-/vscode-languageclient-8.0.0-next.14.tgz",
+			"integrity": "sha512-NqjkOuDTMu8uo+PhoMsV72VO9Gd3wBi/ZpOrkRUOrWKQo7yUdiIw183g8wjH8BImgbK9ZP51HM7TI0ZhCnI1Mw==",
 			"dependencies": {
 				"minimatch": "^3.0.4",
-				"semver": "^7.3.4",
-				"vscode-languageserver-protocol": "3.16.0"
+				"semver": "^7.3.5",
+				"vscode-languageserver-protocol": "3.17.0-next.16"
```
lovettchris, line 4292: Please revert this  version change

vscode-lean4/package-lock.json
```diff
@@ -4275,39 +4275,39 @@
...
-			"resolved": "https://registry.npmjs.org/vscode-languageserver-protocol/-/vscode-languageserver-protocol-3.16.0.tgz",
-			"integrity": "sha512-sdeUoAawceQdgIfTI+sdcwkiK2KU+2cbEYA0agzM2uqaUy2UpnnGHtWTHVEtS0ES4zHU0eMFRGN+oQgDxlD66A==",
+			"version": "3.17.0-next.16",
+			"resolved": "https://registry.npmjs.org/vscode-languageserver-protocol/-/vscode-languageserver-protocol-3.17.0-next.16.tgz",
+			"integrity": "sha512-tx4DnXw9u3N7vw+bx6n2NKp6FoxoNwiP/biH83AS30I2AnTGyLd7afSeH6Oewn2E8jvB7K15bs12sMppkKOVeQ==",
 			"dependencies": {
-				"vscode-jsonrpc": "6.0.0",
-				"vscode-languageserver-types": "3.16.0"
+				"vscode-jsonrpc": "8.0.0-next.7",
+				"vscode-languageserver-types": "3.17.0-next.9"
```
lovettchris, line 4304: Please revert this  version change

vscode-lean4/package-lock.json
```diff
@@ -4275,39 +4275,39 @@
...
 			}
 		},
 		"node_modules/vscode-languageserver-types": {
-			"version": "3.16.0",
-			"resolved": "https://registry.npmjs.org/vscode-languageserver-types/-/vscode-languageserver-types-3.16.0.tgz",
-			"integrity": "sha512-k8luDIWJWyenLc5ToFQQMaSrqCHiLwyKPHKPQZ5zz21vM+vIVUSvsRpcbiECH4WR88K2XZqc4ScRcZ7nk/jbeA=="
+			"version": "3.17.0-next.9",
```
lovettchris, line 4308: Please revert this  version change

vscode-lean4/package.json
```diff
@@ -425,7 +425,7 @@
 		"cheerio": "^1.0.0-rc.10",
 		"mobx": "5.15.7",
 		"semver": "=7.3.5",
-		"vscode-languageclient": "=7.0.0"
+		"vscode-languageclient": "^8.0.0-next.14"
```
lovettchris, line 428: Please revert this  version change

vscode-lean4/package.json
```diff
@@ -5,7 +5,7 @@
 	"version": "0.0.70",
 	"publisher": "leanprover",
 	"engines": {
-		"vscode": "^1.57.0"
+		"vscode": "^1.61.0"
```
lovettchris, line 8: Please revert this  version change

vscode-lean4/package.json
```diff
@@ -434,7 +434,7 @@
 		"@types/node": "^17.0.7",
 		"@types/ps-node": "^0.1.1",
 		"@types/semver": "^5.3.30",
-		"@types/vscode": "^1.57.0",
+		"@types/vscode": "^1.61.0",
```
lovettchris, line 437: Please revert this  version change

vscode-lean4/src/infoview.ts
```diff
@@ -27,10 +27,10 @@ class RpcSession implements Disposable {
 
     constructor(client: LeanClient, public sessionId: string, public uri: ls.DocumentUri) {
         this.client = client;
-        this.keepAliveInterval = setInterval(() => {
+        this.keepAliveInterval = setInterval(async () => {
```
lovettchris, line 30: what is this fixing?

vscode-lean4/src/leanclient.ts
```diff
@@ -216,18 +217,18 @@ export class LeanClient implements Disposable {
                     return;
                 },
 
-                didChange: (data, next) => {
-                    next(data);
+                didChange: async (data, next) => {
```
lovettchris, line 220: this is probably due to the vscode version change, please revert this since Gabriel wants this to be done in a separate PR.

vscode-lean4/src/leanclient.ts
```diff
@@ -216,18 +217,18 @@ export class LeanClient implements Disposable {
...
+                    await next(data);
                     if (!this.running || !this.client) return; // there was a problem starting lean server.
                     const params = this.client.code2ProtocolConverter.asChangeTextDocumentParams(data);
                     this.didChangeEmitter.fire(params);
                 },
 
-                didClose: (doc, next) => {
+                didClose: async (doc, next) => {
```
lovettchris, line 227: this is probably due to the vscode version change, please revert this since Gabriel wants this to be done in a separate PR.

vscode-lean4/src/leanclient.ts
```diff
@@ -216,18 +217,18 @@ export class LeanClient implements Disposable {
...
                     if (!this.isOpen.delete(doc.uri.toString())) {
                         return;
                     }
-                    next(doc);
+                    await next(doc);
```
lovettchris, line 231: this is probably due to the vscode version change, please revert this since Gabriel wants this to be done in a separate PR.

vscode-lean4/src/leanclient.ts
```diff
@@ -270,7 +271,7 @@ export class LeanClient implements Disposable {
         )
         this.patchConverters(this.client.protocol2CodeConverter, this.client.code2ProtocolConverter)
         try {
-            this.client.onDidChangeState((s) =>{
+            this.client.onDidChangeState(async (s) =>{
```
lovettchris, line 274: this is probably due to the vscode version change, please revert this since Gabriel wants this to be done in a separate PR.

vscode-lean4/src/leanclient.ts
```diff
@@ -337,7 +339,7 @@ export class LeanClient implements Disposable {
             diag.fullRange = p2c.asRange(protDiag.fullRange)
             return diag
         }
-        p2c.asDiagnostics = (diags) => diags.map(d => p2c.asDiagnostic(d))
+        p2c.asDiagnostics = async (diags) => diags.map(d => p2c.asDiagnostic(d))
```
lovettchris, line 342: this is probably due to the vscode version change, please revert this since Gabriel wants this to be done in a separate PR.

vscode-lean4/src/leanclient.ts
```diff
@@ -346,7 +348,7 @@ export class LeanClient implements Disposable {
             protDiag.fullRange = c2p.asRange(diag.fullRange)
             return protDiag
         }
-        c2p.asDiagnostics = (diags) => diags.map(d => c2p.asDiagnostic(d))
+        c2p.asDiagnostics = async (diags) => diags.map(d => c2p.asDiagnostic(d))
```
lovettchris, line 351: this is probably due to the vscode version change, please revert this since Gabriel wants this to be done in a separate PR.

vscode-lean4/src/leanclient.ts
```diff
@@ -470,7 +471,7 @@ export class LeanClient implements Disposable {
     }
 
     // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-    sendNotification(method: string, params: any): void {
+    sendNotification(method: string, params: any): Promise<void> | undefined{
```
lovettchris, line 474: this is probably due to the vscode version change, please revert this since Gabriel wants this to be done in a separate PR.

vscode-lean4/src/leanclient.ts
```diff
@@ -491,10 +492,10 @@ export class LeanClient implements Disposable {
         return this.running && range ? this.client?.protocol2CodeConverter.asRange(range) : undefined;
     }
 
-    getDiagnosticParams(uri: Uri, diagnostics: readonly Diagnostic[]) : PublishDiagnosticsParams {
+    async getDiagnosticParams(uri: Uri, diagnostics: readonly Diagnostic[]) : Promise<PublishDiagnosticsParams> {
```
lovettchris, line 495: this is probably due to the vscode version change, please revert this since Gabriel wants this to be done in a separate PR.

vscode-lean4/src/leanclient.ts
```diff
@@ -491,10 +492,10 @@ export class LeanClient implements Disposable {
...
         const params: PublishDiagnosticsParams = {
             uri: this.convertUri(uri)?.toString(),
-            diagnostics: this.client?.code2ProtocolConverter.asDiagnostics(diagnostics as Diagnostic[]) ?? []
+            diagnostics: await this.client?.code2ProtocolConverter.asDiagnostics(diagnostics as Diagnostic[]) ?? []
```
lovettchris, line 498: this is probably due to the vscode version change, please revert this since Gabriel wants this to be done in a separate PR.

lovettchris (changes requested):

vscode-lean4/src/infoview.ts
marianaalanis93: There are errors that were marked in the build after upgrading the language client to 8 version. I'm going to revert this and put it in the PR of the language client update

vscode-lean4/package-lock.json
Vtec234: I would also be happy to merge that one first and then review this one once it is rebased on the new version, which would save the reverting.

lean4-infoview/src/infoview/main.tsx
marianaalanis93: Ok, thanks! fixed

lean4-infoview/src/infoview/messages.tsx
marianaalanis93: Ok, makes sense. I added this change because vscode suggested it (maybe because of vscode upgrade) I will revert this and wait for the final code after language client upgrade.

vscode-lean4/package-lock.json
marianaalanis93: You can find this on: #162

vscode-lean4/package-lock.json
marianaalanis93: You can find this on: #162

vscode-lean4/src/leanclient.ts
marianaalanis93: You can find this on: #162

lean4-infoview/src/infoview/info.tsx
```diff
@@ -271,8 +271,13 @@ function InfoAux(props: InfoProps) {
 
         function onError(err: any) {
             const errS = typeof err === 'string' ? err : JSON.stringify(err);
-            setError(`Error fetching goals: ${errS}`);
-            setStatus('error');
+            // we need to check if this value is empty or not, because maybe we are assigning
+            // a message error with an empty error
+            if (errS === '{}' || errS === undefined) { setError(undefined); }
```
lovettchris, line 276: tiny nit, I think this if test formatting would look better if the setError was on a newline.

lovettchris (approved):
Looks great, just one tiny code formatting comment.

lovettchris:
@Vtec234 , I think this looks good to go IMHO.

lean4-infoview/src/infoview/info.tsx
```diff
@@ -281,6 +286,7 @@ function InfoAux(props: InfoProps) {
             const [goals, termGoal] = await allReq;
             setGoals(goals);
             setTermGoal(termGoal);
+            onError(undefined)
```
Vtec234, line 289: What's the purpose of this call? `onError` is meant to be called, well, on error. But it seems like it's used to clear some state here?

lovettchris: yeah, it's a hack needed because the implementation of useEventResult is stateful.  I suppose we could handle it in main.tsx when we get a serverRestarted we could call setT on the state for serverStopped, but the setT function is hidden inside useEventResult so it gets a bit messy... it was easier to just do this...   but let us know if there's a cleaner way to do it.

Vtec234: Hmm, could you please elaborate on exactly which instance of `useEventResult` is causing trouble? If it is, then it sounds like there might indeed be a deeper issue.

lovettchris: Never mind I got my wires crossed, I'm referring to something else, namely, the line in `infoview.ts` where it calls `await this.webviewPanel?.api.serverStopped('');`  The info.tsx change I'm not so sure about, @marianaalanis93 perhaps you should take this one :-)

marianaalanis93: The reason we need to mark this onError as 'undefined' is that if the server dies indefinitely, a dict is generated with the generated error ({error here}). If we restart it later, this variable is left with an empty dict ({}), that's why we have to force it to undefined (as it was originally and before the server had any problem). It should be a problem at the moment we want to handle errors in main.tsx

Vtec234: Hm, I see. So I think this was quite confusing to me, because in my mind the `Info` component is supposed to maintain the invariant that the server is running. In particular, with the change in this PR when the server is stopped, in `main.tsx` we return the error message `Server unavailable. Please restart it.` and do not return `<Infos />` at all. This means that all the `<Info>` components will be unmounted and remounted when the server starts again, so they should not persist state, including the `error` state here.

lovettchris (changes requested):

lovettchris (approved):
Works great, I think this is ready to merge!

vscode-lean4/src/leanclient.ts
```diff
@@ -279,8 +281,15 @@ export class LeanClient implements Disposable {
                     console.log('client running, started in ', end - startTime, 'ms');
                     this.running = true; // may have been auto restarted after it failed.
                 } else if (s.newState === State.Stopped) {
+                    this.stoppedEmitter.fire('Lean language server has stopped. ');
                     console.log('client has stopped or it failed to start');
                     this.running = false;
+                    const restartItem = 'Restart Lean Language Server';
```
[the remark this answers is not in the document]
lovettchris: works now.

vscode-lean4/src/leanclient.ts
```diff
@@ -279,8 +281,15 @@ export class LeanClient implements Disposable {
...
+                    const item = await window.showErrorMessage('Please restart Lean language server.',  restartItem);
```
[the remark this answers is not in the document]
lovettchris: thanks

gebner:
> So apparently the whitespace that exists in "info.tsx" is getting stipped out entirely by our build tools

Yup: https://reactjs.org/docs/jsx-in-depth.html#string-literals-1

I'm not sure if `&nbsp;` is the best approach to solve this issue, it's literally a non-breaking space so it disables line-wrapping at that point.  There doesn't seem to be an official way to insert spaces, but we could use `{' '}` instead of `&nbsp;`.

lean4-infoview/src/infoview/info.tsx
gebner: Wojciech is right here, we never get to `Infos` with a stopped server.

vscode-lean4/src/leanclient.ts
```diff
@@ -417,10 +429,15 @@ export class LeanClient implements Disposable {
     async stop(): Promise<void> {
         assert(() => this.isStarted())
         if (this.client && this.running) {
-            this.stoppedEmitter.fire(undefined);
-            await this.client.stop()
+            this.noPrompt = true;
+            try {
+                await this.client.stop()
+            } catch (e) {
+                console.log(`Error stopping language client: ${e}`)
```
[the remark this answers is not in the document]
gebner: Good to hear!  Although I don't really see how.

vscode-lean4/src/utils/clientProvider.ts
```diff
@@ -29,6 +29,9 @@ export class LeanClientProvider implements Disposable {
     private clientRemovedEmitter = new EventEmitter<LeanClient>()
     clientRemoved = this.clientRemovedEmitter.event
 
+    private activeClientStoppedEmitter = new EventEmitter<string>()
+    activeClientStopped = this.activeClientStoppedEmitter.event
```
gebner, line 33:
```suggestion
    private clientStoppedEmitter = new EventEmitter<[LeanClient, string]>()
    clientStopped = this.clientStoppedEmitter.event
```

The extension can start more than LSP server process (= LeanClient = `lean --server`).  For example, if you open several projects in a workspace, each of these project directories will have a different server.  (This server process then in turn starts so-called worker processes, i.e., `lean --worker`, one for each file.  But the extension doesn't see that.)

So we should remember which LSP server stopped.

(I've just noticed that we already don't handle this cleanly in `Main`.  Apparently `Waiting for Lean server to start...` is only shown before the first server has started.  But not for the second one.  Ideally we'd show the message for the Lean server of the current file, of course.  We can fix that another time though.)

vscode-lean4/src/leanclient.ts
```diff
@@ -270,17 +273,26 @@ export class LeanClient implements Disposable {
...
-            this.client.onDidChangeState((s) =>{
+            this.client.onDidChangeState(async (s) => {
                 // see https://github.com/microsoft/vscode-languageserver-node/issues/825
                 if (s.newState === State.Starting) {
                     console.log('client starting');
                 } else if (s.newState === State.Running) {
                     const end = Date.now()
                     console.log('client running, started in ', end - startTime, 'ms');
                     this.running = true; // may have been auto restarted after it failed.
+                    return
```
gebner, line 284:
```suggestion
```

I don't think this `return` makes a difference.

gebner (commented):
@Vtec234 Initially I'd hoped we'd be able to get some feedback if the *worker* process died, like with:
```lean
#eval (unsafeCast 0 : String)
```
But from what I can tell there's no way to detect such a crash and the watchdog even ignores all requests afterwards, or did I miss something?

lean4-infoview/src/infoview/info.tsx
```diff
@@ -271,8 +271,15 @@ function InfoAux(props: InfoProps) {
 
         function onError(err: any) {
             const errS = typeof err === 'string' ? err : JSON.stringify(err);
-            setError(`Error fetching goals: ${errS}`);
-            setStatus('error');
+            // we need to check if this value is empty or not, because maybe we are assigning
+            // a message error with an empty error
+            if (errS === '{}' || errS === undefined) {
```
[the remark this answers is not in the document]
marianaalanis93: Ok thanks! committed

vscode-lean4/src/leanclient.ts
marianaalanis93: thanks! committed!

vscode-lean4/src/utils/clientProvider.ts
marianaalanis93: I'll commit the suggestion and we can handle this properly in Main in next iterations 👍

lovettchris: Actually, Gabriel is correct, I will work with Mariana on this one.

lovettchris:
> > So apparently the whitespace that exists in "info.tsx" is getting stipped out entirely by our build tools
> 
> Yup: https://reactjs.org/docs/jsx-in-depth.html#string-literals-1
> 
> I'm not sure if `&nbsp;` is the best approach to solve this issue, it's literally a non-breaking space so it disables line-wrapping at that point. There doesn't seem to be an official way to insert spaces, but we could use `{' '}` instead of `&nbsp;`.

Sure - and please add a comment as to why we are using `{' '}` so we don't forget.

lovettchris:
> ```lean
> #eval (unsafeCast 0 : String)
> ```

This is an interesting one - I notice the "worker" process dies but the "--server" process does not.  So Mariana's code doesn't kick in, we don't get "stopped" state from the LanguageClient.  I do see this pop up though - must be done by vscode?

image

And I do see this get printed to the debug console,  which comes from messages.tsx line 109, so it appears the infoview should be able to know about this and display something like this to the user:

getInteractiveDiagnostics error  {stack: 'Error: Server process for file:///d%3A/Temp/l…Trampoline (node:internal/async_hooks:130:17)', message: 'Server process for file:///d%3A/Temp/lean_ex…likely due to a stack overflow in user code.', code: -32603}

vscode-lean4/src/utils/clientProvider.ts
Vtec234: > the client the infoview cares about

The issue here is that the infoview needs to know the status of *all* clients, not just the active one. For example, using the 'pin' feature a user may pin an `Info` component from `A.lean` while also having `B.lean` open, and these two files can communicate with different clients.

But i would agree with Gabriel that this is a larger change that should be done in another PR, so I think leaving the `activeClientStoppedEmitter` as-is is good for now. For example, iirc the `RpcSession` manager does not really handle multiple clients at the moment.

lean4-infoview/src/infoview/info.tsx
lovettchris: Looks to me like this is clearing a possible error state created on line 297 in the catch block so that an error is not "sticky" and always passed to the <InfoDisplay constructor on line 306.

vscode-lean4/src/utils/clientProvider.ts
lovettchris: Then that is a huge new feature we should not try and do in this PR.  but I still maintain it is redundant for object A to raise an event with parameter A, the receiver of the event knows who it came from because added the event handler. 
 For example, notice that ClientProvider has a captured local variable named "client", so it doesn't need the client again in the event args.

Vtec234:
> But from what I can tell there's no way to detect such a crash and the watchdog even ignores all requests afterwards, or did I miss something?
> This is an interesting one - I notice the "worker" process dies but the "--server" process does not. 

Yes, the watchdog process is meant to never crash unless we have a bug in Lean core, whereas worker processes may crash given inputs such as the `unsafeCast`. The current [logic](https://github.com/leanprover/lean4/blob/4934104eaf879d56cdfbf14c50eaca4cf7539371/src/Lean/Server/Watchdog.lean#L319) in the watchdog attempts to transparently restart the worker and re-send all the in-flight requests that it can to the new worker. I'm not certain, but IIRC we don't send any kind of "worker crashed" message to the client.

vscode-lean4/src/utils/clientProvider.ts
Vtec234: > Then that is a huge new feature we should not try and do in this PR.

I agree.

> I still maintain it is redundant for object A to raise an event with parameter "this", the receiver of the event knows who it came from because it added the event handler.

Sorry, I do not understand. `activeClientStoppedEmitter` is an event on `LeanClientProvider` which stores multiple `LeanClient`s, so a `LeanClient` parameter does not seem redundant there (it is not `this : LeanClientProvider`).

Vtec234:
Having said all that, I took this PR to be fixing what happens when the *watchdog* is not started, whereas the restarting behaviour of file workers seems like a different issue altogether. So the issue of a worker that crashed on e.g. `unsafeCast` and then restarts transparently seems orthogonal.

vscode-lean4/src/utils/clientProvider.ts
lovettchris: Woops you are right, I was confusing it with `LeanClient` which has `private stoppedEmitter = new EventEmitter<string>()`. And I now understand what you mean by info view showing multiple cursor locations from potentially different LeanClients because of the "pinning" feature as shown below.  I will work with Mariana on a solution that handles this properly.  Clearly `activeClientStoppedEmitter` is not sufficient.  It needs to be `clientStoppedEmitter` and the `this.webviewPanel?.api.serverStopped` probably needs to pass something about the server, for example, should the `pinned` state now say `client stopped` ?  Or should it stay alive pinned or perhaps it should move to "paused" ?  And if user tries to unpause it then it says `client stopped` ?

image

lovettchris:
> Having said all that, I took this PR to be fixing what happens when the _watchdog_ (i.e. `lean --server`) is not started, whereas the restarting behaviour of file workers seems like a different issue altogether. So the issue of a worker that crashed on e.g. `unsafeCast` and then restarts transparently seems orthogonal.

I agree, I filed this as a [new issue #175](https://github.com/leanprover/vscode-lean4/issues/175), just so we can get this one finished sooner than later.

vscode-lean4/src/utils/clientProvider.ts
gebner: We need to keep track of which server stopped even if you don't pin anything.  Instead of pinning information from another file, you could just switch back-and-forth between files.  You'd expect the "server stopped" message to update when you switch files.

> should the `pinned` state now say `client stopped` ? Or should it stay alive pinned or perhaps it should move to "paused" ? And if user tries to unpause it then it says `client stopped` ?

I think it should just say "(server stopped)" (or even shorter: "(crashed)").  Changing state shouldn't be necessary.

lovettchris: Good, point, right now if I load the above multi-workspace test, and edit Foo.lean then Main.lean and then kill the worker process for -foo, then switch back to Foo.lean it does not show "server stopped unexpectedly", in fact our LeanClient doesn't even get notified of any of that because of https://github.com/leanprover/vscode-lean4/issues/175 and the --server process auto-restarts the server for Foo workspace and when you click somewhere else in the file it starts working again.  Can you think of another repro for the issue you are raising that is not #175 ?

Vtec234: @lovettchris I think the repro would be to kill the `--server` (watchdog) process for one of the workspaces, not the worker.

lovettchris: Ah, of course, ok, that one pops up this error message:

image

But the InfoView for Foo.lean doesn't say "Lean language server has stopped" until I edit something.  So great, this is the repro:

1. Load multi-folder workspace 
2. Edit Foo.lean from workspace "Foo"
3. Edit Main.lean from workspace "Main"
4. Kill the --server for the "Foo" workspace and do not click the "Restart Lean Language Server" button that pops up.
5. Switch to Foo.lean.

Notice now that "Infoview" does not say "Lean language server has stopped" until you "edit" the file - which is a bug.

lean4-infoview/src/infoview/info.tsx
```diff
@@ -166,10 +166,10 @@ export function InfoDisplay(props0: InfoDisplayProps) {
             </div>
             {nothingToShow && (
                 isPaused ?
-                    <span>Updating is paused.
+                    <span>Updating is paused.{' '}
```
lovettchris, line 169: Can you add a comment here explaining why we need to do the `{' '}` trick?  Something like whitespace is stripped by the typescript compiler here which causes the anchor tags to show up with no space around them, adding a typescript expression containing a string literal makes it something that is not stripped automatically.

marianaalanis93: Sure :)

vscode-lean4/src/utils/clientProvider.ts
marianaalanis93: Resolved

lean4-infoview/src/infoview/info.tsx
gebner: I think I got a bit sidetracked by the `<Infos>` can't be reached if the server is stopped comment.

> If we restart it later, this variable is left with an empty dict ({}), that's why we have to force it to undefined (as it was originally and before the server had any problem).

Did it cause any issues that `error` was set to `{}`?  The idea is that the `error` variable is completely ignored if the status is not `'error'`.  And the end of the `triggerUpdate` function sets the status to `'ready'`.

If we want to reset the `error` variable, then we should just call `setError(undefined)` directly.  There's no need to complicate the `onError` function for this.  (Because the `onError` function is also meant to transition into the `'error'` state.)

lovettchris: I agree `setError(undefined)` is better here.  Mariana, can you try that?

marianaalanis93:
@gebner solved this discussion: https://github.com/leanprover/vscode-lean4/pull/159#discussion_r872342496

lean4-infoview/src/infoview/info.tsx
```diff
@@ -273,8 +274,15 @@ function InfoAux(props: InfoProps) {
...
         function onError(err: any) {
             const errS = typeof err === 'string' ? err : JSON.stringify(err);
-            setError(`Error fetching goals: ${errS}`);
-            setStatus('error');
+            // we need to check if this value is empty or not, because maybe we are assigning
+            // a message error with an empty error
+            if (errS === '{}' || errS === undefined) {
+                setError(undefined);
+            }
+            else {
```
gebner, line 282:
```suggestion
```

We can now revert this part since we call `setError(undefined)` directly, right?

marianaalanis93: Not really, there is an issue here. When the server crashes and then is restarted the variable 'errS' stays in a weird state and it means that after we restart the server, we can't see the 'No info found' message in the infoview (when we put the cursor in a blank space). That's why we need to check if 'errS' is {} or undefined and set the error as 'undefined'

image

lovettchris:
@gebner, this should be good to go.

lean4-infoview/src/infoview/info.tsx
gebner: Hmm, this is weird, but at least we show the right error message now..

gebner:
Thanks!

PR #163, opened 2022-04-07
Update on-push file for Windows workflow

marianaalanis93:
This PR contains:

- Update on-push.yml file for Windows workflow in order to specify the branch to execute on_pull and on_pull_request workflows.


Vtec234:
Thanks! The move from `NoMenu` to `NoPrompt` looks good, but it looks like most of the failures are due to a non-deterministic test issue on Windows that just happens to pass sometimes (e.g. in the latest run). There is either a bug in the infoview itself or in the test suite. We see `AssertionError [ERR_ASSERTION]: Missing "master" in infoview` in the failed cases, i.e. the infoview is not displaying the expected output at the time the output is checked.

Vtec234:
Let's merge this since we should match the new `elan` command line in any case, and resolve the non-deterministic failures in another PR.

PR #167, opened 2022-04-16
feat: shim ESM importmaps

### LINKED ISSUE #166: Broken infoview in Gitpod with 0.0.70+
Starting with version 0.0.69, the infoview is empty on gitpod.io, presumably due to this error in the console:
```
Uncaught TypeError: Error resolving module specifier “@lean4/infoview”. Relative module specifiers must start with “./”, “../” or “/”.
```
Example: https://gitpod.io/#https://github.com/Kha/tba-2022

Vtec234:
Closes #166 . We include [`es-module-shims`](https://github.com/guybedford/es-module-shims) in `media/` directly since it is only on a custom CDN but is not on NPM.


Vtec234:
I will merge this because it may be needed for a demo this week.

Kha:
Thanks!

lovettchris:
@Vtec234 cool, can we get es-module-shims.js from https://www.npmjs.com/package/es-module-shims instead?  Seems it would be easier to track new versions that way...

Vtec234:
@lovettchris hm, I don't remember why I wrote 'it's not on NPM' even though it actually is. I may have just missed it. Note that the [version](https://ga.jspm.io/npm:es-module-shims@1.5.8/dist/es-module-shims.js) recommended in their README is different than [the NPM version](https://unpkg.com/es-module-shims@1.5.8/dist/es-module-shims.js), but it might just be minified. If you can get the NPM one working, it would be better. Since the point is to make Gitpod work, that would be the thing to test.

PR #171, opened 2022-05-03
feat: change "⊢" in conv goals to "|"

KaseQuark:
In conv mode, we rewrite (and thus change) the original goal, the "|" signifies this. This is similar to Lean3.

Suggested by @javra.

### DIFF
--- a/lean4-infoview/src/infoview/goals.tsx
+++ b/lean4-infoview/src/infoview/goals.tsx
@@ -30,6 +30,8 @@ export function goalsToString(goals: InteractiveGoals): string {
 }
 
 export function Goal({pos, goal}: {pos: DocumentPosition, goal: InteractiveGoal}) {
+    let prefix = '⊢'
+    if (goal.goalPrefix === '|') {prefix = '|'}
     return <div className="font-code tl pre-wrap">
         <ul className="list pl0">
             {goal.userName && <li key={'case'}><strong className="goal-case">case </strong>{goal.userName}</li>}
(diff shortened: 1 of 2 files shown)


lean4-infoview/src/infoview/rpcInterface.ts
```diff
@@ -74,6 +74,7 @@ export interface InteractiveGoal {
     hyps: InteractiveHypothesis[]
     type: CodeWithInfos
     userName?: string
+    goalPrefix: string
```
gebner, line 77:
```suggestion
    goalPrefix?: string
```
The property is not present in older Lean versions (or any Lean version yet for that matter).

lean4-infoview/src/infoview/goals.tsx
```diff
@@ -30,6 +30,8 @@ export function goalsToString(goals: InteractiveGoals): string {
 }
 
 export function Goal({pos, goal}: {pos: DocumentPosition, goal: InteractiveGoal}) {
+    let prefix = '⊢'
+    if (goal.goalPrefix === '|') {prefix = '|'}
```
gebner, line 34:
```suggestion
    const prefix = goal.goalPrefix ?? '⊢'
```

gebner (commented):
In any case, this must wait for https://github.com/leanprover/lean4/pull/1141

lean4-infoview/src/infoview/rpcInterface.ts
Vtec234: To elaborate, the issue is that while we can expect users to always have an up-to-date `vscode-lean4` thanks to the auto-update mechanism, they might be using arbitrary server versions, so we commit to supporting any server version up to a few months old or so. So we cannot expect this property to be present in `InteractiveGoal`s arriving from the server, and it must be optional.

Issue #170, opened 2022-04-20
Intermittent test failures


gebner:
Sometimes the tests fail, but work again after restarting:
- https://github.com/leanprover/vscode-lean4/actions/runs/2186659870/attempts/1
- https://github.com/leanprover/vscode-lean4/actions/runs/2194375381/attempts/1

Vtec234:
The second of these is an instance of #165 . The first one I have not seen before.

lovettchris:
I believe this is fixed now.

Vtec234:
We are still seeing them, unfortunately. For example [here](https://github.com/leanprover/vscode-lean4/runs/6528944512?check_suite_focus=true).

lovettchris:
> We are still seeing them, unfortunately. For example [here](https://github.com/leanprover/vscode-lean4/runs/6528944512?check_suite_focus=true).

are you sure your branch is in sync with master?

Vtec234:
It was not! Thank you for pointing it out.

gebner:
And they seem to be back: https://github.com/leanprover/vscode-lean4/runs/6691857368?check_suite_focus=true

lovettchris:
I think this might be a new bug the output contains this, I haven't seen this one before.  Do you recognize this "incorrect position" language from the LSP side?

```
version.lean:3:4
Error updating: Error fetching goals: {"stack":"Error: Incorrect position 'file:///d%3A/a/vscode-lean4/vscode-lean4/vscode-lean4/test/test-fixtures/simple/test/version.lean:2:4' in RPC call\n\tat d:\\a\\vscode-lean4\\vscode-lean4\\vscode-lean4\\dist\\extension.js:2:542515\n\tat d:\\a\\vscode-lean4\\vscode-lean4\\vscode-lean4\\dist\\extension.js:2:542809\n\tat Immediate.<anonymous> (d:\\a\\vscode-lean4\\vscode-lean4\\vscode-lean4\\dist\\extension.js:2:543171)\n\tat processImmediate (node:internal/timers:464:21)","message":"Incorrect position 'file:///d%3A/a/vscode-lean4/vscode-lean4/vscode-lean4/test/test-fixtures/simple/test/version.lean:2:4' in RPC call","code":-32602}. Try again.
All Messages (1)
version.lean:1:0
invalid output from `c:\Users\runneradmin\.elan\toolchains\leanprover--lean4---nightly\bin\lake.exe print-paths Init Lake`:

stderr:
```

[...]

PR #172, opened 2022-05-05
increase retries and fix launch.json test launching configurations.

lovettchris:
Possible fix for Intermittent test failures #170


gebner:
Hmm, this still failed directly after merging.

lovettchris:
yep, this time the toolchain test on "adhoc" files failed, so does indeed seem random, I'm investigating some more...

Issue #90, opened 2021-12-13
Infoview doesn't show errors when lake fails


gebner:
If you have two files like this:
```lean
-- A.lean
does not compile

-- B.lean
import A
```

And put the cursor on the first line of `B`, then the infoview doesn't show the error diagnostic for the import error.  (Presumably because we also request interactive diagnostics, and treat the resulting error as "no messages found").

lovettchris:
@gebner - interesting, I see a red squiggly on import A and the tool tip matches what you see below, this message flashes real quick in the infoview then it disappears.

![image](https://user-images.githubusercontent.com/18707114/149026744-5df543d1-76be-4e03-b127-d8696e3caf41.png)

and I think it is because the call to 'Lean.Widget.getInteractiveDiagnostics' returns [] which clears the infoview.  So the infoview must be showing the red squiggle info first, then calls getInteractiveDiagnostics and that wipes the message.  Is this a Lean LSP bug ?

marianaalanis93:
I was able to reproduce this issue. I will check the possible solution.

gebner:
> Is this a Lean LSP bug ?

In some sense, maybe.  I guess you could argue that they are not "real" Lean diagnostics, and `getInteractiveDiagnostics` should only return diagnostics created by the Lean compiler.  In either case, this part of the RPC API is underspecified.

leodemoura:
@gebner Is it clear to you what needs to be done to address this issue?

gebner:
Yes, I just need to close it. :smile:  Mariana fixed this in #173.

PR #173, opened 2022-05-13
[issue-90] Infoview doesn't show errors when lake fails

marianaalanis93:
This PR contains a fix related with #90 Infoview doesn't show errors when lake fails. In this case the message returns 'nothing' instead of undefined in some cases.


gebner:
You're on the right track with the general idea.  If the `getInteractiveDiagnostics` call returns the empty array, then we should fall back to the LSP diagnostics.  That should solve the issue in #90.

However this is the wrong place to add this rule.  The `rpcInterface.ts` file contains lightweight wrappers.  These wrappers are there to 1) give the correct typescript types, and 2) register the result with the garbage collector.  They shouldn't modify the result.

The proper place to add it is in `messages.tsx`, where the `getInteractiveDiagnostics` function is called.

lovettchris (approved):

lean4-infoview/src/infoview/messages.tsx
```diff
@@ -177,7 +177,8 @@ export function useMessagesForFile(uri: DocumentUri, line?: number): Interactive
             try {
                 const diags = await getInteractiveDiagnostics(rs, { uri, line: 0, character: 0 },
                     line ? { start: line, end: line + 1 } : undefined)
-                if (diags) {
+                if (diags && diags.length > 0) {
+                    // diags may be '' in some cases
```
gebner, line 181:
```suggestion
                    // diags may be [] when lake fails
```

There's also another one on line 101 for the "All Messages" section.

marianaalanis93: Ok checking :)

Vtec234 (approved):
Nice, thanks!

Vtec234:
Thanks again! By the way, could you please follow the lean4 [commit convention](https://github.com/leanprover/lean4/blob/master/doc/dev/commit_convention.md) (which I guess we should also link somewhere in this repo)? In particular, it's better to state what is being changed in the commit title rather than what problem is that the commit solves. E.g. even something like "fix: show errors when lake fails" instead of "Infoview doesn't show errors when lake fails". You can then also link the relevant issue in the commit message like `git commit -m "fix: show errors when lake fails" -m "Fixes #90."`.

PR #178, opened 2022-05-18
fix: elan installation on Unix

Kha:
`||` and `&&` have the same precedence in bash


Kha:
Yes, correct :) . It even went on to start the server successfully after pressing Enter.

PR #179, opened 2022-05-18
found a fix for flakey tests

lovettchris:
Some timing conditions can happen while running unit tests that cause the LanguageClient stop method to throw an exception which then causes those tests to fail.  We really don't care if stop fails, so I added a try/catch and this seems to help make the tests less flakey.

Note the first workflow run failed, but that is with an unrealted error: `Failed to get VS Code archive location`.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -132,6 +132,17 @@
             "cwd": "${workspaceFolder}/vscode-lean4/out/",
             "outFiles": ["${workspaceFolder}/vscode-lean4/out/test/suite/**/*.js"],
             "preLaunchTask": "watchTest"
-        }
+        },
+        {
+            "name": "Launch runTests.js",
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -132,6 +132,17 @@
             "cwd": "${workspaceFolder}/vscode-lean4/out/",
             "outFiles": ["${workspaceFolder}/vscode-lean4/out/test/suite/**/*.js"],
             "preLaunchTask": "watchTest"
-        }
+        },
+        {
+            "name": "Launch runTests.js",
```
Vtec234, line 137: Did you mean to add this launch config?

lovettchris: yes, turns out it is handy to be able to debug this file also.

gebner:
Ok, let's see if this fixes the failures.

PR #180, opened 2022-05-18
feat: add go-to-definition to infoview

Vtec234:
- Add go-to-definition to the infoview with a ctrl+click UI mimicking VSCode.

https://user-images.githubusercontent.com/13901751/169145205-15d6cfe8-8fa6-422e-ba51-e1ed36fcf756.mp4
- Add `subexprPos` to `CodeToken` to know subexpression positions for visual tactics like interactive `rw` and `conv`.
- ~~Put RPC calls in a namespace and remove inconsistent namespace prefixes.~~

This needs a server PR (leanprover/lean4#1157).

### DIFF
--- a/lean4-infoview/src/infoview/interactiveCode.tsx
+++ b/lean4-infoview/src/infoview/interactiveCode.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react'
 
-import { RpcContext } from './contexts'
+import { EditorContext, RpcContext } from './contexts'
 import { DocumentPosition } from './util'
-import { CodeToken, CodeWithInfos, InfoPopup, InfoWithCtx, InteractiveDiagnostics_infoToInteractive, TaggedText } from './rpcInterface'
+import { CodeToken, CodeWithInfos, InfoPopup, InfoWithCtx, InteractiveDiagnostics_infoToInteractive, Lean_Widget_getGoToLocation, TaggedText } from './rpcInterface'
 import { DetectHoverSpan, HoverState, WithTooltipOnHover } from './tooltips'
+import { Location } from 'vscode-languageserver-protocol'
 
 export interface InteractiveTextComponentProps<T> {
   pos: DocumentPosition
@@ -76,16 +77,31 @@ function InteractiveCodeTag({pos, tag: ct, fmt}: InteractiveTagProps<CodeToken>)
         redrawTooltip={redrawTooltip} />
     </div>, [pos.uri, pos.line, pos.character, ct.info])
 
+  // We mimick the VSCode ctrl-hover and ctrl-click UI for go-to-definition
+  const rs = React.useContext(RpcContext)
+  const ec = React.useContext(EditorContext);
+  const [goToLoc, setGoToLoc] = React.useState<Location | undefined>(undefined)
   const [hoverState, setHoverState] = React.useState<HoverState>('off')
+
   return (
-    <WithTooltipOnHover tooltipContent={mkTooltip} onClick={e => {
-      if (e.ctrlKey) {
-        console.log("gotodef", fmt)
-      }
+    <WithTooltipOnHover mkTooltipContent={mkTooltip} onClick={(e, next) => {
+      // On ctrl-click, if location is known, go to it in the editor
+      if (e.ctrlKey && goToLoc !== undefined) ec.revealPosition({ uri: goToLoc.uri, ...goToLoc.range.start })
+      if (!e.ctrlKey) next(e)
     }}>
       <DetectHoverSpan
-        setHoverState={setHoverState}
-        className={'highlightable ' + (hoverState !== 'off' ? 'highlight ' : '') + (hoverState === 'ctrlOver' ? 'underline ' : '')}
+        setHoverState={st => {
+          // On ctrl-hover, fetch the go-to location
+          if (st === 'ctrlOver') {
+            void Lean_Widget_getGoToLocation(rs, pos, ct.info, 'definition').then(loc => {
+              if (loc) setGoToLoc(loc)
+            }).catch(e => console.error('Error in go-to-definition: ', e.toString()))
+          }
+          setHoverState(st)
+        }}
+        className={'highlightable '
+                    + (hoverState !== 'off' ? 'highlight ' : '')
+                    + (hoverState === 'ctrlOver' && goToLoc !== undefined ? 'underline ' : '')}
       >
         <InteractiveCode pos={pos} fmt={fmt} />
       </DetectHoverSpan>
--- a/lean4-infoview/src/infoview/interactiveCode.tsx
+++ b/lean4-infoview/src/infoview/interactiveCode.tsx
@@ -98,29 +98,6 @@ function InteractiveCodeTag({pos, tag: ct, fmt}: InteractiveTagProps<SubexprInfo
     return undefined
   }, [rs, pos.uri, pos.line, pos.character, ct.info, goToLoc])
 
-  React.useEffect(() => {
-    const onKeyDown = (e : KeyboardEvent) => {
-      if (e.key === 'Control')
-        setHoverState(st => {
-          const newst = st === 'over' ? 'ctrlOver' : st
-          if (newst === 'ctrlOver') void fetchGoToLoc()
-          return newst
-        })
-    }
-
-    const onKeyUp = (e : KeyboardEvent) => {
-      if (e.key === 'Control')
-        setHoverState(st => st === 'ctrlOver' ? 'over' : st)
-    }
-
-    document.addEventListener('keydown', onKeyDown)
-    document.addEventListener('keyup', onKeyUp)
-    return () => {
-      document.removeEventListener('keydown', onKeyDown)
-      document.removeEventListener('keyup', onKeyUp)
-    }
-  }, [fetchGoToLoc])
-
   return (
     <WithTooltipOnHover
       mkTooltipContent={mkTooltip}
@@ -132,16 +109,6 @@ function InteractiveCodeTag({pos, tag: ct, fmt}: InteractiveTagProps<SubexprInfo
           void ec.revealPosition({ uri: loc.uri, ...loc.range.start })
         })
         if (!e.ctrlKey) next(e)
-      }}
-      onPointerMove={e => {
-        if (e.ctrlKey)
-          setHoverState(st => {
-            const newst = st === 'over' ? 'ctrlOver' : st
-            if (newst === 'ctrlOver') void fetchGoToLoc()
-            return newst
-          })
-        else
-          setHoverState(st => st === 'ctrlOver' ? 'over' : st)
       }}
     >
       <DetectHoverSpan
--- a/lean4-infoview/src/infoview/interactiveCode.tsx
+++ b/lean4-infoview/src/infoview/interactiveCode.tsx
@@ -86,7 +86,7 @@ function InteractiveCodeTag({pos, tag: ct, fmt}: InteractiveTagProps<CodeToken>)
   return (
     <WithTooltipOnHover mkTooltipContent={mkTooltip} onClick={(e, next) => {
       // On ctrl-click, if location is known, go to it in the editor
-      if (e.ctrlKey && goToLoc !== undefined) ec.revealPosition({ uri: goToLoc.uri, ...goToLoc.range.start })
+      if (e.ctrlKey && goToLoc !== undefined) void ec.revealPosition({ uri: goToLoc.uri, ...goToLoc.range.start })
       if (!e.ctrlKey) next(e)
     }}>
       <DetectHoverSpan
--- a/lean4-infoview/src/infoview/interactiveCode.tsx
+++ b/lean4-infoview/src/infoview/interactiveCode.tsx
@@ -79,14 +79,14 @@ function InteractiveCodeTag({pos, tag: ct, fmt}: InteractiveTagProps<SubexprInfo
 
   // We mimick the VSCode ctrl-hover and ctrl-click UI for go-to-definition
   const rs = React.useContext(RpcContext)
-  const ec = React.useContext(EditorContext);
+  const ec = React.useContext(EditorContext)
   const [hoverState, setHoverState] = React.useState<HoverState>('off')
 
   const [goToLoc, setGoToLoc] = React.useState<Location | undefined>(undefined)
-  const fetchGoToLoc = async () => {
+  const fetchGoToLoc = React.useCallback(async () => {
     if (goToLoc !== undefined) return goToLoc
     try {
-      const lnks = await getGoToLocation(rs, pos, 'definition', ct.info);
+      const lnks = await getGoToLocation(rs, pos, 'definition', ct.info)
       if (lnks !== undefined && lnks.length > 0) {
         const loc = { uri: lnks[0].targetUri, range: lnks[0].targetSelectionRange }
         setGoToLoc(loc)
@@ -96,7 +96,30 @@ function InteractiveCodeTag({pos, tag: ct, fmt}: InteractiveTagProps<SubexprInfo
       console.error('Error in go-to-definition: ', JSON.stringify(e))
     }
     return undefined
-  }
+  }, [rs, pos.uri, pos.line, pos.character, ct.info, goToLoc])
+
+  React.useEffect(() => {
+    const onKeyDown = (e : KeyboardEvent) => {
+      if (e.key === 'Control')
+        setHoverState(st => {
+          const newst = st === 'over' ? 'ctrlOver' : st
+          if (newst === 'ctrlOver') void fetchGoToLoc()
+          return newst
+        })
+    }
+
+    const onKeyUp = (e : KeyboardEvent) => {
+      if (e.key === 'Control')
+        setHoverState(st => st === 'ctrlOver' ? 'over' : st)
+    }
+
+    document.addEventListener('keydown', onKeyDown)
+    document.addEventListener('keyup', onKeyUp)
+    return () => {
+      document.removeEventListener('keydown', onKeyDown)
+      document.removeEventListener('keyup', onKeyUp)
+    }
+  }, [fetchGoToLoc])
 
   return (
     <WithTooltipOnHover mkTooltipContent={mkTooltip} onClick={(e, next) => {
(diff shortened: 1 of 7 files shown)


gebner:
> Put RPC calls in a namespace and remove inconsistent namespace prefixes.

Can you split this in two PRs please?  These unrelated cleanups make it hard to see the actual changes.

lean4-infoview/src/infoview/interactiveCode.tsx
```diff
@@ -68,18 +69,43 @@ function TypePopupContents({pos, info, redrawTooltip}: {pos: DocumentPosition, i
...
+  const ec = React.useContext(EditorContext);
+  const [goToLoc, setGoToLoc] = React.useState<Location | undefined>(undefined)
+  const [hoverState, setHoverState] = React.useState<HoverState>('off')
+
   return (
-    <WithTooltipOnHover tooltipContent={mkTooltip}>
-      <HighlightOnHoverSpan>
+    <WithTooltipOnHover mkTooltipContent={mkTooltip} onClick={(e, next) => {
+      // On ctrl-click, if location is known, go to it in the editor
+      if (e.ctrlKey && goToLoc !== undefined) void ec.revealPosition({ uri: goToLoc.uri, ...goToLoc.range.start })
```
lovettchris, line 89: This works, but only if I move the mouse into the token AFTER holding control key.  Can you also add an explicit keydown event handler so it works consistently with how it works in the text editor ?

Vtec234: Good point!

Vtec234: Done. I didn't want the overhead of every single subexpression registering an event handler, so I made it only react when clicked. The UI is almost the same, with the minor difference that the underline doesn't show up when ctrl-clicked but not ctrl-hovered.

lovettchris:
Is this a bug? I tried editing this file `lean4\script\reformat.lean` and the 'main' function behaves weirdly...

image

lovettchris:
Another bug, after playing with it in the info view for a bit (same file reformat.lean) I start getting these errors in the popup:

```
Error: Cannot decode params in RPC call 'Lean.Widget.InteractiveDiagnostics.infoToInteractive({"p":"85"})'
RPC reference '85' is not valid
```
(and yes, I'm running your latest Lean bits from `Vtec234:widget-gotodef`).

lovettchris:
Another bug, I'm not sure why we added this code in `infoview.ts showDocument`:
```
            const client = this.clientProvider.findClient(show.uri);
            if (!client?.running) return;
```
but it stops a bunch of your goto definition requests from working where it needs for example to jump to a Lean source file where there is no LeanClient for that file open yet.  I can see the client constraint is because it is trying to reuse the protocol2CodeConverter that lives in the LeanClient class to convert the selection range.  So we'll need to find another way to do that - perhaps lift the whole protocol2CodeConverter out as a helper class (problem is it comes from BaseLanguageClient - but I assume any BaseLanguageClient will do, the conversion is probably the same for all of them?)

lean4-infoview/src/infoview/interactiveCode.tsx
```diff
@@ -68,18 +69,57 @@ function TypePopupContents({pos, info, redrawTooltip}: {pos: DocumentPosition, i
...
     </div>, [pos.uri, pos.line, pos.character, ct.info])
+
+  // We mimick the VSCode ctrl-hover and ctrl-click UI for go-to-definition
+  const rs = React.useContext(RpcContext)
+  const ec = React.useContext(EditorContext);
+  const [hoverState, setHoverState] = React.useState<HoverState>('off')
+
+  const [goToLoc, setGoToLoc] = React.useState<Location | undefined>(undefined)
+  const fetchGoToLoc = async () => {
+    if (goToLoc !== undefined) return goToLoc
```
lovettchris, line 87: could the cached value ever be wrong? for example, after a Refresh File Dependencies command?  Do we need to cache?

lean4-infoview/src/infoview/rpcInterface.ts
```diff
@@ -152,3 +154,13 @@ export async function InteractiveDiagnostics_msgToInteractive(rs: RpcSessions, p
     if (ret) TaggedMsg_registerRefs(rs, pos, ret)
     return ret
 }
+
+export type GoToKind = 'definition' | 'declaration' | 'type'
```
lovettchris, line 158: Your requesthandling.lean defines GoToKind in a different order `| declaration | definition | type`

lovettchris:
> I can sometimes see this error flash very briefly (as in for milliseconds), but I'm not able to reproduce it showing up and persisting. Do you have a test case I could use to reproduce this?

I don't have the repro handy, but in my case the error was permanent.  But I wonder if the reason was I jumped to a Lean source file and toolchain used there was different and did not have your new `Lean.Widget.getGoToLocation` implementation.

Vtec234:
> So we'll need to find another way to do that - perhaps lift the whole protocol2CodeConverter out as a helper class

I think this is the way to go. Looking at the [implementation](https://github.com/microsoft/vscode-languageserver-node/blob/1da58fe99a92c334450bd01b8956104962005ab0/client/src/common/protocolConverter.ts#L387) of `asRange`, there is nothing there that depends on the client state. I suggest we copy the implementation of all the pure-functional conversions that we use into a `conversions` file and drop all the `client.protocol2CodeConverter` stuff. It should simplify `leanclient.ts`.

lean4-infoview/src/infoview/interactiveCode.tsx
Vtec234: Hm, after a refresh the whole goal state and every message would be recreated. The `InteractiveCode` component usually has a much shorter lifetime. The caching is just for UX - by starting to fetch as soon as the mouse enters the area we can often immediately resolve to target when it is clicked.

lovettchris:
Would be cool if I could ctrl click on "←" and jump to `def doIfLetBind`

lean4-infoview/src/infoview/interactiveCode.tsx
lovettchris: Got it.

Kha:
> Would be cool if I could ctrl click on "←" and jump to `def doIfLetBind`

See also https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Custom.20tooltips.20from.20macros/near/275926969

lovettchris:
InfoView Control key click works, but when I hit the ctrl key it does not show underline like it does in the VS code editor.

lovettchris:
When I hover my way into a tooltip chain then Ctrl+Click the "Type 4" it does a goto def on the original type (which was my own custom Point structure in this case).  This seems a bit odd... couldn't there be real things in these chained popups one might want to navigate to?

image

Vtec234:
> When I hover my way into a tooltip chain then Ctrl+Click the "Type 4" it does a goto def on the original type 

Good catch, thanks!

> when I hit the ctrl key it does not show underline like it does in the VS code editor

I am happy with the current UI, I don't think it's necessary to try to recreate exactly in every detail how VSCode works.

lovettchris:
> I am happy with the current UI, I don't think it's necessary to try to recreate exactly in every detail how VSCode works.

I'm a perfectionist :-)  This git patch handles the case when infoview has the keyboard focus:
[keydown.zip](https://github.com/leanprover/vscode-lean4/files/8734731/keydown.zip)

gebner:
> I am happy with the current UI, I don't think it's necessary to try to recreate exactly in every detail how VSCode works.

I agree with Chris, if the underline only appears in certain circumstances then it is more confusing than it helps.  Right now it only appears when you move into the expression with ctrl already pressed, or after you've clicked.

> I start getting these errors in the popup:

Speaking of errors, I've seen some (noncritical) issues when the go-to-definition goes to a different file.
1. The goal display in the infoview is not updated (still shows the goal from the old file).
2. An error message is printed in the tooltip for a brief moment before it closes.
3. Going to a previously unopened file prints `getInteractiveDiagnostics error  Error: already connecting or connected at 'file:///...'`

I imagine the error Chris is seeing could be because of the mismatch between position and goal in (1).

Vtec234:
> I agree with Chris, if the underline only appears in certain circumstances then it is more confusing than it helps. Right now it only appears when you move into the expression with ctrl already pressed, or after you've clicked.

I added the keydown/up handlers, but they don't work anyway. It looks like the webview is not receiving these events unless it is focused (i.e. you click somewhere in the infoview first). If you just move the mouse from the focused source editor to the infoview, no event is received. Could you check if this is also the case on your setup @lovettchris?

gebner:
Can we maybe hook into onmousemove?

Vtec234:
> Can we maybe hook into onmousemove?

Yes, this works although one has to move the mouse to see the state update. Should we keep the keydown/up handlers? Since they are only relevant in one case (infoview focused with previous click, ctrl pressed after having hovered over the element already, mouse isn't moved), I am inclined to remove them.

lovettchris:
> > Can we maybe hook into onmousemove?
> 
> Yes, this works although one has to move the mouse to see the state update. Should we keep the keydown/up handlers? Since they are only relevant in one case (infoview focused with previous click, ctrl pressed after having hovered over the element already, mouse isn't moved), I am inclined to remove them.

Yes I said with my patch it only "handles the case when infoview has the keyboard focus" - which happens as soon as yo click anything in the infoview (like pinning, or a button press, or a keyboard expand collapse of a section and so on).  I'm also looking for a way to funnel keyboard ctrl key state changes from the vscode extension in the other case where the editor still has the keyboard focus.

Note, I've spotted another bug a couple times, where underlining shows up on a random symbol in the text editor when you are ctrl+hovering in the infoview.  But it is hard to repro.  But we may have some code to write to intercept cntrl states in the text editor anyway to stop this from happening.

Vtec234:
> Yes I said with my patch it only "handles the case when infoview has the keyboard focus"

Duh, I missed that, sorry!

> This is also related to 'accessibility' we will need at some point to enable keyboard only navigation of the infoview, but that's a separate PR.

I see. This is important, but as you write it is a larger change. We will probably want things like `tabindex`. The keydown handlers don't help here though as the underline hint is visual.

> Note, I've spotted another bug a couple times, where underlining shows up on a random symbol in the text editor when you are ctrl+hovering in the infoview.

I imagine it could be some kind of lag, but I never saw this happen.

> An error message is printed in the tooltip for a brief moment before it closes.

Pushed a fix.

> Going to a previously unopened file prints getInteractiveDiagnostics error Error: already connecting or connected at 'file:///...'

I have a fix for this in another branch, will PR later.

lovettchris:
I filed this issue against vscode to see if they can find a way to finish our control key hooking so infoview is fully consistent with the next editor and works even when it doesn't have the focus, but I'm not holding my breath.

https://github.com/microsoft/vscode-languageserver-node/issues/959

lean4-infoview/src/infoview/interactiveCode.tsx
```diff
@@ -68,18 +69,93 @@ function TypePopupContents({pos, info, redrawTooltip}: {pos: DocumentPosition, i
...
-    <WithTooltipOnHover tooltipContent={mkTooltip}>
-      <HighlightOnHoverSpan>
+    <WithTooltipOnHover
+      mkTooltipContent={mkTooltip}
+      onClick={(e, next) => {
+        // On ctrl-click, if location is known, go to it in the editor
+        if (e.ctrlKey) void fetchGoToLoc().then(loc => {
+          if (loc === undefined) return
+          setHoverState(st => st === 'over' ? 'ctrlOver' : st)
+          void ec.revealPosition({ uri: loc.uri, ...loc.range.start })
```
lovettchris, line 132: `loc.range.start` is an interesting choice, there's some weird things you can do with a mouse drag operation so the mouse up occurs some place else.  Currently it seems to do the right thing, (namely vscode sends no onclick if you select more than one word) but you might be able to think of some ways to break it...

lean4-infoview/src/infoview/tooltips.tsx
```diff
@@ -70,11 +70,14 @@ export const Tooltip = forwardAndUseRef<HTMLDivElement,
   return ReactDOM.createPortal(popper, document.body)
 })
 
-/** An element which gets highlighted when its DOM children are hovered over. It is implemented
+export type HoverState = 'off' | 'over' | 'ctrlOver'
```
lovettchris, line 73: It all works, but I'm a little confused about how the code never has to set the HoverState to 'off' except in the intitalState of the React.useState...

lean4-infoview/src/infoview/tooltips.tsx
```diff
@@ -83,18 +86,19 @@ export const HighlightOnHoverSpan = forwardAndUseRef<HTMLSpanElement, React.HTML
     // tree and not just a logical React child (see useLogicalDom and
     // https://reactjs.org/docs/portals.html#event-bubbling-through-portals).
     if (ref.current && e.target instanceof Node && ref.current.contains(e.target)) {
-      if ('_HighlightOnHoverSpanSeen' in e) return
-      (e as any)._HighlightOnHoverSpanSeen = {}
-      setIsPointerOver(b)
+      if ('_DetectHoverSpanSeen' in e) return
+      (e as any)._DetectHoverSpanSeen = {}
+      if (!b) setHoverState('off')
```
lovettchris, line 91: Oh, never mind, it's because it's over here.  Hmm.  Its a bit messy that DetectHoverSpan is involved in the setHoverState book keeping along with InteractiveCode.tsx.  Why wouldn't WithTooltipOnHover do all the other setHoverState bookkeeping instead of doing it in InteractiveCode.tsx that way localizing it all to ToolTip.tsd ?  The only thing InteractiveCode really needs is the onClick to call ec.revealPosition, no?

vscode-lean4/src/utils/converters.ts
```diff
@@ -0,0 +1,51 @@
...
+import * as ls from 'vscode-languageserver-protocol'
+import * as code from 'vscode'
+import { Code2ProtocolConverter, Protocol2CodeConverter } from 'vscode-languageclient';
+
+interface Lean4Diagnostic extends ls.Diagnostic {
+    fullRange: ls.Range;
+}
+
+export const p2cConverter = createP2CConverter(undefined, true, true)
+export const c2pConverter = createC2PConverter(undefined)
```
lovettchris, line 24: I'm not sure about this, the default for `uriConverters` in `LanguageClientOptions` is:
```
uriConverters?: {
        code2Protocol: c2p.URIConverter;
        protocol2Code: p2c.URIConverter;
    };
```
which means this initialization code in BaseLanguageClient is passing `c2p.URIConverter` instead of `undefined`, no?
```
this._c2p = c2p.createConverter(clientOptions.uriConverters ? clientOptions.uriConverters.code2Protocol : undefined);
this._p2c = p2c.createConverter(
	clientOptions.uriConverters ? clientOptions.uriConverters.protocol2Code : undefined,
	this._clientOptions.markdown.isTrusted,
	this._clientOptions.markdown.supportHtml);
```

lovettchris (changes requested):
Seems to work well now, just a couple new code review comments...

vscode-lean4/src/utils/converters.ts
lovettchris: Ah, you are right, I do see undefined at runtime:
image

lean4-infoview/src/infoview/tooltips.tsx
Vtec234: So some level of interaction between `InteractiveCodeTag` and the tooltips.tsx components will always be required because the former holds the go-to implementation, and importantly we need to trigger a state update in the `InteractiveCodeTag` in order to re-render it. But you're right that the event handlers didn't have to be there, so I moved them to `DetectHoverSpan`.

Also, we could in principle do everything in `WithTooltipOnHover` and delete `DetectHoverSpan`, but it would become a monster component so I rather keep them separate but smaller.

lean4-infoview/src/infoview/interactiveCode.tsx
Vtec234: I went with just the start of the range since otherwise the whole range is selectec by our `showDocument` implementation. VSCode has some way to show a range by just pulsing rather than selecting it, but it doesn't seem exposed in the API.

lean4-infoview/src/infoview/tooltips.tsx
lovettchris: thanks, this is better, but I think you broke ctrl key in infoview in the process, because it when Tooltip calls `setHoverState(st => st === 'over' ? 'ctrlOver' : st)` it's not calling the real setHoverState it's calling this:
```
        setHoverState={st => {
          // On ctrl-hover, fetch the go-to location
          if (st === 'ctrlOver') void fetchGoToLoc()
          setHoverState(st)
        }}
```
and so the check ` if (st === 'ctrlOver')` desn't work when st is a lambda (st => st === 'over' ? 'ctrlOver' : st) and you can't use `hoverState` here, so perhaps you have to do this:

```
      // On ctrl-hover, fetch the go-to location
      // process the given st which could be a lambda
      setHoverState(st);
      // now use setHoverState again to query the resulting new state.
      setHoverState(st => {
          if (st === 'ctrlOver')
              void fetchGoToLoc();
          return st;
      });
```
but this doesn't fully fix it either...

Vtec234: I did indeed, thanks!

lovettchris: still having trouble, I've had your lean4 build installed and everything working before, I just updated your lean4 bits, and ctrl+click works fine in text editor but in the info view I get an exception saying `Lean.Widget.getGoToLocation` doesn't exist... which is odd because I do see it in your repo in `C:\msys64\home\clovett\git\vtec234\lean4\src\Lean\Server\FileWorker\WidgetRequests.lean`
perhaps I have a setup error... I'm using `elan toolchain link master C:\msys64\home\clovett\git\vtec234\lean4/build/release/stage1` which is what I did before when everything was working.... weird.

lovettchris: Never mind, I was in a multi-workspace and forgot to change one of my lean-toolchains to point to the right version.  It works now.  But I did get one print out early on while lean was still loading I guess:
```
Error in go-to-definition:  {"0":"C","1":"l","2":"i","3":"e","4":"n","5":"t","6":" ","7":"i","8":"s","9":" ","10":"n","11":"o","12":"t","13":" ","14":"r","15":"u","16":"n","17":"n","18":"i","19":"n","20":"g","length":21}
```
but it started working after this happened.

Issue #76, opened 2021-11-12
Resurrect tactic state filters


gebner:
Provide a way to filter out hypotheses in the tactic state / expected type.  For example:
 * Hide hidden assumptions (i.e., the ones with the dagger)
 * Hide type class instances.
 * Hide types (i.e., assumptions like `α : Type`).
 * Hide the gigantenormous let-values inside do-notation.

Some of that may require additional server support though.

gebner:
We've had a similar filter in the Lean 3 extension, which might be good for inspiration.

leodemoura:
@gebner I am missing this feature too. I am happy to implement the LSP server part. I think @lovettchris can help with the VSCode part. Do you have a design in mind? If yes, could you please share it here?

gebner:
I would have added a couple of flags in `InteractiveHypothesis` on the server side:
```lean
structure InteractiveHypothesis where
  names : Array String
  type : CodeWithInfos
  val? : Option CodeWithInfos := none
  -- new:
  isInstance : Bool
  isType : Bool
  deriving Inhabited, RpcEncoding
```

The filtering then happens here in the extension.  We can just add a dropdown like we had in Lean 3 (or a menu with checkboxes).

In (pre-widget) Lean 3 we also had filtering based on user-provided regexes.  I'm not sure how much people were using that feature, but it can be implemented purely in the extension.

> Hide the gigantenormous let-values inside do-notation.

For this I would first try to use let_fun instead of let in the desugaring.  (Or if that doesn't work because of VM or delaboration issues, a custom let_no_val elaborator that elaborates into a let, but adds a value-less free variable to the local context.)

Kha:
It feels wrong to me that we show `let`s from macros at all. On the other hand, for tactics we do often want to show hygiened local variables, especially when they can be un-hygiened by simply passing a name. Could we turn this around and make only tactic-introduced hygiened fvars visible (by default)? We already have special casing for `tactic.hygiene`, shouldn't these two things coincide?

leodemoura:
@gebner Thanks for the useful feedback.

> For this I would first try to use let_fun instead of let in the desugaring. (Or if that doesn't work because of VM or delaboration issues, a custom let_no_val elaborator that elaborates into a let, but adds a value-less free variable to the local context.)

[...]

PR #182, opened 2022-05-21
Add type and instance filters for hypotheses

lovettchris:
Part II of issue https://github.com/leanprover/vscode-lean4/issues/76  is adding filter buttons for hiding hypotheses that have either isType or isInstance set to true.  

https://user-images.githubusercontent.com/18707114/169914051-a802e4b9-5fa4-4fc6-9d7c-34ed6af91517.mp4

### DIFF
--- a/lean4-infoview/src/infoview/goalCompat.ts
+++ b/lean4-infoview/src/infoview/goalCompat.ts
@@ -31,6 +31,8 @@ function transformGoalToInteractive(g: string): InteractiveGoal {
         } else if (p.match(/^([^:\n< ][^:\n⊢{[(⦃]*) :/mg)) {
             const ss = p.split(':')
             const hyp: InteractiveHypothesis = {
+                isType: false,
+                isInstance: false,
                 names: ss[0].split(' ')
                     .map(s => s.trim())
                     .filter(s => s.length !== 0),
--- a/lean4-infoview/src/infoview/goals.tsx
+++ b/lean4-infoview/src/infoview/goals.tsx
@@ -30,9 +30,13 @@ export function goalsToString(goals: InteractiveGoals): string {
 }
 
 export interface GoalFilterState {
+    /** If true reverse the list of InteractiveHypothesis, if false present the order received from LSP */
     reverse: boolean,
+    /** If true show InteractiveHypothesis that have isType=True, if false, hide InteractiveHypothesis that have isType=True. */
     isType: boolean,
+    /** If true show InteractiveHypothesis that have isInstance=True, if false, hide InteractiveHypothesis that have isInstance=True. */
     isInstance: boolean,
+    /** If true show InteractiveHypothesis that contain a dagger in the name, if false, hide InteractiveHypothesis that contain a dagger in the name. */
     isHiddenAssumption: boolean
 }
 
--- a/lean4-infoview/src/infoview/goals.tsx
+++ b/lean4-infoview/src/infoview/goals.tsx
@@ -32,13 +32,19 @@ export function goalsToString(goals: InteractiveGoals): string {
 export interface GoalFilterState {
     reverse: boolean,
     isType: boolean,
-    isInstance: boolean
+    isInstance: boolean,
+    isHiddenAssumption: boolean
+}
+
+function isHiddenAssumption(h: InteractiveHypothesis){
+    return h.names.every(n => n.indexOf('✝') >= 0);
 }
 
 function getFilteredHypotheses(hyps: InteractiveHypothesis[], filter: GoalFilterState): InteractiveHypothesis[] {
     return hyps.filter(h =>
         (!h.isInstance || filter.isInstance) &&
-        (!h.isType || filter.isType));
+        (!h.isType || filter.isType) &&
+        (filter.isHiddenAssumption || !isHiddenAssumption(h)));
 }
 
 export function Goal({pos, goal, filter}: {pos: DocumentPosition, goal: InteractiveGoal, filter: GoalFilterState}) {
--- a/lean4-infoview/src/infoview/goals.tsx
+++ b/lean4-infoview/src/infoview/goals.tsx
@@ -36,22 +36,9 @@ export interface GoalFilterState {
 }
 
 function getFilteredHypotheses(hyps: InteractiveHypothesis[], filter: GoalFilterState): InteractiveHypothesis[] {
-    const result : InteractiveHypothesis[] = [];
-    for (const h of hyps) {
-        if (h.isInstance) {
-            if (filter.isInstance)
-                result.push(h);
-        }
-        else if (h.isType) {
-            if (filter.isType)
-                result.push(h);
-        }
-        else{
-            // TODO: are there other types?
-            result.push(h);
-        }
-    }
-    return result;
+    return hyps.filter(h =>
+        (!h.isInstance || filter.isInstance) &&
+        (!h.isType || filter.isType));
 }
 
 export function Goal({pos, goal, filter}: {pos: DocumentPosition, goal: InteractiveGoal, filter: GoalFilterState}) {
--- a/lean4-infoview/src/infoview/goals.tsx
+++ b/lean4-infoview/src/infoview/goals.tsx
@@ -2,7 +2,7 @@ import * as React from 'react'
 import { DocumentPosition } from './util'
 import { ConfigContext } from './contexts'
 import { InteractiveCode } from './interactiveCode'
-import { InteractiveGoal, InteractiveGoals, TaggedText, TaggedText_stripTags } from './rpcInterface'
+import { InteractiveGoal, InteractiveGoals, InteractiveHypothesis, TaggedText_stripTags } from './rpcInterface'
 
 function goalToString(g: InteractiveGoal): string {
     let ret = ''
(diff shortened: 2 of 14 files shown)


lean4-infoview/src/infoview/rpcInterface.ts
```diff
@@ -65,6 +65,8 @@ export async function InteractiveDiagnostics_infoToInteractive(rs: RpcSessions,
 }
 
 export interface InteractiveHypothesis {
+    isInstance: boolean,
+    isType: boolean,
```
gebner, line 69:
```suggestion
    isInstance?: boolean,
    isType?: boolean,
```
Since this property is not present in older Lean versions.

lean4-infoview/src/infoview/goals.tsx
```diff
@@ -29,38 +29,60 @@ export function goalsToString(goals: InteractiveGoals): string {
...
+        else if (h.isType) {
+            if (filter.isType)
+                result.push(h);
+        }
+        else{
+            // TODO: are there other types?
+            result.push(h);
+        }
+    }
+    return result;
```
gebner, line 54:
```suggestion
    return hyps.filter(h =>
        (!h.isInstance || filter.isInstance) &&
        (!h.isType || filter.isType));
```

lean4-infoview/src/infoview/info.tsx
```diff
@@ -123,8 +124,37 @@ export function InfoDisplay(props0: InfoDisplayProps) {
     const hasGoals = status !== 'error' && goals;
     const hasTermGoal = status !== 'error' && termGoal;
     const hasMessages = status !== 'error' && messages.length !== 0;
-    const filterClasses = 'link pointer mh2 dim codicon fr ' + (reverseOrder ? 'codicon-arrow-up' : 'codicon-arrow-down');
-    const sortButton = <a className={filterClasses} onClick={e => { setReverseOrder(!reverseOrder); }} title="reverse list"/>
+    const sortClasses = 'link pointer mh2 dim codicon fr ' + (goalFilters.reverse ? 'codicon-arrow-up' : 'codicon-arrow-down');
+    const sortButton = <a className={sortClasses} title="reverse list" onClick={e => {
+        setGoalFilters(s => {
+            return { reverse: !s.reverse, isType: s.isType, isInstance: s.isInstance }
+        } ); }
```
gebner, line 131:
```suggestion
        setGoalFilters(s => ({ ...s, reverse: !s.reverse })); }
```

lean4-infoview/src/infoview/info.tsx
```diff
@@ -123,8 +124,37 @@ export function InfoDisplay(props0: InfoDisplayProps) {
...
+    } />
+
+    const filterMenu = <span>
+        <a className='link pointer' onClick={e => {
+            setGoalFilters(s => {
+                return { reverse: s.reverse, isType: !s.isType, isInstance: s.isInstance }
+            } ); }}>
+                <span className={'filterMenuIcon codicon ' + (goalFilters.isType ? 'codicon-eye' : 'codicon-eye-closed')}>&nbsp;</span>
```
gebner, line 139: The "eye" codicon looks very cryptic to me.  Maybe it would be more intuitive if this was an actual checkbox, and the text was "show instances" (or "hide instances").

lean4-infoview/src/infoview/goalCompat.ts
```diff
@@ -31,6 +31,8 @@ function transformGoalToInteractive(g: string): InteractiveGoal {
         } else if (p.match(/^([^:\n< ][^:\n⊢{[(⦃]*) :/mg)) {
             const ss = p.split(':')
             const hyp: InteractiveHypothesis = {
+                isType: false,
+                isInstance: false,
```
gebner, line 35:
```suggestion
                isType: true,
                isInstance: true,
```

We shouldn't hide anything by default (especially when there's no button to change the filtering).

lean4-infoview/src/infoview/rpcInterface.ts
lovettchris: thanks

lean4-infoview/src/infoview/goalCompat.ts
gebner: Oh then it's ok.  Got the flags backwards.

lean4-infoview/src/infoview/info.tsx
lovettchris: Ok, I found another filter menu in the Problems UI that does use checkboxes, so consistency wins:
image

Even though this is exactly what the eye icon was invented for :-) It literally means "show it so I can see it" or "hide it so I cannot see it", but I guess not enough apps have used it over the years which results in your reaction...

Problem is now I have to find a "blank filler" that is exactly the same size as the checkmark icon but codeicons has no such blank icon defined...

lean4-infoview/src/infoview/goalCompat.ts
gebner: I actually confused the filter flags with the hypothesis flags. :bulb:

lean4-infoview/src/infoview/info.tsx
lovettchris: I can't use a fixed pixel sized filler because it changes with vs code font size (Ctrl + and Ctrl -)

gebner: I understand the eye icon in general, it's just that the codicons one doesn't look at all like an eye at all.  The checkmark is certainly clearer (at least to me).

> Problem is now I have to find a "blank filler" that is exactly the same size as the checkmark icon but codeicons has no such blank icon defined...

The two eye icons also had different sizes, it's not the end of the world if they don't match exactly.

lean4-infoview/src/infoview/info.tsx
lovettchris: Thanks, I fixed all 3 places this was needed.

lean4-infoview/src/infoview/info.tsx
lovettchris: Turns out fixed pixel icon does work, the Ctrl+ key must be scaling the DPI.  I updated the demo video to show:

image

lean4-infoview/src/infoview/goalCompat.ts
lovettchris: Ah but I did find another place in traceExplorer.tsx that needed a related fix.

lovettchris:
While I'm at it, I added hidden assumption filter based on presence of the dagger in the hyp name since you asked for that in issue# 76 as follows:
image

which filters down to:
image

lean4-infoview/src/infoview/info.tsx
lovettchris: Note: I did find one eye icon usage in vscode for lazily expanded variables in the debugger:
image

gebner:
Works great!  Two minor nitpicks:
 1. The contrast between the links and the tooltip background is pretty bad with light themes.
 2. If only hidden assumptions is checked, then the filter icon is not "filled".

filters_lightmode

lovettchris:
> Works great! Two minor nitpicks:
> 1. The contrast between the links and the tooltip background is pretty bad with light themes.
> 2. If only hidden assumptions is checked, then the filter icon is not "filled".

Good catch thanks, we also have a contrast problem in regular tooltips with the white on light blue shown here:
image

I pushed a fix so the light theme now looks like this:
image

And the dark theme like this:
image

I also fixed the filter icon state bug you found.

gebner:
I like the white background in the tooltips, but the text should be black (for contrast; the same color as the text outside the tooltip).

@Vtec234 When you added the tooltips originally, you made them black even in light themes (but not white in dark themes).  (There's even a comment: "tooltips are _always_ dark.")  Did you have a reason for that?

lovettchris:
> I like the white background in the tooltips, but the text should be black (for contrast; the same color as the text outside the tooltip).

As you can see in index.css I'm using the following official VS code themed colors for editor widgets, so I'm inclined to stick with that:
```
    color: var(--vscode-editorWidget-foreground);
    background: var(--vscode-editorWidget-background);
```

Vtec234:
>  When you added the tooltips originally, you made them black even in light themes (but not white in dark themes). (There's even a comment: "tooltips are always dark.") Did you have a reason for that?

There.. is? :) You're right, and I don't remember why. Chris's changes look much better, let's go with those. I think for the dark-on-dark tooltips, though, we should change the shadow colour to a light one, if possible. In light themes the dark shadow makes it easy to see the tooltip boundary whereas in dark themes a dark shadow doesn't.

> As you can see in index.css I'm using the following official VS code themed colors for editor widgets, so I'm inclined to stick with that:

Note though that in your screenshot, in the light theme, text inside the tooltip is lighter than other infoview text (unless that's an illusion?).

lean4-infoview/src/infoview/goals.tsx
```diff
@@ -29,38 +29,53 @@ export function goalsToString(goals: InteractiveGoals): string {
     return goals.goals.map(goalToString).join('\n\n')
 }
 
-export function Goal({pos, goal, reverse}: {pos: DocumentPosition, goal: InteractiveGoal, reverse: boolean}) {
+export interface GoalFilterState {
```
Vtec234, line 32: Let's put a comment on the interface as to which way the `isFoo` flags go. Does `true` mean show or hide?

Vtec234 (approved):
Looks great!

lean4-infoview/src/infoview/goals.tsx
lovettchris: Done

Vtec234:
As another visual suggestion, I would remove the `HighlightOnHoverSpan`. It doesn't align with the icon screenshot, and I think the `dim` effect is enough of an indication for 'here is something you can interact with'.

lovettchris:
> we should change the shadow colour to a light one

Turns out that VS code does define `--vscode-widget-shadow` and they've decided to make it black in the dark theme for some reason.  I switched to that for now.

> I like the white background in the tooltips, but the text should be black (for contrast; the same color as the text outside the tooltip).

I switched from `color: var(--vscode-editorWidget-foreground);` to `@extend .mtk1; which is the color used by the text editor, so the text is now black on light theme :
image
and white on dark theme:
image

lovettchris:
> As another visual suggestion, I would remove the `HighlightOnHoverSpan`. It doesn't align with the icon screenshot, and I think the `dim` effect is enough of an indication for 'here is something you can interact with'.

Done.

lean4-infoview/src/infoview/index.css
```diff
@@ -118,3 +121,17 @@ select {
 .tooltip[data-popper-placement^='right'] > .tooltip-arrow {
     left: -4px;
 }
+
+.popup-menu-text { text-align: center; padding: 5px; vertical-align:top; display:inline-block; }
```
Vtec234, line 125: I missed this before, but so far we have mostly been using [Tachyons](http://tachyons.io/docs/table-of-styles/) rather than bespoke CSS classes for styling of this kind (with the exception of this exact file, of course). So instead of `popup-menu-text` one would write `tc pa2 v-top dib` as the `className` (`pa1` or `pa2`, Tachyons intentionally doesn't provide specific pixel offsets for padding so as to keep a consistent, fixed scale). All the other classes either use `var` or *need* specific offsets to match a font, so it only applies to `popup-menu-text`. I will leave to you whether you want to change this, it's not the end of the world if we use classes.

lovettchris: Interesting, I have lots of questions, so I moved this to a new issue  https://github.com/leanprover/vscode-lean4/issues/183

lovettchris:
I went through all the comments again, I think this is done.  Did I miss anything?

Vtec234:
Merged with history since this is a large change, but please do follow the lean4-like [commit convention](https://github.com/leanprover/lean4/blob/master/doc/dev/commit_convention.md).

lovettchris:
> Merged with history since this is a large change, but please do follow the lean4-like [commit convention](https://github.com/leanprover/lean4/blob/master/doc/dev/commit_convention.md).

woops, yeah, I assumed it would be squash merged...

PR #193, opened 2022-06-14
refactor: preparing for user-widgets and context suggestions

EdAyers:
This is a set of refactors that me and @vtec234 made while working on user widgets and context suggestions.

- some docstrings
- refactor InteractiveHypothesis => InteractiveHypothesisBundle
- refactor: extract HypothesisBundle render to own component.
- introduce a `useAsync` hook for use with rpcs
- refactor TypePopupContents
- InteractiveGoal and InteractiveHypothesisBundle have mvarId? and fvarIds? fields
- helper function mapRpcError
- helper enum RpcErrorCode
- new override of useEventResult
- fix a bug in handleInsertText where it was doing reference eq instead of string eq.

Co-authored-by: @Vtec234

### DIFF
--- a/lean4-infoview/src/components.tsx
+++ b/lean4-infoview/src/components.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 
-import { InteractiveDiagnostics_msgToInteractive, TaggedText, MsgEmbed, CodeWithInfos, MessageData, mapRpcError } from './infoview/rpcInterface';
+import { InteractiveDiagnostics_msgToInteractive, MessageData, mapRpcError } from './infoview/rpcInterface';
 import { InteractiveMessage } from './infoview/traceExplorer';
 import { DocumentPosition, useAsync } from './infoview/util';
 import { RpcContext } from './infoview/contexts';
@@ -25,7 +25,7 @@ export function InteractiveMessageData({ pos, msg }: { pos: DocumentPosition, ms
     } else if (status === 'pending') {
         return <>...</>
     } else {
-        return <div>failed to load messages
+        return <div>Failed to display message:
             {error && <span>{mapRpcError(error).message}</span>}
         </div>
     }
--- a/lean4-infoview/src/components.tsx
+++ b/lean4-infoview/src/components.tsx
@@ -1,0 +1,27 @@
+import * as React from 'react';
+
+import { InteractiveCode } from './infoview/interactiveCode';
+import { InteractiveDiagnostics_msgToInteractive, TaggedText, MsgEmbed, ExprWithCtx, CodeWithInfos, MessageData } from './infoview/rpcInterface';
+import { InteractiveMessage } from './infoview/traceExplorer';
+import { DocumentPosition } from './infoview/util';
+import { RpcContext } from './infoview/contexts';
+
+export { DocumentPosition };
+export { EditorContext, RpcContext, VersionContext } from './infoview/contexts';
+export { EditorConnection } from './infoview/editorConnection';
+export { RpcSessions } from './infoview/rpcSessions';
+export { ServerVersion } from './infoview/serverVersion';
+
+/** Display the given message data as interactive, pretty-printed text. */
+export function InteractiveMessageData({pos, msg}: {pos: DocumentPosition, msg: MessageData}) {
+    const rs = React.useContext(RpcContext)
+    const [tt, setTt] = React.useState<TaggedText<MsgEmbed> | undefined>(undefined)
+
+    React.useEffect(() => {
+        void InteractiveDiagnostics_msgToInteractive(rs, pos, msg, 0)
+            .then(tt => tt && setTt(tt))
+    }, [pos.character, pos.line, pos.uri, msg])
+
+    if (tt) return <InteractiveMessage pos={pos} fmt={tt} />
+    else return <></>
+}
--- a/lean4-infoview/src/components.tsx
+++ b/lean4-infoview/src/components.tsx
@@ -1,8 +1,8 @@
 import * as React from 'react';
 
-import { InteractiveDiagnostics_msgToInteractive, MessageData, mapRpcError } from './infoview/rpcInterface';
+import { InteractiveDiagnostics_msgToInteractive, MessageData } from './infoview/rpcInterface';
 import { InteractiveMessage } from './infoview/traceExplorer';
-import { DocumentPosition, useAsync } from './infoview/util';
+import { DocumentPosition, useAsync, mapRpcError } from './infoview/util';
 import { RpcContext } from './infoview/contexts';
 
 export { DocumentPosition };
--- a/lean4-infoview/src/components.tsx
+++ b/lean4-infoview/src/components.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 
-import { InteractiveCode } from './infoview/interactiveCode';
 import { InteractiveDiagnostics_msgToInteractive, TaggedText, MsgEmbed, CodeWithInfos, MessageData, mapRpcError } from './infoview/rpcInterface';
 import { InteractiveMessage } from './infoview/traceExplorer';
 import { DocumentPosition, useAsync } from './infoview/util';
@@ -17,7 +16,7 @@ export function InteractiveMessageData({ pos, msg }: { pos: DocumentPosition, ms
     const rs = React.useContext(RpcContext)
 
     const [status, tt, error] = useAsync(
-        () => InteractiveDiagnostics_msgToInteractive(rs, pos, { msg, indent: 0 }),
+        () => InteractiveDiagnostics_msgToInteractive(rs, pos, msg, 0),
         [pos.character, pos.line, pos.uri, msg]
     )
 
--- a/lean4-infoview/src/components.tsx
+++ b/lean4-infoview/src/components.tsx
@@ -1,9 +1,9 @@
 import * as React from 'react';
 
 import { InteractiveCode } from './infoview/interactiveCode';
-import { InteractiveDiagnostics_msgToInteractive, TaggedText, MsgEmbed, ExprWithCtx, CodeWithInfos, MessageData } from './infoview/rpcInterface';
+import { InteractiveDiagnostics_msgToInteractive, TaggedText, MsgEmbed, CodeWithInfos, MessageData, mapRpcError } from './infoview/rpcInterface';
 import { InteractiveMessage } from './infoview/traceExplorer';
-import { DocumentPosition } from './infoview/util';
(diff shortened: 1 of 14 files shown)


lean4-infoview/src/infoview/goals.tsx
```diff
@@ -1,8 +1,28 @@
...
+    pos: DocumentPosition
+    hyp: InteractiveHypothesisBundle
+    index: number
+}
+
+export function Hyp({ pos, hyp : h, index }: HypProps) {
+    const names = InteractiveHypothesisBundle_accessableNames(h).map((n, i) =>
+            <span className="mr1">{n}</span>
+        )
+    const hypKey = (h.fvarIds?.[0] ?? index)
```
gebner, line 16:
```suggestion
    const hypKey = index
```

I'm not sure if it's a good idea to include "random" information in a key.  But this is should not be *too* bad, since the hypothesis fvarids remain somewhat stable if you move along a proof.

lean4-infoview/src/infoview/goals.tsx
```diff
@@ -30,56 +50,67 @@ export function goalsToString(goals: InteractiveGoals): string {
...
+    filter: GoalFilterState
+}
+
+export function Goals({ pos, goals, filter }: GoalsProps) {
     if (goals.goals.length === 0) {
         return <>Goals accomplished 🎉</>
     } else {
         return <>
-            {goals.goals.map ((g, i) => <Goal key={i} pos={pos} goal={g} filter={filter}/>)}
+            {goals.goals.map((g, i) => <Goal key={g.mvarId || i} pos={pos} goal={g} filter={filter} index={i} />)}
```
gebner, line 113:
```suggestion
            {goals.goals.map((g, i) => <Goal key={i} pos={pos} goal={g} filter={filter} index={i} />)}
```
This is almost certainly a bad idea since the mvarid changes on every line of the proof.

lean4-infoview/src/infoview/interactiveCode.tsx
```diff
@@ -32,48 +32,41 @@ export function InteractiveTaggedText<T>({pos, fmt, InnerTagUi}: InteractiveTagg
...
+  return <>
+    {ip && <>
       {ip.exprExplicit && <InteractiveCode pos={pos} fmt={ip.exprExplicit} />} : {ip.type && <InteractiveCode pos={pos} fmt={ip.type} />}
       {ip.doc && <hr />}
       {ip.doc && ip.doc} {/* TODO markdown */}
-    </>
-  } else return <>Loading..</>
+    </>}
+    {err && <>Error: {mapRpcError(err).message}</>}
+    {(!ip && !err) && <>Loading..</>}
```
gebner, line 61: Not sure if this is an improvement over if/then/else, but I'm not against it.

lean4-infoview/src/infoview/rpcSessions.tsx
```diff
@@ -96,13 +96,14 @@ export class RpcSessions implements Disposable {
 
     private async sessionAt(uri: DocumentUri): Promise<RpcSession | undefined> {
         if (this.#connected.has(uri)) return this.#connected.get(uri)
-        else if (this.#connecting.has(uri)) return this.#connecting.get(uri)
+        else if (this.#connecting.has(uri)) return await this.#connecting.get(uri)
```
gebner, line 99: The `await` shouldn't be required here, or am I missing something?

lean4-infoview/src/infoview/util.ts
```diff
@@ -261,3 +263,61 @@ export function useLogicalDom(ref: React.RefObject<HTMLElement>): [LogicalDomTra
     React.useMemo(() => ({registerDescendant}), [parentCtx])
   ]
 }
+
+type Status = 'pending' | 'fulfilled' | 'rejected'
+
+/** This hook will run the given promise function whenever the deps change.
+ * There will only ever be one in-flight promise at a time. If the deps
+ * change while a promise is still in-flight, then the function will be run
+ * again after the first promise has resolved.
```
gebner, line 272: The purpose of this function is to prevent race conditions where the RPC requests return in a different order than they were sent, right?
1. Request 1 is sent with, say, line=42.
2. Request 2 is sent with line=90.
3. Request 2 returns with diags=[].
4. Request 1 returns with diags=['error'].
And without `useAsync` we would now return the diagnostics for line 42 even though we're at line 90.

If that's the case, then that should be written in the docstring.

Is there a reason for "only ever be one in-flight promise at a time"?

vscode-lean4/src/infoview.ts
```diff
@@ -606,7 +606,7 @@ export class InfoProvider implements Disposable {
     private async handleInsertText(text: string, kind: TextInsertKind, uri?: Uri, pos?: Position) {
         let editor: TextEditor | undefined
         if (uri) {
-           editor = window.visibleTextEditors.find(e => e.document.uri === uri);
+           editor = window.visibleTextEditors.find(e => e.document.uri.path === uri.path);
```
gebner, line 609:
```suggestion
           editor = window.visibleTextEditors.find(e => e.document.uri.toString() === uri.toString());
```

Before somebody complains that we're inserting text in `http://example.com/home/gebner/foo.lean` instead of `/home/gebner/foo.lean`. :smile:

lean4-infoview/src/infoview/rpcInterface.ts
```diff
@@ -66,19 +67,35 @@ export async function InteractiveDiagnostics_infoToInteractive(rs: RpcSessions,
...
+     */
     names: string[]
+    /** The free variable id associated with each of the vars listed in `names`. */
+    fvarIds?: string[]
     type: CodeWithInfos
     val?: CodeWithInfos
 }
 
+/** Filter out inaccessible / anonymous pretty names from the names list. */
+export function InteractiveHypothesisBundle_accessableNames(ih : InteractiveHypothesisBundle) : string[] {
```
gebner, line 86:
```suggestion
export function InteractiveHypothesisBundle_accessibleNames(ih : InteractiveHypothesisBundle) : string[] {
```
Typo?

lean4-infoview/src/infoview/rpcSessions.tsx
EdAyers: Hmm I'm not sure. If you do `return p` where `p : Promise<T>` in an async function  does the compiler automatically squash the `Promise<Promise<T>>` to a `Promise<T>`? I guess it must do.

lean4-infoview/src/infoview/util.ts
EdAyers: The reason for only having one at a time is so that you don't spam server with requests but maybe the server can handle it.

EdAyers: Instead of the 'one at a time' model we could instead have the callback of the promise check whether it has been superceded by a later rpc call and discard its result. That would also work.

lean4-infoview/src/infoview/errors.tsx
```diff
@@ -0,0 +1,26 @@
...
+  constructor(props: {}) {
+    super(props);
+    this.state = { error: undefined };
+  }
+
+  static getDerivedStateFromError(error: any) {
+    // Update state so the next render will show the fallback UI.
+    return { error: error.toString() };
+  }
+
```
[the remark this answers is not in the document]
EdAyers: componentDidCatch is needed if you want to log errors. I added a stub that does nothing in case react is looking for that method.

lean4-infoview/src/infoview/goals.tsx
EdAyers: removed this for now as key only matters on parent `Hyp` component.

lean4-infoview/src/components.tsx
```diff
@@ -0,0 +1,27 @@
...
+export { EditorConnection } from './infoview/editorConnection';
+export { RpcSessions } from './infoview/rpcSessions';
+export { ServerVersion } from './infoview/serverVersion';
+
+/** Display the given message data as interactive, pretty-printed text. */
+export function InteractiveMessageData({pos, msg}: {pos: DocumentPosition, msg: MessageData}) {
+    const rs = React.useContext(RpcContext)
+    const [tt, setTt] = React.useState<TaggedText<MsgEmbed> | undefined>(undefined)
+
+    React.useEffect(() => {
```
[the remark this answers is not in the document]
EdAyers: redid this component for `useAsync`

lean4-infoview/src/infoview/goals.tsx
```diff
@@ -30,56 +49,67 @@ export function goalsToString(goals: InteractiveGoals): string {
...
+
+export function Goal({ pos, goal, filter, index }: GoalProps) {
     const prefix = goal.goalPrefix ?? '⊢ '
     const filteredList = getFilteredHypotheses(goal.hyps, filter);
-    const hyps = filter.reverse  ? filteredList.slice().reverse() : filteredList;
-    const goalLi  = <li key={'goal'}>
-                        <strong className="goal-vdash">{prefix}</strong><InteractiveCode pos={pos} fmt={goal.type} />
-                     </li>
+    const hyps = filter.reverse ? filteredList.slice().reverse() : filteredList;
+    const goalId = goal.mvarId || index
```
Vtec234, line 86: Seems unused? (for now)

lean4-infoview/src/infoview/rpcSessions.tsx
Vtec234: It's not required, I just put it in for clarity.

lean4-infoview/src/infoview/util.ts
Vtec234: It might be worth it to check if the more eager (multiple requests) approach increases responsiveness or not.

lean4-infoview/src/infoview/info.tsx
```diff
@@ -186,7 +186,7 @@ export function InfoDisplay(props0: InfoDisplayProps) {
                         Expected type {sortButton} {filterButton}
                     </summary>
                     <div className='ml1'>
-                        {hasTermGoal && <GoalUi pos={pos} goal={termGoal} filter={goalFilters} />}
+                        {hasTermGoal && <GoalUi pos={pos} goal={termGoal} filter={goalFilters} index={0} />}
```
Vtec234, line 189: @EdAyers not sure if this is the right fix. Maybe it should be an optional prop?

lean4-infoview/src/infoview/util.ts
EdAyers: I added an unthrottled version, not sure what the best way to test is. They both seem to work.

gebner: Let's just stick with the unthrottled one for now.

PR #194, opened 2022-06-14
feat: user widgets

EdAyers:
Introduces arbitrary javascript widgets to Lean 4.

## Todo

- [x] documentation
- [x]  https://github.com/leanprover/lean4/pull/1238
- [x] merge #193


lean4-infoview/src/infoview/userWidget.tsx
```diff
@@ -0,0 +1,125 @@
...
+            throw Error(`RPC Error: ${RpcErrorCode[e.code]}: ${e.message}`)
+        }
+    } else if (e instanceof Error) {
+        throw e
+    } else {
+        throw Error(`Unknown rpc error ${JSON.stringify(e)}`)
+    }
+}
+
+export function Widget_getWidget(rs: RpcSessions, pos: DocumentPosition): Promise<GetWidgetResponse | undefined> {
```
gebner, line 32: I think this should go to `rpcInterface.ts`.

lean4-infoview/src/infoview/userWidget.tsx
```diff
@@ -0,0 +1,125 @@
...
+    return rs.call<GetWidgetResponse | undefined>(pos, 'Lean.Widget.getWidget', DocumentPosition.toTdpp(pos))
+        .catch<undefined>(handleWidgetRpcError);
```
gebner, line 34:
```suggestion
```
I'd rather keep the logic uniform among all the RPC wrapper functions.  The wrappers usually just assert the right type and register the RPC references.

lean4-infoview/src/infoview/userWidget.tsx
```diff
@@ -0,0 +1,125 @@
...
+            }
+            return result
+        }
+        return cache.get(key)
+    }
+    return r
+}
+
+const dynamicallyLoadComponent = memoize(function (hash: number, code: string,) {
+    return React.lazy(async () => {
```
gebner, line 74: Can `React.lazy` be memoized safely?  (For example, hooks would definitely break.)

lean4-infoview/src/infoview/userWidget.tsx
```diff
@@ -0,0 +1,125 @@
...
+    javascript: string
+    hash: number
+}
+
+/** Gets the static JS code for a given widget.
+ *
+ * We make the assumption that either the code doesn't exist, or it exists and does not change for the lifetime of the widget.
+ * [todo] cache on widgetId, but then there needs to be some way of signalling that the widgetId's code has changed if the user edits it?
+ */
+export async function Widget_getStaticJS(rs: RpcSessions, pos: DocumentPosition, widgetId: string): Promise<StaticJS | undefined> {
```
gebner, line 47:
```suggestion
export async function Widget_getStaticJS(rs: RpcSessions, pos: DocumentPosition, hash: number): Promise<StaticJS | undefined> {
```
Just treat the server as a content-addressed javascript CDN.  Then it doesn't matter if the code changes while the extension is running.

lean4-infoview/src/infoview/userWidget.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+import { EditorContext, RpcContext } from './contexts';
+import { Details } from './collapsing';
+import { mapRpcError, } from './rpcInterface';
+import { DocumentPosition, useAsync, useEventResult } from './util';
+import { ErrorBoundary } from './errors';
+import { RpcSessions } from './rpcSessions';
+import { isRpcError, RpcErrorCode } from '@lean4/infoview-api';
+
+export interface GetWidgetResponse {
```
gebner, line 12: Some documentation would really be helpful here...

lean4-infoview/src/infoview/userWidget.tsx
```diff
@@ -0,0 +1,125 @@
...
+    id: string
+    hash: number
```
gebner, line 14: If `hash` is a `UInt64`, then you'll need to rethink the type here.  Because the default javascript JSON deserializer will perform a lossy conversion to a floating-point number.

lean4-infoview/src/infoview/userWidget.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+interface GetWidgetResult {
+    component: any
+    id: string
+    hash: number
+    props: any
+}
+
+const getCode = memoize(
+    (rc: RpcSessions, pos: DocumentPosition, widget: GetWidgetResponse) => Widget_getStaticJS(rc, pos, widget.id),
```
gebner, line 89: This is almost certainly a really bad idea, because it will cache the `Promise` and won't retry if it fails.

lean4-infoview/src/infoview/userWidget.tsx
```diff
@@ -0,0 +1,125 @@
...
+            const component = dynamicallyLoadComponent(widget.hash, code.javascript)
+            return component
+        },
+        [pos.uri, pos.line, pos.character, widget.hash])
+
+    const widgetId = widget.id
+    const componentProps = { pos, ...widget.props }
+
+    return (
+        <React.Suspense fallback={`Loading widget: ${widgetId} ${status}.`}>
```
gebner, line 118: I just learned about `React.Suspense`!  I wonder if could replace `useAsync` with `React.Suspense` entirely.

lean4-infoview/src/infoview/info.tsx
```diff
@@ -200,6 +208,18 @@ export function InfoDisplay(props0: InfoDisplayProps) {
...
             </div>
+            <div style={{display: hasWidget ? 'block' : 'none'}}>
+                <Details initiallyOpen>
+                    <summary className="mv2 pointer">
+                    Widget: {widget && widget.id}
+                      {widgetStatus === 'pending' && ' (pending)'}
+                      {widgetStatus === 'rejected' && ' (errored)'}
+                    </summary>
+                    <div className="ml1">
+                      {widget && <UserWidget pos={pos} widget={widget}/>}
```
gebner, line 219: Ordering-wise, I think it would make more sense to put the widgets above the messages.

I'm not sure how I feel about associating widgets to source-code positions.  Does this show all widgets on the line, only the one at the exact position, or do they have a range server-side?  Most of the examples I have seen for widgets associate one widget per goal.  (But then there's also `#eval`, which kind of associates a widget to a diagnostic.)

lean4-infoview/src/infoview/userWidget.tsx
EdAyers: I think it's ok, the official usage is like

```
const MyComponent = React.lazy( () => ...)

function MyOtherComponent() {
  return <MyComponent foo="asdf"/>
}
```

so memoising is just going to store a dictionary of the `MyComponent`s. Nothing with hooks happens until `MyComponent` is rendered.

lean4-infoview/src/infoview/userWidget.tsx
EdAyers: I think fixed now

lean4-infoview/src/infoview/userWidget.tsx
EdAyers: working on it!

lean4-infoview/src/infoview/info.tsx
EdAyers: Sorry I should have pushed the RFC that I am writing before this.

lean4-infoview/src/infoview/userWidget.tsx
EdAyers: From reading [the react docs](https://reactjs.org/docs/react-api.html#reactsuspense) it sounds like using `React.Suspense` for arbitrary data loading is on the roadmap but not supported yet. But yes one day.

lean4-infoview/src/infoview/info.tsx
EdAyers: The idea is that widgets exist in the infotree same as any other info, and given a pos in the infoview we just render all of the widgets there, one after the other. But for most cases we would expect there to just be at most one widget.

lean4-infoview/src/infoview/info.tsx
```diff
@@ -121,10 +122,29 @@ export function InfoDisplay(props0: InfoDisplayProps) {
     });
 
     const rs = React.useContext(RpcContext);
+
+    // # widgets
+    const [widgetTrig, setWidgetTrig] = React.useState(0)
+    useClientNotificationEffect('textDocument/didChange', () => setWidgetTrig(widgetTrig + 1), [widgetTrig])
```
EdAyers, line 128: This feels nonstandard

gebner: Indeed.  More importantly, the widget should be part of the "pausable" "display props" which are set in `InfoAux`.  The widget request should be analogous to the goal request.  And certainly not separately triggered on didChange events.

EdAyers: This is fixed now

lean4-infoview/src/infoview/rpcInterface.ts
```diff
@@ -183,3 +184,45 @@ export async function getGoToLocation(rs: RpcSessions, pos: DocumentPosition, ki
...
+/** Code that should be dynamically loaded by the UserWidget component. */
+export interface WidgetSource {
+    /** JavaScript sourcecode. Should be a plain JavaScript ESModule whose default export is
+     * the component to render.
+     */
+    sourcetext: string
+    hash: string
+}
+
+/** Gets the static code for a given widget.
```
gebner, line 222:
```suggestion
/** Gets the static javascript code for a given widget.
```

@Vtec234 At some point you said that you called this function `getWidgetSourceJS` to clarify that it returns javascript, and so that we could potentially have `getWidgetSourceNeovim` in the future.  Did you have a change of heart?

EdAyers: Originally it was called getWidgetStaticJS, but one conclusion of the RFC was that it should also work with other languages in the future, so the mention of JS should be removed from API. There is no check on the Lean side as to what language the source is. I guess we can add that, perhaps a language tag on source items?

lean4-infoview/src/infoview/userWidget.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export function UserWidget({ pos, widget }: UserWidgetProps) {
+    const rs = React.useContext(RpcContext);
+    if (!pos) {
+        return <>Waiting for a location.</>
+    }
+    const [status, component, error] = useAsync(
+        async () => {
+            if (componentCache.has(widget.hash)) {
+                return componentCache.get(widget.hash)
```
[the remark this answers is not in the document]
EdAyers: Sorry I don't understand this.

lean4-infoview/src/infoview/util.ts
```diff
@@ -279,6 +279,17 @@ export function mapRpcError(err : unknown) : Error {
...
 }
 
+/** Catch handler for RPC methods that just returns undefined if the method is not found.
+ * This is useful for compatibility with versions of Lean that do not yet have the given RPC method.
+*/
+ export function discardMethodNotFound(e: unknown) : undefined {
+  if (isRpcError(e) && (e.code === RpcErrorCode.MethodNotFound)) {
+    return undefined
+  } else {
+      throw mapRpcError(e)
```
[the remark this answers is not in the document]
EdAyers: yep, added

lean4-infoview/src/infoview/info.tsx
```diff
@@ -190,7 +199,20 @@ export function InfoDisplay(props0: InfoDisplayProps) {
                     </div>
                 </Details>
             </div>
-            <div style={{display: hasMessages ? 'block' : 'none'}}>
+            {widgets && widgets.map(widget =>
+                <div style={{display: hasWidget ? 'block' : 'none'}}
+                     key={`widget::${widget.widgetSourceId}::${widget.range?.toString()}`}>
```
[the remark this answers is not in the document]
EdAyers: Yeah I was assuming multiple widgets with the same widgetid wouldn't be attached to the same point in the document which might be a bad assumption.

lean4-infoview/src/infoview/info.tsx
```diff
@@ -190,7 +199,20 @@ export function InfoDisplay(props0: InfoDisplayProps) {
...
+                    <Details initiallyOpen>
+                        <summary className="mv2 pointer">
+                            Widget: {widget.widgetSourceId}
```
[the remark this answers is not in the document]
EdAyers: ok, I don't see why though. What is the case where decl name and user-facing id would be different?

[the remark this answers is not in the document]
gebner: The "user-facing" name is `Lean.Widget.foobar` now, right?  That's not what we should show to the user (except as a tooltip maybe).

lean4-infoview/src/infoview/userWidget.tsx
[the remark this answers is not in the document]
gebner: When the position changes, `UserWidget` first shows the "pending" status, and starts a promise to fetch the javascript (via `useAsync`).  When this promise is finished, the `UserWidget` is rerendered.  The `UserWidget` function is called very often, and every call with a different position requires a second update.

Most of the time, the widget javascript is already in the cache (because there's like one widget and it's loaded once).  Then we don't need to do a second update.

I'm not sure if there's an easy and nice way to make `useAsync` return the cached value immediately (without going through the Promise.then -> set React reference -> React update dance).

lean4-infoview/src/infoview/rpcInterface.ts
gebner: See also [here](https://github.com/leanprover/lean4/pull/1238#discussion_r907679198).

lean4-infoview/src/infoview/info.tsx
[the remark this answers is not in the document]
EdAyers: What should we show to user?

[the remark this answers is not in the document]
gebner: Something (server-side) configurable? "Rubik's cube" or "Commutative diagram"?

lean4-infoview/src/infoview/userWidget.tsx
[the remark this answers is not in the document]
EdAyers: ok I think the solution here is remove async function's dependency on pos, it just needs _a_ pos to find a snapshot.

lean4-infoview/src/infoview/info.tsx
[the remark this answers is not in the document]
EdAyers: Also should the user-facing name be per `WidgetSource` or per `UserWidgetInfo`?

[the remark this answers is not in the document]
EdAyers: This is done

lean4-infoview/src/infoview/userWidget.tsx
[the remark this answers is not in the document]
EdAyers: fixed

lean4-infoview/src/infoview/rpcInterface.ts
EdAyers: Ok so the way this works now is that `UserWidget` has a field `javascript: String`, we can add more strings later if we really want. Then all of these source code strings are stored separately by their hash, and `getWidgetSource` just takes a hash and returns a string, but the hash is now stored as `javascriptHash`. Hence the getWidgetSource RPC just serves strings based on hash without knowing what type the string has.

gebner:
There are some merge conflicts due to #226 now.  I'm happy to migrate the code here to the new API if you want me to, but maybe it's better if you check out the differences yourself.

lean4-infoview/src/infoview/info.tsx
```diff
@@ -319,35 +347,46 @@ function InfoAux(props: InfoProps) {
...
-                setError(undefined);
-            }
-            else {
-                setError(`Error fetching goals: ${errS}`);
-                setStatus('error');
-            }
+            allReq = Promise.all([
+                goalsReq,
+                termGoalReq,
+                new Promise<undefined>(resolve => resolve(undefined))
```
gebner, line 353:
```suggestion
                undefined
```
In javascript, `T` is a subtype of `Promise<T>`.  For a particularly enlightening example, check this out:
```javascript
(async () => await console.log(await await await 42))()
```

lean4-infoview/src/infoview/rpcInterface.ts
```diff
@@ -183,3 +184,48 @@ export async function getGoToLocation(rs: RpcSessions, pos: DocumentPosition, ki
...
+    range?: Range
+}
+
+/** The response type for the RPC call `Widget_getWidgets`. */
+export interface UserWidgets {
+    widgets : UserWidgetInstance[]
+}
+
+/** Given a position, returns all of the user-widgets on the infotree at this position. */
+export function Widget_getWidgets(rs: RpcSessions, pos: DocumentPosition): Promise<UserWidgets | undefined> {
```
Vtec234, line 213: I started on the merge, but noticed the following thing: in our RPC calls, sending `DocumentPosition` actually makes no sense because we always use the URI at which the RPC call was made (so now, the `RpcSessionAtPos` URI). `getInteractiveDiagnostics` just takes in a line range, whereas `getInteractiveGoals` takes the whole `TextDocumentPositionParams` but the the handler just ignores it. Therefore I think we should change the widget APIs to just take a `Position` which is a line+character, but no URI. The goal APIs have been around so I guess we should keep them as-is for compat.

Vtec234: Ok, I made this change everywhere. Can always revert if we decide otherwise.

Issue #175, opened 2022-05-16
`lean --worker` crash goes unreported in InfoView.


lovettchris:
```lean
#eval (unsafeCast 0 : String)
```

This is an interesting one - the "worker" process dies but the "--server" process does not.  So Mariana's code doesn't kick in, we don't get "stopped" state from the LanguageClient.  I do see this pop up though - must be done by vscode?

![image](https://user-images.githubusercontent.com/18707114/168344668-02ecc403-58fb-4c2d-8fd5-97a19a0b7813.png)

And I do see this get printed to the debug console,  which comes from messages.tsx line 109, so it appears the infoview should be able to know about this and display something like this to the user:

getInteractiveDiagnostics error  {stack: 'Error: Server process for file:///d%3A/Temp/l…Trampoline (node:internal/async_hooks:130:17)', message: 'Server process for file:///d%3A/Temp/lean_ex…likely due to a stack overflow in user code.', code: -32603}

leodemoura:
It would be great to get this one fixed.

lovettchris:
Great, Mariana is actively working on it.

PR #195, opened 2022-06-17
fix: lean --worker crash goes unreported in InfoView.

marianaalanis93:
Issue related with this fix: https://github.com/leanprover/vscode-lean4/issues/175

This fix includes the next:

- Add a proper message in case the worker fails (for a bug or stackoverflow)
- Add a pop-up message in case (after fixing the bug) the user needs to restart the server
- Add new ex.codes related to worker exited or crashed. Current fix needs the latest version of lean4. Fix related to this: https://github.com/leanprover/lean4/pull/1209
      - Old ex.code: -32603 
      - New ex.code: -32901 and -32902


lovettchris (changes requested):

vscode-lean4/src/infoview.ts
```diff
@@ -96,7 +96,17 @@ export class InfoProvider implements Disposable {
...
-                return client.sendRequest(method, params);
+                try {
+                    const result = await client.sendRequest(method, params);
+                    return result
+                } catch (ex) {
+                    if (ex.code === -32901 || ex.code === -32902) {
+                        // ex codes related with worker exited or crashed
+                        console.log(`Lean worker exited or crashed: ${ex.message}`)
+                        // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+                        await this.onActiveClientStopped(client, false, 'Lean worker exited or crashed: ', ex.message)
```
lovettchris, line 107: I think this needs to be:
```typescript
const activeClient = client === this.clientProvider.getActiveClient();
await this.onActiveClientStopped(client, activeClient, 'Lean worker exited or crashed: ', ex.message as string)
```

marianaalanis93: good idea

vscode-lean4/src/leanclient.ts
```diff
@@ -100,10 +101,13 @@ export class LeanClient implements Disposable {
...
     async showRestartMessage(): Promise<void> {
-        const restartItem = 'Restart Lean Language Server';
-        const item = await window.showErrorMessage('Lean Language Server has stopped unexpectedly.', restartItem)
-        if (item === restartItem) {
-            void this.start();
+        if (!this.showingRestartMessage){
+            this.showingRestartMessage = true;
+            const restartItem = 'Restart Lean Language Server';
+            const item = await window.showErrorMessage('Lean Language Server has stopped unexpectedly.', restartItem)
+            if (item === restartItem) {
```
lovettchris, line 108: Also need to add `this.showingRestartMessage = false;` here after window.showErrorMessage returns (remember it is async and blocks as long as the popup is visible).

vscode-lean4/src/leanclient.ts
```diff
@@ -292,9 +296,9 @@ export class LeanClient implements Disposable {
                         this.restartedEmitter.fire(undefined)
                     }
                 } else if (s.newState === State.Stopped) {
-                    this.stoppedEmitter.fire('Lean language server has stopped. ');
-                    console.log('client has stopped or it failed to start');
                     this.running = false;
+                    this.stoppedEmitter.fire(['Lean language server has stopped. ', '']);
```
lovettchris, line 300: probably no need for the space after the period at the end of the message here.

vscode-lean4/src/leanclient.ts
marianaalanis93: The `this.showingRestartMessage = false;` is inside start() (inside restart()) function (finishing the function), so we can be sure that the client was restarted correctly. Do we need this `this.showingRestartMessage = false; ` again after showing the error message?

lovettchris: Fine, but it doesn't need to be that complicated, it can be right here on line 108.  This boolean is only to block multiple calls to this method from attempting to show the error message when it is already being shown.

lovettchris (approved):

lean4-infoview-api/src/infoviewApi.ts
```diff
@@ -105,7 +105,7 @@ export interface InfoviewApi {
   /** Must fire with the server's initialization message when the server is started or restarted. */
   serverRestarted(serverInitializeResult: InitializeResult): Promise<void>;
   /** Must fire with the server's message when the server is stopped. */
-  serverStopped(serverStoppedResult: string): Promise<void>;
+  serverStopped(serverStoppedMsg: string, serverStoppedReason: string): Promise<void>;
```
gebner, line 108: Is there a reason to split up the message here?  From what I can tell, the only difference is that reason is printed in red.  As far as I'm concerned, we can just print everything in red.  I'd rather keep it simple if possible.

*If* we really need to split it, then please introduce an interface:
```suggestion
  serverStopped(serverStoppedReason: ServerStoppedReason): Promise<void>;
  
...


export interface ServerStoppedReason {
  message: string;
  reason: string;
}
```

gebner (commented):
Another thing that should work is resetting the error on "refresh file dependencies":
1. Make a file with `#eval (unsafeCast 0 : String)`
2. Get the nice error message
3. Comment out the line
4. Run the "refresh file dependencies" command
5. Now the worker is running again, and the error message should be gone.

lean4-infoview-api/src/infoviewApi.ts
lovettchris: Thanks for the feedback, yes it is just for clarity of the error reporting, a title of what we think went wrong and a reason from the actual exception that we have no idea what it says but it might be useful.

Can you educate us on why the InfoView Api can't take multiple parameters and needs a structured object instead?

vscode-lean4/src/infoview.ts
```diff
@@ -96,7 +96,17 @@ export class InfoProvider implements Disposable {
...
-                return client.sendRequest(method, params);
+                try {
+                    const result = await client.sendRequest(method, params);
+                    return result
+                } catch (ex) {
+                    if (ex.code === -32901 || ex.code === -32902) {
+                        // ex codes related with worker exited or crashed
+                        console.log(`Lean worker exited or crashed: ${ex.message}`)
+                        const activeClient = client === this.clientProvider.getActiveClient();
+                        await this.onActiveClientStopped(client, activeClient, 'Lean worker exited or crashed: ', ex.message as string)
```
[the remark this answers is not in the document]
lovettchris: Yes, you are right we piggy backed the server stopped machinery for worker stopped because for a single file edit scenario the user cannot really see the difference, all they see is that the infoview stops working and the only way we have of fixing it is the sledge hammer of "Restart Server", so since the user experience is basically the same we saw reason not to reuse the same code we already had for server stopped.

lean4-infoview-api/src/infoviewApi.ts
gebner: Oh it can take multiple fields, no problem.  It's just that `[string, string]` is a very uninformative type and it would be better to use descriptive fields names here.

lovettchris: Got it, makes sense, Mariana, can you make that change?

vscode-lean4/src/infoview.ts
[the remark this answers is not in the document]
lovettchris: Oh, rats, it would have been nice to know this sooner.  Mariana and I have not been using "Refresh File Dependencies" in all our testing of this scenario and didn't realize that would fix the missing worker problem.   Gabriel are you saying this PR needs to change the error message terminology accordingly and use "Refresh File Dependencies" as the fix instead of "Restart Server" ?   Do you have some recommended language for the error message and for the restart button that would make sense to folks?  Something like:

message: "The Lean language service has stopped processing this file"
reason: ...
button: "Restart the Lean language service on this file" 

?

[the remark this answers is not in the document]
gebner: I honestly thought that this was obvious, as that command is the only way to restart a single file and I use it every few minutes. :facepalm:  But going again over the documentation, I can see how you missed it.  The name is misleading, and the readme even calls it a "work around for [technical issue]".  See also: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Renaming.20.22refresh.20file.20dependencies.22.3F

> Restart the Lean language service on this file

I don't think we ever use the term "language service" in user-facing text.  Usually it's "Lean server", so I'd go with "Restart the Lean server for this file".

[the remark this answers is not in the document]
lovettchris: It makes sense now, not knowing much about how the LSP works, somehow I thought refresh file dependencies was doing more that just restarting the worker process, like there were some sort of caches of dependencies that needed updating and that would fix all open files.  So this means if I have lots of files open and I change a core type they are all using then I have to run refresh file dependencies N times across all N open files?  If so, that kinda sucks...

vscode-lean4/src/leanclient.ts
```diff
@@ -100,17 +102,35 @@ export class LeanClient implements Disposable {
...
+            const item = await window.showErrorMessage('Lean Server has stopped unexpectedly.', restartItem)
+            this.showingRestartMessage = false;
+            if (item === restartItem) {
+                void this.start();
+            }
+        }
+    }
+
+    async showRestartFileMessage(): Promise<void> {
+        if (!this.showingRestartFileMessage) {
```
lovettchris, line 117: I think you can unify 'showingRestartMessage' and 'showingRestartMessage' as we will never be showing both messages at the same time.

vscode-lean4/src/leanclient.ts
```diff
@@ -100,17 +102,35 @@ export class LeanClient implements Disposable {
...
+            this.showingRestartFileMessage = true;
+            const restartItem = 'Restart the Lean Server on this file';
+            const item = await window.showErrorMessage('The Lean Server has stopped processing this file.', restartItem)
+            this.showingRestartFileMessage = false;
+            if (item === restartItem) {
+                if (window.activeTextEditor) {
+                    return this.restartFile(window.activeTextEditor.document);
```
lovettchris, line 124: I think await should be here instead of return.

vscode-lean4/src/leanclient.ts
```diff
@@ -460,6 +484,7 @@ export class LeanClient implements Disposable {
                 'text': doc.getText()
             }
         })
+        this.showingRestartFileMessage = false;
```
lovettchris, line 487: This doesn't belong here, the flag was already reset in showRestartFileMessage.

vscode-lean4/src/leanclient.ts
```diff
@@ -310,6 +330,7 @@ export class LeanClient implements Disposable {
             }
             // if we got this far then the client is happy so we are running!
             this.running = true;
+            this.showingRestartMessage = false;
```
lovettchris, line 333: modifying `this.showingRestartMessage` doesn't belong here, it should be localized to  the showRestartMessage method.

vscode-lean4/src/leanclient.ts
```diff
@@ -409,6 +430,7 @@ export class LeanClient implements Disposable {
         assert(() => this.isStarted())
         if (this.client && this.running) {
             this.noPrompt = true;
+            this.showingRestartMessage = true
```
lovettchris, line 433: modifying `this.showingRestartMessage` doesn't belong here, it should be localized to  the showRestartMessage method.

vscode-lean4/test/suite/utils/helpers.ts
```diff
@@ -54,6 +54,14 @@ export async function insertText(text: string) : Promise<void> {
     });
 }
 
+export async function deleteAllText() : Promise<void> {
+    const editor = vscode.window.activeTextEditor;
+    assert(editor !== undefined, 'no active editor');
+    await editor.edit((builder) => {
+        builder.delete(new vscode.Range(new vscode.Position(0, 0), new vscode.Position(editor.document.lineCount-1, 1000)));
```
lovettchris, line 61:
```typescript
builder.delete(new vscode.Range(new vscode.Position(0, 0), editor.document.lineAt(editor.document.lineCount-1).range.end));
```

vscode-lean4/test/suite/toolchains/toolchain.test.ts
```diff
@@ -3,12 +3,56 @@ import { suite } from 'mocha';
...
+
+		console.log('Checking that it crashed again.')
+		await assertStringInInfoview(info, expectedMessage);
+
+		console.log('deleting the problematic string closing active editors and restarting the server')
+		await deleteAllText();
+		await insertText(`#eval "${hello}"`);
+		console.log('Now invoke the restart server command')
+		client = assertActiveClient(clients);
+		await restartLeanServer(client);
```
lovettchris, line 46: Let's make this one a "restartFile" command so we have some test coverage that this works also.

vscode-lean4/src/leanclient.ts
marianaalanis93: Ok, makes sense

lovettchris (approved):

vscode-lean4/src/infoview.ts
```diff
@@ -61,7 +61,7 @@ export class InfoProvider implements Disposable {
 
     private rpcSessions: Map<string, RpcSession> = new Map();
 
-    private clientsFailed: Map<string, string> = new Map();
+    private clientsFailed: Map<string, [string, string]> = new Map();
```
gebner, line 64:
```suggestion
    private clientsFailed: Map<string, ServerStoppedReason> = new Map();
```

vscode-lean4/src/infoview.ts
```diff
@@ -314,30 +318,52 @@ export class InfoProvider implements Disposable {
...
+        }
+        await this.sendPosition();
+    }
+
     onClientRemoved(client: LeanClient) {
         // todo: remove subscriptions for this client...
     }
 
-    async onActiveClientStopped(client: LeanClient, activeClient: boolean, msg: string) {
+    async onActiveClientStopped(client: LeanClient, activeClient: boolean, msg: string, reason: string, restartFile: boolean = false) {
```
gebner, line 345:
```suggestion
    async onActiveClientStopped(client: LeanClient, activeClient: boolean, reason: ServerStoppedReason, restartFile: boolean = false) {
```

vscode-lean4/src/leanclient.ts
```diff
@@ -71,7 +72,7 @@ export class LeanClient implements Disposable {
     private progressChangedEmitter = new EventEmitter<[string, LeanFileProgressProcessingInfo[]]>()
     progressChanged = this.progressChangedEmitter.event
 
-    private stoppedEmitter = new EventEmitter<string>()
+    private stoppedEmitter = new EventEmitter<[string, string]>()
```
gebner, line 75:
```suggestion
    private stoppedEmitter = new EventEmitter<ServerStoppedReason>()
```

vscode-lean4/src/infoview.ts
```diff
@@ -314,30 +318,52 @@ export class InfoProvider implements Disposable {
...
             // means that client and active client are the same and just show the error message
-            await this.webviewPanel?.api.serverStopped(msg);
+            await this.webviewPanel?.api.serverStopped({message:msg, reason});
         }
 
-        console.log(`client stopped: ${client.getWorkspaceFolder()}`)
-
         // remember this client is in a stopped state
-        this.clientsFailed.set(client.getWorkspaceFolder(), msg)
+        const key = client.getWorkspaceFolder();
```
gebner, line 355: This will still show an error message for *all* files in the same directory, right?  We only want to show an error message for the file that crashed.  (Maybe I'm wrong, didn't try it out.)

lovettchris: Good catch...

marianaalanis93: yeah you're right, fixing it

vscode-lean4/src/infoview.ts
```diff
@@ -97,7 +102,18 @@ export class InfoProvider implements Disposable {
...
+                    const result = await client.sendRequest(method, params);
+                    return result
+                } catch (ex) {
+                    if (ex.code === -32901 || ex.code === -32902) {
+                        // ex codes related with worker exited or crashed
+                        console.log(`The Lean Server has stopped processing this file: ${ex.message}`)
+                        await this.onWorkerStopped(uri, client, {message:'The Lean Server has stopped processing this file: ', reason: ex.message as string})
+                    } else {
+                        // file updated, do nothing
+                    }
```
gebner, line 115:
```suggestion
                    }
                    throw ex;
```
Exceptions are also used for "normal" control flow in the RPC interface (and the infoview code wants to get them to relay it to the user).

Note that this function here only intercepts some of the LSP request, namely those that are initiated by the infoview.  (Which should be enough since the infoview does lots of them while it's open.  I'm not even sure if we can intercept all LSP requests with the vscode-languageserver-node library.)

gebner:
This works quite nicely now!

lovettchris (changes requested):
There is something different about how restart server works in this branch such that the toolchain tests fail pretty reliably.  One stage of the test after a restart the info view is not updating.  Can you investigate?

[this is resolved now]

gebner:
The only thing missing here is the failing tests.  Rest lgtm.

lovettchris:
> The only thing missing here is the failing tests. Rest lgtm.

@gebner the tests are now passing, I only saw one random failure after multiple retries which is [this one](https://github.com/leanprover/vscode-lean4/issues/228) which seems unrelated.

lovettchris (approved):

PR #198, opened 2022-06-20
Add pinning infoview test.

lovettchris:
Remove obsolete index.ts file that was missed previously.


lean4-infoview/src/infoview/rpcSessions.tsx
```diff
@@ -146,6 +146,7 @@ export class RpcSessions implements Disposable {
     async call<T>(pos: DocumentPosition, method: string, params: any): Promise<T | undefined> {
         const sesh = await this.sessionAt(pos.uri)
         if (!sesh) {
+            console.log(`call ${method} dropped because session does not yet exist`)
```
Vtec234, line 149: We may not want this log in production, my guess is it might show up a lot and spam the console.

lean4-infoview/src/infoview/messages.tsx
```diff
@@ -188,13 +188,14 @@ export function useMessagesForFile(uri: DocumentUri, line?: number): Interactive
...
                     // This can happen while typing quickly, so server will catch up on next edit.
+                    console.log('getInteractiveDiagnostics: document has been changed since we made the request ')
                 } else {
                     console.log('getInteractiveDiagnostics error ', err)
                 }
             }
         }
     }
-    React.useEffect(() => void updateDiags(), [uri, line, lspDiags.get(uri)])
+    React.useEffect(() => void updateDiags(), [uri, line, rs, lspDiags.get(uri)])
```
Vtec234, line 198: Thanks for catching this! I think we should use `rs.sessionIdAt(uri)` here instead, it was made specifically to serve as a dep but I forgot to actually use it :| It might be missing in other places.

lean4-infoview/src/infoview/messages.tsx
```diff
@@ -188,13 +188,14 @@ export function useMessagesForFile(uri: DocumentUri, line?: number): Interactive
                 if (err?.code === -32801) {
                     // Document has been changed since we made the request.
                     // This can happen while typing quickly, so server will catch up on next edit.
+                    console.log('getInteractiveDiagnostics: document has been changed since we made the request ')
```
Vtec234, line 191: Similarly this log, we often take this branch so it's a lot of debug output.

lean4-infoview/src/infoview/rpcSessions.tsx
lovettchris: These console.log messages only show up in the console of the debugger we use while developing the extension.  For example, the debugger in the background has the console.log messages but the "user's vs code" instance shows nothing:

image

lean4-infoview/src/infoview/messages.tsx
lovettchris: Thanks, I switched to this, it seems to work.  I also checked AllMessages and it already had the "rs" dependency which is probably why AllMessages was fine.

lean4-infoview/src/infoview/messages.tsx
lovettchris: again this is only vscode extension debug console output, user's will not see this stuff.

lovettchris:
@Vtec234 - the test still fails after merging with master, at line 97 in the info.tests.ts tha I just pushed, the pinned content should show this:

image

but instead it switches to this when I add the new eval statement.

image

lovettchris:
@Vtec234, well I pushed a "workaround" fix for this by removing the insertText call that triggeres that bug.  So this test now passes, but definitely need to debug the former problem because it appears to reveal a bug in how pinning works after editing the file.

lean4-infoview/src/infoview/messages.tsx
Vtec234: Yep I know, sorry, 'production' was maybe not the best choice of word. What I mean is that I'm not sure these messages are useful even for us developers outside of debugging this particular PR, so I would remove them as leftover debug statements.

Vtec234:
Thank you for finding these bugs, sounds like having tests is useful :) I do not have the cycles currently. It would be awesome if you could look into it.

lean4-infoview/src/infoview/messages.tsx
lovettchris: Well it helped me find the very thing that caused the bug I fixed in messages.tsx - but it is not that noisy, I don't see this message very often, usually only during startup of a new editing session.  This is what a typical debug session console output looks like these days:
image

then nothing more while editing the .lean file unless I do a command like Restart Server, or change the version.  The biggest  noise in the console is from things like this that pop up randomly:

image

lovettchris: But I'll agree to remove this one, since I can't guarantee it doesn't get noisy in some edge case I haven't found.

Vtec234: Hmmm, the 'Got unexpected error form' looks like another bug.

lovettchris:
> Thank you for finding these bugs, sounds like having tests is useful :) I do not have the cycles currently. It would be awesome if you could look into it.

I looked into it and propose this fix: https://github.com/leanprover/vscode-lean4/issues/200

lovettchris:
@Vtec234 - I'd like to merge this new test now so that I can try adding back the insertText test code in my new branch and other tricky edits that could break pinned values.

PR #199, opened 2022-06-21
feat: preserve pins when cursor position is undefined

Vtec234:
When we close a Lean file, the cursor position temporarily becomes `undefined` and clears all pins. This is fixed here by keeping the `Infos` component alive for as long as the server is running and adapting it to handle the `undefined` case.

### DIFF
--- a/lean4-infoview/src/infoview/infos.tsx
+++ b/lean4-infoview/src/infoview/infos.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { DidChangeTextDocumentParams, DidCloseTextDocumentParams, Location, TextDocumentContentChangeEvent } from 'vscode-languageserver-protocol';
 
 import { EditorContext } from './contexts';
-import { DocumentPosition, Keyed, PositionHelpers, useClientNotificationEffect, useClientNotificationState, useEvent } from './util';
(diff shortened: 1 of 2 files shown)


lovettchris:
Can you give full repro steps so I can add this to the new unit test?  I assume you are not talking about closing the only lean file that is open, in that case the info view switches to "Click somewhere in the Lean file to enable the infoview".... ?

Vtec234:
A test would be great :) Yes, it's not what you describe. The repro is:
1. Open A.lean and B.lean.
2. While in A.lean, pin some location.
3. Go to B.lean.
4. Close B.lean. At this point, while the editor auto-switches back to A.lean, the cursor is undefined for a split second.
5. Back in A.lean, before the PR the pin was gone. Now it is preserved.

lovettchris:
Ok, I added the new test, see 'Pin survives file close' and it currently fails, it should pass when your fix goes in.

I also found and fixed another bug this test found which is the infoview sometimes doesn't refresh during initialization so it says "no info found", this was due to a [missing dependency](https://github.com/leanprover/vscode-lean4/pull/198/files#diff-f3205301d3c8487365c663b3f0147c3cb14360f4d68f4409554ad27328c56025) in messages.tsx implementation of `useMessagesForFile` where the interactive diagnostics need to be recomputed if the rpcSessions object changes, which happens after connectAt finishes on line 150 of rpcSessions.tsx.

lovettchris (approved):

PR #201, opened 2022-06-23
fix: pin repositioning algorithm

### LINKED ISSUE #200: Pinned selections are not properly anchored
Repro:
1. Open a new .lean file and add this to the top of the file:
    `#eval 123`
2. Pin that, so you should see this:
image
3. Go to the end of the line and type ENTER

**Actual**: infoview shows this error
image

**Expected**: pinned selections should be properly anchored.

The bug appears to be in infos.tsx where it is listening to `textDocument/didChange` and trying up update the pin positions from the change information in the `DidChangeTextDocumentParams` but the code is buggy and very incomplete.  Notice in this case the code has moved the pin to line 2 character 9, which is a position that doesn't exist in the document - hence the error.

What we really need is a proper "anchored selection", or what some people call a Tracked Selection that moves with prior code edits and can also be "deleted" if the user deletes a range of text including this tracked selection position.  Then the infoview should listen to "move" and "delete" events on this tracked selection and update the pin positions accordingly.  So I propose we put all this tracking in the editor extension, and simplify the infoview code so it is simply using a new InfoViewApi "createTrackedSelection" or something, and a `useClientNotificationState` to listen to changes on these tracked selections.

Vtec234:
Hopefully fixes #200.


lovettchris:
Can you try your fixes with this updated test: https://github.com/leanprover/vscode-lean4/pull/204

lean4-infoview/src/infoview/infos.tsx
```diff
@@ -29,19 +29,38 @@ export function Infos() {
...
-                    newPin.character = newPin.line > chg.range.end.line ?
-                        newPin.character :
-                        lines === 0 ?
-                            chg.range.start.character + Math.max(0, newPin.character - chg.range.end.character) + chg.text.length :
-                            9999;
+                    if (PositionHelpers.isLessThanOrEqual(newPin, chg.range.start)) continue;
+                    // We can assume chg.range.start < pin
+
+                    // If the pinned position is replaced with new text, just delete the pin.
+                    if (PositionHelpers.isLessThanOrEqual(newPin, chg.range.end)) {
```
lovettchris, line 36: is LessThanOrEqual correct here?  If I pin the end of this line:
```lean
#eval [1,2,3]
```
Then replace the string `[1,2,3]` with `456` the pin disappears which is not really expected in this case?

lean4-infoview/src/infoview/infos.tsx
```diff
@@ -69,38 +88,38 @@ export function Infos() {
 
     // Update pins on UI actions
     const pinKey = React.useRef<number>(0);
-    const isPinned = (pinnedPoss: DocumentPosition[], pos: DocumentPosition) => {
-        return pinnedPoss.some(p => DocumentPosition.isEqual(p, pos));
+    const isPinned = (pinnedPositions: DocumentPosition[], pos: DocumentPosition) => {
```
lovettchris, line 91: thanks, much better name.

lean4-infoview/src/infoview/infos.tsx
```diff
@@ -69,38 +88,38 @@ export function Infos() {
...
         if (!curPos) return
-        setPinnedPoss(pinnedPoss => {
-            if (isPinned(pinnedPoss, curPos)) {
-                return pinnedPoss.filter(p => !DocumentPosition.isEqual(p, curPos));
+        setPinnedPositions(pinnedPositions => {
+            if (isPinned(pinnedPositions, curPos)) {
+                return pinnedPositions.filter(p => !DocumentPosition.isEqual(p, curPos));
             } else {
                 pinKey.current += 1;
-                return [ ...pinnedPoss, { ...curPos, key: pinKey.current.toString() } ];
```
lovettchris, line 98: Can you explain to me what the key is for? I see each pin is getting a unique key, but why? I don't really see the key being used anywhere.  Everything is using DocumentPosition.isEqual(p, pos) to compare saved pinned positions, nothing is comparing the key.  Or is it just to trigger a render in React when the pinKey.current value is incremented?

lean4-infoview/src/infoview/infos.tsx
Vtec234: Yep, the start/end of the line are edge cases that have to be handled somehow, so I chose to do it this way. The interpretation in this PR is that the pin is "at" the character preceding the cursor when the position is pinned. In your example, when the cursor is at the end, the pin is at `]`. When replaced with `456`, the bracket is gone and so is the pin. The symmetric choice would be to go with the character *after* the cursor. But that would be a newline which is just weird (it seems nonsensical to pin a newline).

lean4-infoview/src/infoview/infos.tsx
Vtec234: Definitely! The `key` property is required by React to uniquely identify components within a list. It is not used by the components themselves. It is explained [here](https://reactjs.org/docs/lists-and-keys.html#keys).

Vtec234:
> Can you try your fixes with this updated test: #204

Seems happy!

lovettchris:
> Seems happy!

👍

PR #208, opened 2022-07-02
fix: allow lean language server to run without elan installed.

lovettchris:


vscode-lean4/src/utils/leanInstaller.ts
```diff
@@ -331,34 +328,40 @@ export class LeanInstaller implements Disposable {
...
             // If folderPath is undefined, this will use the process environment for cwd.
             // Specifically, if the extension was not opened inside of a folder, it
             // looks for a global (default) installation of Lean. This way, we can support
             // single file editing.
+            console.log(`executeWithProgress ${cmd} ${options}`)
             const stdout = await this.executeWithProgress('Checking Lean setup...', cmd, options, folderPath)
             if (!stdout) {
-                return { version: '', error: 'lean not found' };
+                result.error = 'lean not found'
             }
```
Kha, line 342: I think this conditional is redundant: if we have `!stdout`, we will also have `!match` and so we `return` below without actually using `result`

Kha (commented):
Thanks! I can give this a try on Monday.

vscode-lean4/src/utils/leanInstaller.ts
lovettchris: the diff is confusing, it is now an `if (!stdout) { ... } else if (stdout.indexOf('no default toolchain') > 0) { ... } else { ... }` and the version matching is in the last else, so there is no redundant conditional.

Kha: Alright, but then the inner `!stdout` conditional is redundant

Kha:
Yup, seems to work! Thanks!

vscode-lean4/src/utils/leanInstaller.ts
```diff
@@ -331,34 +328,40 @@ export class LeanInstaller implements Disposable {
...
-                    return { version: '', error: `lean4: '${cmd}' program not found.` }
-                } else {
-                    return { version: '', error: `lean4: '${cmd} ${options}' returned incorrect version string '${stdout}'.` }
+            else {
+                const filterVersion = /version (\d+)\.\d+\..+/
+                const match = filterVersion.exec(stdout)
+                if (!match) {
+                    if (!stdout) {
+                        return { version: '', error: `lean4: '${cmd}' program not found.` }
+                    } else {
```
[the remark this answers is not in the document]
lovettchris: ooooh, thanks. sorry missed that one.  Fixed now.

lovettchris:
@gebner - are you good with this one?

gebner:
Okay, was just waiting for Sebastian.

PR #209, opened 2022-07-05
Add setting to specify the path to the Lake executable

tydeu:
This is a very selfish addition, but one I desperately need in order to test `lake serve` without needing to setup a custom Lean toolchain on every rebuild.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,8 @@ This extension contributes the following settings (for a complete list, open the
 
 * `lean4.toolchainPath`: specifies the location  of the Lean toolchain to be used when starting the Lean language server. Most users (i.e. those using `elan`) should not ever need to change this. If you are bundling Lean and `vscode-lean` with [Portable mode VS Code](https://code.visualstudio.com/docs/editor/portable), you might find it useful to specify a relative path to Lean. This can be done by starting this setting string with `%extensionPath%`; the extension will replace this with the absolute path of the extension folder. For example, with the default directory setup in Portable mode, `%extensionPath%/../../../lean` will point to `lean` in the same folder as the VS Code executable / application.
 
+* `lean4.lakePath`: specifies the location of the Lake executable to be used when starting the Lean language server (when possible). If left unspecified, the extension defaults to the Lake executable bundled with the Lean toolchain. Most users thus do not need to use this setting. It is only really helpful if you are building a Lake executable from the source and wish to use it with this extension.
+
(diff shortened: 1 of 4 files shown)


lovettchris:
Would https://github.com/leanprover/vscode-lean4/pull/208 give you the same benefit?

tydeu:
Also, @gebner, thanks for the merge!

Kha:
@tydeu Hmm, couldn't you put your custom `lake` in front of `elan` in the `PATH`? Not that the new option is much of a maintenance burden.

lovettchris:
if you modify the path in your windows command prompt and run "code ." vscode does inherit the PATH from that command prompt.  I have command prompts setup on windows that do things like this and then I pin this to my taskbar, so each time I login and start work, I click that thing I get a terminal window all in the right configuration for that work, without touching my system PATH:
```
call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
set PATH=%PATH%;C:\Program Files (x86)\Microsoft SDKs\Azure\AzCopy\
set PATH=%PATH%;d:\tools
set PATH=%PATH%;C:\Program Files (x86)\WiX Toolset v3.11\bin
set PATH=%PATH%;C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\Llvm\x64\bin\
set PATH=%PATH%;C:\Program Files (x86)\Windows Kits\10\Debuggers\x64
set PATH=%PATH%;C:\Program Files\PostgreSQL\13\bin
set PATH=%PATH%;%ALLUSERSPROFILE%\chocolatey\bin
set PATH=%PATH%;C:\Program Files\Git\cmd\;C:\Program Files\Git\usr\bin
set PATH=%PATH%;C:\Users\clovett\AppData\Roaming\npm
d:\tools\alias -f d:\tools\alias.txt
pushd D:\git\leanprover\vscode-lean4
```

tydeu:
@lovettchris Fair enough. I do think this PR is a more user-friendly approach, though. It also doesn't override outside the specific projects I want too, which is nice. I am also a weird dev in that I much prefer point-and-click GUI interfaces to CLIs (hence one of the reasons I use Windows & VSCode and not Linux & vim/emacs). 😅

lovettchris:
Yeah, being able to have a per-project specific setting is nice.  I agree with Sebastian this is not a huge support burden, so long as you don't expect this to be fully tested and therefore never break :-) Unless you want to also add some unit tests... hint hint...  PR https://github.com/leanprover/vscode-lean4/pull/208 contains a new test that messes with the user's PATH for testing that everything works when elan is missing.

tydeu:
>  as you don't expect this to be fully tested and therefore never break

That seems fine to me. If it breaks, I can just fix it myself and PR. Especially since I am likely to be the only person regularly using it.

PR #213, opened 2022-07-07
chore: add restartFile command

gebner:


gebner:
@lovettchris Does the new command name and the explanation in the readme make it clearer what's going on?

lovettchris:
I see thanks, it is more clear to me now that this restarts the file worker, but I'm not sure it will be more clear to users that know nothing about what a file worker is.  I think refreshFileDependencies is more "user centric" in terms of what it is usually used for.

lovettchris (approved):

gebner:
> I'm not sure it will be more clear to users that know nothing about what a file worker is.

Note that the final version of this PR does not contain the words "file worker" anymore.  It now only talks about "restarting server for a single file", which is hopefully understable.

lovettchris:
Nice catch, "refresh file dependencies" is still there in vs code, it is just an alias for "restart file".  But I did a bigger search and found "refresh file dependencies" is also mentioned in the quickstart.md, and on line 96 of the vscode-lean4 readme.  It was not mentioned in TPIL or fp-lean books.  I'll leave it to Gabriel to decide if we want to switch all mentions of "refresh file dependencies" over to the new name...

PR #217, opened 2022-07-11
fix: issue# 216 about powershell execution policy.

### LINKED ISSUE #216: Problem with PowerShell ExecutionPolicy
See https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/VSCode.20.22Waiting.20for.20lean.20server.20to.20start.2E.2E.2E.22

Students may not know how to run `Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope CurrentUser` needed in order for `elan-init.ps1` to work.

We could run it for them before running `elan-init.ps1` then set it back to whatever level it was before.  We can also set a safer level of `Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser` which still works because `elan-init.ps1` is run locally, not remotely.

lovettchris:
Fixes #216 

PowerShell by default has execution policy set to Restricted, meaning it will not run random powershell scripts.  This causes `elan-init.ps1` to fail during bootstrapping which is not ideal.  This fix temporarily changes the powershell execution policy to RemoteSigned so that the script can run, then restores the policy to what it was before after  `elan-init.ps1` completes.

This also fixes a recent regression that was causing: "waiting for lean server to start..." to happen in the case where the `lean` version check failed - it was caching the fact that it fails and didn't clear the cache after installing elan.

### DIFF
--- a/vscode-lean4/src/utils/leanInstaller.ts
+++ b/vscode-lean4/src/utils/leanInstaller.ts
@@ -5,6 +5,7 @@ import { LocalStorageService} from './localStorage'
 import { readLeanVersion, findLeanPackageRoot, isCoreLean4Directory } from './projectInfo';
 import { join, dirname } from 'path';
 import { fileExists } from './fsHelper'
+import { addDefaultElanPath, getDefaultElanPath, addToolchainBinPath} from '../config';
 
 export class LeanVersion {
     version: string;
@@ -42,6 +43,7 @@ export class LeanInstaller implements Disposable {
     }
 
     async testLeanVersion(packageUri: Uri) : Promise<LeanVersion> {
+
         // see if there is a lean-toolchain file and use that version info.
         let leanVersion : string | null = await readLeanVersion(packageUri);
         if (!leanVersion) {
@@ -148,6 +150,11 @@ export class LeanInstaller implements Disposable {
             // no prompt, just do it!
             console.log('Installing Lean via Elan during testing')
             await this.installElan();
+            if (typeof(process.env.DISABLE_ELAN) === 'string') {
+                addToolchainBinPath(getDefaultElanPath());
+            } else {
+                addDefaultElanPath();
+            }
             this.installChangedEmitter.fire(uri);
             return;
         }
@@ -185,6 +192,8 @@ export class LeanInstaller implements Disposable {
                 if (s === 'reset') {
                     s = '';
                 }
+                // ensure this version is actually installed.
+                await this.executeWithProgress('Ensure toolchain available...', 'lean', ['+' + s, '--version'], null);
                 this.localStorage.setLeanPath(''); // make sure any local full path override is cleared.
                 this.localStorage.setLeanVersion(s); // request the specified version.
                 this.installChangedEmitter.fire(uri);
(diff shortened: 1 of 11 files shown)


vscode-lean4/src/utils/leanInstaller.ts
```diff
@@ -463,23 +475,33 @@ export class LeanInstaller implements Disposable {
...
             });
 
             if (process.platform === 'win32') {
+                executionPolicy = (await this.getExecutionPolicy()).trim();
+                if (executionPolicy === 'Restricted'){
+                    console.log('Setting ExecutionPolicy to RemoteSigned so we can run elan-init.ps1')
+                    await this.setExecutionPolicy('RemoteSigned');
+                }
                 terminal.sendText(
                     `Invoke-WebRequest -Uri "${this.leanInstallerWindows}" -OutFile elan-init.ps1\n` +
```
Kha, line 506: Can't you just run `Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope Process` in this script?

tydeu: I suppose the goal is to make sure the script in question is properly signed, which is more secure against the (admittedly unlikely) case that the downloaded `elan-init.ps1` was corrupted.

lovettchris: @Kha , you are right, that is a much better solution, is now pushed.
@tydeu , yes, since the scope is now Process, and this will be limited to the terminal window we open to install elan and the terminal is closed after the install finishes, the only real attack vector there is elan-init.ps1 itself, so yes.

Kha: I mostly used `Unrestricted` because `RemoteSigned` kind of implies that `elan-init.ps1` is signed, which it's not. It feels like an unnecessary gotcha that `Invoke-WebRequest` does not mark the downloaded script as a remote file.

lovettchris: RemoteSigned is safer so I'd like to leave it there.  Nothing we can do about the fact that powershell chose some weird names.  The theory is that if the app has permission to do Invoke-WebRequest to download something locally, then the app already has enough trust to run these scripts, but as Mac pointed out, RemoteSigned  still guards against that script doing something unexpected.

lovettchris: Never mind, I decided to switch to Unrestricted because I found other problems with RemoteSigned.

vscode-lean4/src/utils/leanInstaller.ts
```diff
@@ -465,15 +481,15 @@ export class LeanInstaller implements Disposable {
...
             let terminalOptions: TerminalOptions = { name: terminalName };
             if (process.platform === 'win32') {
-                const windir = process.env.windir
-                terminalOptions = { name: terminalName, shellPath: `${windir}\\System32\\WindowsPowerShell\\v1.0\\powershell.exe` };
+                terminalOptions = { name: terminalName, shellPath: getPowerShellPath() };
             }
             const terminal = window.createTerminal(terminalOptions);
             terminal.show();
 
+            const thisObject = this;
```
gebner, line 489: Unused.

lovettchris: fixed.

PR #218, opened 2022-07-12
test: Add timestamping on all console.log output.

lovettchris:
This makes it easier to monitor relative performance and timeouts in our tests.

### DIFF
--- a/vscode-lean4/src/utils/logger.ts
+++ b/vscode-lean4/src/utils/logger.ts
@@ -1,17 +1,16 @@
 import { Console } from 'console'
-const util = require('util');
 
 class Logger extends Console {
     constructor(stdout: NodeJS.WritableStream, stderr?: NodeJS.WritableStream) {
         super(stdout, stderr);
     }
 
-    log(...args : any) {
-        super.log(new Date().toLocaleTimeString(), '-', util.format(...args));
+    log(msg: string) {
+        super.log(new Date().toLocaleTimeString(), '-', msg);
     }
 
-    error(...args : any) {
-        super.error(new Date().toLocaleTimeString(), '-', util.format(...args));
+    error(msg: string) {
+        super.error(new Date().toLocaleTimeString(), '-', msg);
     }
 }
 
--- a/vscode-lean4/src/utils/logger.ts
+++ b/vscode-lean4/src/utils/logger.ts
@@ -1,0 +1,19 @@
+import { Console } from 'console'
+const util = require('util');
+
+class Logger extends Console {
+    constructor(stdout: NodeJS.WritableStream, stderr?: NodeJS.WritableStream) {
+        super(stdout, stderr);
+    }
+
+    log(...args : any) {
+        super.log(new Date().toLocaleTimeString(), '-', util.format(...args));
+    }
+
+    error(...args : any) {
+        super.error(new Date().toLocaleTimeString(), '-', util.format(...args));
+    }
+}
+
+const logger = new Logger(process.stdout, process.stderr);
+export { logger }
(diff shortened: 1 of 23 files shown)


vscode-lean4/src/utils/logger.ts
```diff
@@ -0,0 +1,18 @@
+import { Console } from 'console'
+
+class Logger extends Console {
+    constructor(stdout: NodeJS.WritableStream, stderr?: NodeJS.WritableStream) {
+        super(stdout, stderr);
+    }
+
+    log(msg: string) {
+        super.log(new Date().toLocaleTimeString(), '-', msg);
```
Vtec234, line 9:
```suggestion
        super.log(new Date().toLocaleTimeString('en-US'), '-', msg);
```
So it doesn't depend on the runner's locale.

vscode-lean4/src/utils/logger.ts
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    error(msg: string) {
+        super.error(new Date().toLocaleTimeString(), '-', msg);
```
Vtec234, line 13:
```suggestion
        super.error(new Date().toLocaleTimeString('en-US'), '-', msg);
```

vscode-lean4/test/suite/bootstrap/bootstrap.test.ts
```diff
@@ -22,34 +23,10 @@ suite('Lean4 Bootstrap Test Suite', () => {
         assert(info, 'No InfoProvider export');
 
         // give it a extra long timeout in case test machine is really slow.
+        logger.log('Wait for elan install of Lean nightly build...')
 		await waitForInfoviewHtml(info, '4.0.0-nightly-', 600);
 
-        // test goto definition to lean toolchain works
-        await waitForActiveEditor();
```
Vtec234, line 28: Did you intend to remove all this?

Vtec234 (approved):

vscode-lean4/test/suite/bootstrap/bootstrap.test.ts
lovettchris: yes, I copied this from the simple test which is also testing goto definition, but I didn't need that in the bootstrap test.

vscode-lean4/src/utils/logger.ts
lovettchris: I don't really like that because it's too long, I don't really need absolute date/time anyway, this is mostly just for relative timing information - the number of seconds ticking over mostly.  Do you want the full UTC string for some other reason?
```
Wed, 13 Jul 2022 23:33:20 GMT - ...
```
I don't feel super strongly about it, just curious mostly...

Vtec234: Nope, I thought it might just be a bit confusing since times will be out of whack if, say, one action runs in `us-east` and another in `us-west`. If you only need relative timestamps, many languages provide a [monotonic system clock](https://doc.rust-lang.org/std/time/struct.Instant.html). But I'm not sure JS does, so let's just go with whatever you prefer here.

Vtec234: Maybe you could use `Math.round(performance.now() / 1000)` ([`performance.now`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now))

vscode-lean4/src/utils/logger.ts
lovettchris: let's just go with what I have now we can easily switch to something else if we need to later.

lovettchris:
I discovered this logger in vccode typescript-language-features which is interesting - it is integrated into the vscode.OutputChannela and has this for timestamping:

```typescript
private now(): string {
    const now = new Date();
    return padLeft(now.getUTCHours() + '', 2, '0')
	+ ':' + padLeft(now.getMinutes() + '', 2, '0')
	+ ':' + padLeft(now.getUTCSeconds() + '', 2, '0') + '.' + now.getMilliseconds();
```
and it also has a `LogLevel`...

PR #222, opened 2022-07-15
feat: show loading message in trace explorer

Vtec234:
A marginal improvement that makes it clear that the trace is just loading and not broken.


gebner:
Could we use the useAsync helper from the widgets PR here?  I'm not really a fan of hand-written state machine code.

Vtec234:
Hm, the problem is that `useAsync` has 'eager' semantics which fetch immediately, but in this component we only want to fetch when the subtrace is opened. We can do
```typescript
    type State = 'collapsed' | 'open'
    const [st, setSt] = React.useState<State>('collapsed')

    const [status, tt, err] = useAsync<TaggedText<MsgEmbed>>(async () => {
        if (st === 'collapsed') throw ''
        const tt = await InteractiveDiagnostics_msgToInteractive(rs, pos, msg, col)
        if (tt) return tt
        else throw 'no trace found'
    }, [st, pos.uri, pos.line, pos.character, msg, col])
```
but I am not sure this is any clearer.

gebner:
Ugh, you're right.  Note that we still have the same race condition that motivated `useAsync` (but it's on master as well so not a regression).  Let's merge this for now.

PR #229, opened 2022-07-28
feat: configurable output autofocus

tydeu:
As discussed in [this Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/VSCode.20Disable.20Output.20auto-focus/near/291145653), this PR makes the extension's behavior of automatically focusing the Output panel on new messages configurable by the user. I also defaulted the option to `false`, because as @Vtec234 [mentions](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/VSCode.20Disable.20Output.20auto-focus/near/291236099), this feature is generally undesired and primarily useful when debugging the server (which should be rare).

Issue #181, opened 2022-05-19
"Failed to start language server" popup is too generic


Kha:
The server could fail to start due to various reasons, but the popup does not show the actual error (or instruct users how to find it), nor does it hide the "install Elan" button when that is unlikely to be helpful.

![image](https://user-images.githubusercontent.com/109126/169322661-e1137b7b-38bf-41ac-9920-96318e2a713f.png)

Kha:
I don't think it's displayed by default, is it?

lovettchris:
Good point, I used to pop up the output window automatically in this case, but the someone said they didn't like that :-(

lovettchris:
But I could add a button "see output" that does that...

PR #230, opened 2022-07-28
fix: Failed to start language server: popup is too generic

marianaalanis93:
This fix adds a button to see the output (it redirects you to the output section inside VSCode)  to know the problem when restarting the server. Issue #181 
Before:

image

So I have two possible solutions:
   1.- Show a button to activate output section inside VSCode (included in this PR)
image

   2.- Automatically activate outputChannel (NOT included in this PR):
i.e.
By default, in Azure section:
image

But after detecting the failure (Going to Output section and displaying the error message): 
image

@Kha and @lovettchris let me know your thoughts.

*UPDATE: This PR contains 2nd solution.*

Final solution : 

After checking CR comments, the output message stays like this:
image
Based on: https://github.com/leanprover/vscode-lean4/pull/229

### DIFF
--- a/vscode-lean4/src/utils/leanInstaller.ts
+++ b/vscode-lean4/src/utils/leanInstaller.ts
@@ -166,7 +166,10 @@ export class LeanInstaller implements Disposable {
         } else {
             const outputItem = 'Go to output';
             const outPrompt = 'See output window for more information';
-            window.showErrorMessage(outPrompt, outputItem)
+            const outItem = await window.showErrorMessage(outPrompt, outputItem)
+            if (outItem === outputItem) {
+                this.outputChannel.show(true);
+            }
         }
         const item = await window.showErrorMessage(prompt, installItem, selectItem)
         if (item === installItem) {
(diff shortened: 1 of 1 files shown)


lovettchris (approved):

Kha (approved):

vscode-lean4/src/utils/leanInstaller.ts
```diff
@@ -161,6 +161,7 @@ export class LeanInstaller implements Disposable {
         }
 
         this.promptingInstallEmitter.fire(uri);
+        this.outputChannel.show(true);
```
Vtec234, line 164: Are we certain this is the right change? It does not seem to introduce the 'Go to output' button.

Vtec234: Oh, I missed the update at the bottom of the PR description.

lovettchris: I'm also confused now, what happened to the show output button idea?  and show does this new outputChannel.show call relate to the new user configurable `lean4.autofocusOutput` option?  https://github.com/leanprover/vscode-lean4/pull/229

marianaalanis93:
After checking CR comments, the output message stays like this: 
image
Based on: https://github.com/leanprover/vscode-lean4/pull/229

vscode-lean4/src/utils/leanInstaller.ts
Kha: Start failure messages should always be of interest to the user, so I don't think #229 applies here. But an explicit button isn't that bad either.

Vtec234 (approved):

PR #231, opened 2022-07-28
fix: adhoc files should not pick up default lean version from open folder

### LINKED ISSUE #227: Untitled files no longer work
Using untitled files does not (longer?) seem to work with the current release of the extension.  I am a bit confused, because we have tests for this and they pass.  I have `elan` installed, the default Lean version is set to today's nightly, and I've opened a Lean 4 project.  Concretely:

1. Open `mathlib4`.
2. Ctrl-Shift-P "Create: New File..."
3. "Text File"
4. "Select a language"
5. "lean4"
6. No diagnostics are shown, and the infoview says "Waiting for Lean server to start..."

lovettchris:
fixes #227

### DIFF
--- a/vscode-lean4/src/utils/leanInstaller.ts
+++ b/vscode-lean4/src/utils/leanInstaller.ts
@@ -20,8 +20,8 @@ export class LeanInstaller implements Disposable {
     private localStorage: LocalStorageService;
     private subscriptions: Disposable[] = [];
     private prompting : boolean = false;
-    private defaultToolchain : string;
-    private elanDefaultToolchain : string = '';
+    private defaultToolchain : string; // the default to use if there is no elan installed
+    private elanDefaultToolchain : string = ''; // the default toolchain according to elan (toolchain marked with '(default)')
     private workspaceSuffix : string = '(workspace override)';
     private defaultSuffix : string = '(default)'
     private versionCache: Map<string,LeanVersion> = new Map();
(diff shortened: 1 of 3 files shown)


Vtec234:
Hm, could we ask Elan for what the default version is instead, and then use the 'select Lean toolchain' machinery to run that? I guess this will work but it's a bit complicated to make a temp folder just because the Lean version is selected based on `cwd`.

lovettchris:
> Hm, could we ask Elan for what the default version is instead, and then use the 'select Lean toolchain' machinery to run that? I guess this will work but it's a bit complicated to make a temp folder just because the Lean version is selected based on `cwd`.

ok, I'll try that instead.  I guess that also means if there is no default then you will get the mathlib version which is probably reasonable.

vscode-lean4/src/utils/leanInstaller.ts
```diff
@@ -21,6 +21,7 @@ export class LeanInstaller implements Disposable {
     private subscriptions: Disposable[] = [];
     private prompting : boolean = false;
     private defaultToolchain : string;
+    private elanDefaultToolchain : string = '';
```
Vtec234, line 24: I am a bit confused about the difference between `defaultToolchain` and `elanDefaultToolchain`. Do we need both? If so, could you document this field explaining why?

lovettchris: I added comments.

Vtec234:
Thanks!

PR #233, opened 2022-08-01
fix: issue 232: extension should work with hard coded lake path in the lean4/lakePath setting

### LINKED ISSUE #232: `checkLakeVersion` breaks when used with a Lake not in `PATH`
When determining whether to use Lake when no toolchain file is present, `checkLakeVersion` is called:

https://github.com/leanprover/vscode-lean4/blob/a154c97ef6541ec7d1b6b5f7480d684c627d54c0/vscode-lean4/src/leanclient.ts#L168

`checkLakeVersion` invokes Lake via `batchExecute`:

https://github.com/leanprover/vscode-lean4/blob/a154c97ef6541ec7d1b6b5f7480d684c627d54c0/vscode-lean4/src/leanclient.ts#L546

which requires that the executable specified be in `PATH`:

https://github.com/leanprover/vscode-lean4/blob/a154c97ef6541ec7d1b6b5f7480d684c627d54c0/vscode-lean4/src/utils/batch.ts#L25-L29

Thus, the invocation fails when the Lake has been manually specified -- either through a general toolchain path or a direct Lake path.

Based on the comment in `batchExecute`, the `PATH` check only appears to be for testing purposes, so one way to fix this issue would be to condition the check based on whether `batchExecute` is being used for testing. Alternatively, a different method could be used for invoking Lake.

lovettchris:
fixes #232

### DIFF
--- a/vscode-lean4/src/utils/batch.ts
+++ b/vscode-lean4/src/utils/batch.ts
@@ -1,6 +1,6 @@
 import { OutputChannel } from 'vscode'
 import { spawn } from 'child_process';
-import { findProgramInPath } from '../config'
+import { findProgramInPath, isRunningTest } from '../config'
 import { logger } from './logger'
 
 export async function batchExecute(
@@ -16,16 +16,18 @@ export async function batchExecute(
             options = { cwd: workingDirectory };
         }
 
-        // The mocha test framework listens to process.on('uncaughtException')
-        // which is raised if spawn cannot find the command and the test automatically
-        // fails with "Uncaught Error: spawn elan ENOENT".  Therefore we manually
-        // check if the command exists so as not to trigger that exception.
-
         try {
-            const fullPath = findProgramInPath(executablePath);
-            if (!fullPath) {
-                resolve(undefined);
-                return;
+            if (isRunningTest())
+            {
+                // The mocha test framework listens to process.on('uncaughtException')
+                // which is raised if spawn cannot find the command and the test automatically
+                // fails with "Uncaught Error: spawn elan ENOENT".  Therefore we manually
+                // check if the command exists so as not to trigger that exception.
+                const fullPath = findProgramInPath(executablePath);
+                if (!fullPath) {
+                    resolve(undefined);
+                    return;
+                }
             }
             const proc = spawn(executablePath, args, options);
 


Vtec234 (commented):
Afaik any kind of "check if running in a test" outside of test code is an antipattern because at that point you are no longer testing the real program, just some test-specific version of it. But I'm not too familiar with what is happening here, so if you think this is necessary then LGTM.

lovettchris:
> Afaik any kind of "check if running in a test" outside of test code is an antipattern because at that point you are no longer testing the real program, just some test-specific version of it. But I'm not too familiar with what is happening here, so if you think this is necessary then LGTM.

Agreed, I'm trying to keep it to a minimum.  The problem is mocha is changing the runtime behavior in unexpected ways which is what this code is dealing with.

vscode-lean4/src/utils/batch.ts
```diff
@@ -16,16 +16,18 @@ export async function batchExecute(
...
-        // check if the command exists so as not to trigger that exception.
-
         try {
-            const fullPath = findProgramInPath(executablePath);
-            if (!fullPath) {
-                resolve(undefined);
-                return;
+            if (isRunningTest())
+            {
+                // The mocha test framework listens to process.on('uncaughtException')
```
Vtec234, line 22: I understand now. So the question is, could we turn this behaviour off in Mocha and delete the whole `findProgramInPath` check? If not, let's go with the `isRunningTest` check.

Vtec234: Hm, but why is the exception raised by `spawn` uncaught? Maybe that's the problem in the first place?

lovettchris: I don't think so, the mocha code that registers `uncaughtException` is buried pretty deep in mocha, and I couldn't find a way to disable it or undo it.

lovettchris: > Hm, but why is the exception raised by `spawn` uncaught? Maybe that's the problem in the first place?

yeah, it's weird I know, we have a try/catch which catches it, but spawn still raises it as a process.uncaughtException for some reason, and this is then causing mocha to fail the test.  It is all very weird.

lovettchris: PS: this code was already there by the way, I'm just making it test only code now.

Issue #239, opened 2022-08-09
Confusing error messages when importing a file with errors


PatrickMassot:
Steps to reproduce:
* mkdir /tmp/issue && cd /tmp/issue && lake init Issue
* echo "This is bad" > Issue.lean
* Open `Main.lean` and go inside the main function.

The info view shows `Error updating: Error fetching goals: Rpc error: InvalidParams: Incorrect position 'file:///tmp/issue/Main.lean:3:1' in RPC call. Try again.`.

Then put your cursor on the import line. The expected error message is something like "File Issue.lean contains errors" but you get
```
`/home/pmassot/.elan/toolchains/leanprover--lean4---nightly-2022-08-06/bin/lake print-paths Init Issue` failed:

stderr:
info: > LEAN_PATH=./build/lib LD_LIBRARY_PATH=/home/pmassot/.elan/toolchains/leanprover--lean4---nightly-2022-08-06/lib:./build/lib:./build/lib /home/pmassot/.elan/toolchains/leanprover--lean4---nightly-2022-08-06/bin/lean ./././Issue.lean -R ././. -o ./build/lib/Issue.olean -i ./build/lib/Issue.ilean -c ./build/ir/Issue.c
error: stdout:
./././Issue.lean:1:0: error: expected command
error: external command /home/pmassot/.elan/toolchains/leanprover--lean4---nightly-2022-08-06/bin/lean exited with status 1
error: build failed
```

gebner:
I cannot reproduce the infoview issue.  For me it shows the second message, which says that `Issue.lean` has an error.  What Lean version is this?

I think we all agree that the second message is not entirely optimal.  See also https://github.com/leanprover/lean4/issues/1438

BTW, you should *never* use `lake init` or `lake new` without a version specifier.  Always call it with `lake +leanprover/lean4:nightly-2022-08-09 new issue`, particularly when reporting bugs (so that we both get the same version).

PatrickMassot:
Here is a video showing how I get the issue you can't reproduce (and specifying the lean version).
![bug](https://user-images.githubusercontent.com/14060883/183760800-888962cc-6f25-4661-9507-ce674bac1260.gif)

marianaalanis93:
@gebner and @PatrickMassot I'll start checking this!

lovettchris:
Note: to get the rpc error you have to click somewhere else in Main.lean.  Then you get this:

```
Main.lean:4:32
[...]

PR #242, opened 2022-08-11
fix: Confusing error messages when importing a file with errors

marianaalanis93:
Fix for issue #239 Confusing error messages when importing a file with errors.

This fix includes an extra rule for RPC errors, they can be confusing to the user (such as outdated session or Rpc error: InvalidParams: Incorrect position (for bad or problem imports).

The idea is just to show the error that is displayed in the messages part as seen below:

image


marianaalanis93:
Returning this to draft until we confirm if it's a bug from RPC

PR #244, opened 2022-08-15
fix: Confusing error messages when importing a file with errors

marianaalanis93:
Regression error fix. 

Fix for issue https://github.com/leanprover/vscode-lean4/issues/239 Confusing error messages when importing a file with errors.

This fix includes an extra rule for RPC errors, they can be confusing to the user (such as outdated session or Rpc error: InvalidParams: Incorrect position (for bad or problem imports).

The error IS an Invalid Params error (because we are passing invalid params from and import) just we need to re-categorize it in this case and avoid confusing messages (since it can be confused as an invalid position of the lean file)

- B.lean:
image

- A.lean:
image

**Before:**
image

**After**

image

### DIFF
--- a/lean4-infoview/src/infoview/info.tsx
+++ b/lean4-infoview/src/infoview/info.tsx
@@ -140,7 +140,6 @@ export function InfoDisplay(props0: InfoDisplayProps) {
     const hasGoals = status !== 'error' && goals;
     const hasTermGoal = status !== 'error' && termGoal;
     const hasMessages = status !== 'error' && messages.length !== 0;
-    const isRpcErr = error && error.includes('Rpc error')
 
     const sortClasses = 'link pointer mh2 dim codicon fr ' + (goalFilters.reverse ? 'codicon-arrow-up ' : 'codicon-arrow-down ');
     const sortButton = <a className={sortClasses} title="reverse list" onClick={e => {
@@ -184,7 +183,7 @@ export function InfoDisplay(props0: InfoDisplayProps) {
     <Details initiallyOpen>
         <InfoStatusBar {...props} triggerUpdate={triggerDisplayUpdate} isPaused={isPaused} setPaused={setPaused} copyGoalToComment={copyGoalToComment} />
         <div className="ml1">
-            {hasError && !isRpcErr &&
+            {hasError &&
                 <div className="error" key="errors">
                     Error updating:{' '}{error}.
                     <a className="link pointer dim" onClick={e => { e.preventDefault(); void triggerDisplayUpdate(); }}>{' '}Try again.</a>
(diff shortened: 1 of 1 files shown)


lean4-infoview/src/infoview/info.tsx
```diff
@@ -375,26 +376,37 @@ function InfoAux(props: InfoProps) {
...
-                // unrecognised error
-                errorString = `Unrecognised error: ${JSON.stringify(ex)}`
+            if (isRpcError(ex) && ex.code === RpcErrorCode.InvalidParams) {
+                // Check if the position is valid, if yes just update error message
+                if(pos.line !== undefined || null && pos.line !== undefined || null)
+                {
+                    errorString = 'Invalid Lean code';
+                } else {
+                    errorString = mapRpcError(ex).message
+                }
```
lovettchris, line 392: I think this could instead be moved to mapRpcError in Util.ts, then this file would not need to change.

lovettchris (changes requested):

lean4-infoview/src/infoview/info.tsx
marianaalanis93: Ok, good point!

gebner:
Sorry I didn't have time to look into #239 until now.  If I understand it correctly, this PR just changes the error message to "Invalid Lean code".  Aside from the question whether this is best done in the editor extension, I don't think the text "Invalid Lean code" would have helped Patrick understand what's going on.  I don't think that the current approach taken in this PR is the best one.

What's happening under the hood is the following:
 1. The error messages are fetched with a custom RPC protocol that is layered over LSP.
 2. Each RPC request contains a position in the file.  This is not the position used to find the error messages.  Instead this position is used to find extra, custom RPC handlers that the user can add in the same file.  Every RPC handler above the position can be called.
 3. When the server (the file worker) receives an RPC request, it has to wait until the file has processed until that position (so that it can collect the custom RPC handlers).  This step breaks when processing stops in the middle of the file.

Another example where Lean stops in the middle of a file is the `#exit` command:
```lean
#exit
#eval 0 --< put cursor here to get the same RPC error
```

I see two reasonable solutions to #239.
 1. We change the server to pick the final snapshot if Lean stops in the middle of a file.
 2. We do away with the extra position argument in RPC requests.  Custom RPC requests can still be declared in imported modules, so it's no worse than environment extensions or attributes, etc.  Removing the extra position argument has quite a few advantages:
   - It fixes this issue.
   - RPC requests don't block unnecessarily.  For example if you pin an info from a position further down the file, then all tooltips etc. will be paused while you're editing (and after that until processing has caught up with the second position again).
   - It removes some inconsistent states.  For example, say you pin a widget that calls an RPC function in the current file.  If you edit the RPC function, the old widget will call the new function (potentially type-unsafe, but not memory unsafe).
   - It simplifies the RPC protocol and server implementation a bit.
 We can implement this in a backwards compatible way by just having the extension send `position: {line: 0, character: 0}` for every RPC request.

I'm kind of preferring (2).  I've often run into blocking RPC requests.  So far, I've thought that's only because of the edit grouping in the watchdog, but if the worker waits for (a part of) the file to compile, then that's another significant delay.  @Vtec234 what do you think?

lovettchris:
I'd love to copy into some real developer design documentation.  I suspected this was actually an LSP issue.,  I would prefer to fix it properly in the LSP.  I'm confused about this statement: 

> Each RPC request contains a position in the file. This is not the position used to find the error messages. 

Which of these calls that take a position do not need an valid position?
image
Presumably the getInteractiveGoals does need a valid position.

gebner:
> Which of these calls that take a position do not need an valid position?

All of them (take a position but would get `(0,0)` with the proposal).  We have a nice API that encapsulates this position business, so the actual RPC calls never even see it: https://github.com/leanprover/vscode-lean4/blob/89b37d3f4c5df30ae51ba4aa7e45df366df4a686/lean4-infoview-api/src/rpcSessions.ts#L27-L45

There's 3 classes here:
 - `RpcSessions` (singleton)
 - `RpcSessionForFile` (bundles the file name)
 - `RpcSessionAtPos` (bundles the file name and position)

Eliminating the position argument is mostly a question of replacing `TextDocumentPositionParams` by `DocumentUri` and removing the extra `RpcSessionAtPos` layer in this file.

gebner:
> Presumably the getInteractiveGoals does need a valid position.

At the moment it gets *two* position arguments: one to look up the custom RPC functions, and one that tells it where to look for the goals.  My proposal would remove the first position but keep the second one.

lovettchris:
Oh, I see now, thanks, it is coming from this RpcSessionAtPos instance:

```typescript
    at(pos: TextDocumentPositionParams): RpcSessionAtPos {
        return { call: (method, params) => this.call(pos, method, params) };
    }
```
Definitely need Wojciech's input on this one... presumably there's a reason he wanted that position info, what if we change LSP instead to be less fussy about the position so it never reports that as an invalid parameters, especially when the position is not invalid according to the current contents of the file...

marianaalanis93:
> Oh, I see now, thanks, it is coming from this RpcSessionAtPos instance:
> 
> ```ts
>     at(pos: TextDocumentPositionParams): RpcSessionAtPos {
>         return { call: (method, params) => this.call(pos, method, params) };
>     }
> ```
> 
> Definitely need Wojciech's input on this one... presumably there's a reason he wanted that position info, what if we change LSP instead to be less fussy about the position so it never reports that as an invalid parameters, especially when the position is not invalid according to the current contents of the file...

Totally agree with checking @Vtec234 input for this one.

Vtec234:
Hi, sorry for the delay. I am currently on holiday but will be back on Sept 3 and can give more detailed comments then. Basically the position argument is so that, as Gabriel noted, we can call RPC handlers in the current file. This is quite useful for experimentation because we can see changes to the handler reflected in real-time in the infoview when a widget using that handler is pinned.

> RPC requests don't block unnecessarily. For example if you pin an info from a position further down the file, then all tooltips etc. will be paused while you're editing (and after that until processing has caught up with the second position again).

We have the invariant that rpc handler names are unique and cannot be overriden so we could fix this simply by having the server look up the handler in the imported environment and return immediately if the handler is found there. Otherwise we block and wait for the snapshot.

PR #246, opened 2022-08-17
fix: Switch Windows script download to use BITS

Mortein:


Mortein:
We use some Windows Server 2016 systems, which default to SSL and TLS/1.0 for SecurityProtocol in PowerShell. `Invoke-WebRequest` fails, as GitHub is using TLS/1.2.

The BITS client doesn't appear to care about the defined security protocols, and downloads the file successfully.

Mortein:
Alternative is to change the protocols for the session to TLS/1.2 (using the below), but this will stop working when TLS/1.2 is eventually replaced.
`[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12`

Mortein:
I've tested the download using `Start-BitsTransfer` on Windows 10, Server 2016, and Server 2019.

lovettchris (approved):
Looks good, works on my Windows 11 box.  I guess I didn't run into this because I use the new `pwsh` not the older Powershell.

PR #247, opened 2022-08-24
fix: Error loading webview: Error: Could not register service workers InvalidStateError

marianaalanis93:
Error loading webview: Error: Could not register service workers InvalidStateError


lovettchris:
I'm hoping we can fix the vscode extension in a way that we don't have to do this.

PR #249, opened 2022-08-30
feat: bring back inaccessible names highlighting

Kha:
As discussed in https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/De-emphasizing.20inaccessible.20hypothesis.20names

image

image

image

### DIFF
--- a/lean4-infoview/src/infoview/index.css
+++ b/lean4-infoview/src/infoview/index.css
@@ -23,13 +23,17 @@ html,body {
 .goal-vdash { color: #569cd6; }
 .goal-case { color: #a1df90; }
 .goal-hyp { color: #ffcc00; }
-.goal-inaccessible { color: #808080 }
+.goal-inaccessible {
+    color: var(--vscode-editor-foreground);
+    opacity: 0.7;
+    font-style: italic;
+    font-weight: normal;
+}
 
 .vscode-light .goal-goals { color: #367cb6; }
 .vscode-light .goal-vdash { color: #367cb6; }
 .vscode-light .goal-case { color: #1f7a1f; }
 .vscode-light .goal-hyp { color: #cc7a00; }
-.vscode-light .goal-inaccessible { color: #777; }
 
 /* Used to denote text highlighted by the pretty-print expression widget. */
 .highlightable {
--- a/lean4-infoview/src/infoview/goals.tsx
+++ b/lean4-infoview/src/infoview/goals.tsx
@@ -7,9 +7,15 @@ interface HypProps {
     hyp: InteractiveHypothesisBundle
 }
 
+
+/** Returns true if `h` is inaccessible according to Lean's default name rendering. */
+function isInaccessibleName(h: string): boolean {
+    return h.indexOf('✝') >= 0;
+}
+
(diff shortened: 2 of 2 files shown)


lean4-infoview/src/infoview/interactiveCode.tsx
```diff
@@ -25,8 +25,12 @@ export interface InteractiveTaggedTextProps<T> extends InteractiveTextComponentP
  * various embedded information such as `InfoTree`s and `Expr`s.
  * */
 export function InteractiveTaggedText<T>({fmt, InnerTagUi}: InteractiveTaggedTextProps<T>) {
-  if ('text' in fmt) return <>{fmt.text}</>
-  else if ('append' in fmt) return <>
+  if ('text' in fmt) {
+    if (isInaccessibleName(fmt.text))
+      return <span className="goal-inaccessible">{fmt.text}</span>
+    else
```
gebner, line 31:
```suggestion
```
This check is super flaky and will trigger on any text containing `✝`.  (Including `logInfo "this will all be gray ✝ not tested though"` for example.)

lean4-infoview/src/infoview/index.css
```diff
@@ -23,13 +23,17 @@ html,body {
...
+    font-style: italic;
+    font-weight: normal;
+}
 
 .vscode-light .goal-goals { color: #367cb6; }
 .vscode-light .goal-vdash { color: #367cb6; }
 .vscode-light .goal-case { color: #1f7a1f; }
 .vscode-light .goal-hyp { color: #cc7a00; }
-.vscode-light .goal-inaccessible { color: #777; }
+.vscode-light .goal-inaccessible { color: #aaa; }
```
gebner, line 36: This color choice was intentional because otherwise it's hard to read.  https://github.com/leanprover/vscode-lean4/commit/921151063ba40bda31490011829245fb33139737

gebner:
> Yes, I did make the font color a bit lighter for light themes to further deemphasize it

For me it actually has the opposite effect (i.e., emphasize instead of deemphasize).  It's the only thing that has a different color in the expression, so it looks special.

The hypothesis name looks deemphasized though (because the regular hyps are bold and orange).

lean4-infoview/src/infoview/interactiveCode.tsx
Kha: I guess that's technically true... I don't know if this part is important enough to introduce a more robust server-side solution

Kha:
It sure is interesting how different individual perception can be, for me it makes me immediately gloss over those parts of the expression. Anyway, this is the output with both changes reverted.
image
I think I mostly disliked the original color because it has more contrast than the default font color - but only in Solarized Light, all other default themes look fine. So it's ok I guess.

gebner:
> but only in Solarized Light, all other default themes look fine

Ah, that's where you're coming from!  Maybe we should do `color: rgba(var(--vscode-editor-foreground), 0.7)` or `opacity: 0.7` then.

Kha:
Yeah, looks pretty good to me! I've updated the screenshots.

PR #251, opened 2022-09-16
feat: goal-diffs

EdAyers:
See https://github.com/leanprover/lean4/pull/1610


gebner (changes requested):
This seems to be completely missing the corresponding RPC API definitions.  No idea why this even builds on windows.

EdAyers:
Now that lean4#1666 is in this is ready for a review.

I've also changed `.highlight` CSS class to now use `var(--vscode-editor-selectionBackground)` rather than a hardcoded value so it looks a bit better on non-standard themes.

lean4-infoview/src/infoview/goals.tsx
```diff
@@ -7,16 +7,21 @@ interface HypProps {
...
+    let namecls : string = ''
+    if (h.isInserted) {
+        namecls += 'inserted-text '
+    } else if (h.isRemoved) {
+        namecls += 'removed-text '
+    }
     const names = InteractiveHypothesisBundle_accessibleNames(h).map((n, i) =>
-            <span className={'mr1 ' + (isInaccessibleName(n) ? 'goal-inaccessible' : '')} key={i}>{n}</span>
-        )
+        <span className={ 'mr1 ' + (isInaccessibleName(n) ? 'goal-inaccessible ' : '') + namecls} key={i}>{n}</span>
```
Vtec234, line 23: Minor, but you could default to `namecls : string = 'mr1 '` to shorten this line a bit.

lean4-infoview/src/infoview/interactiveCode.tsx
```diff
@@ -85,12 +85,31 @@ function TypePopupContents({ info, redrawTooltip }: TypePopupContentsProps) {
...
+const DIFF_TAG_TO_CLASS : {[K in DiffTag] : string} = {
+  'wasChanged': 'inserted-text',
+  'willChange': 'removed-text',
+  'wasInserted': 'inserted-text',
+  'willInsert': 'inserted-text',
+  'willDelete': 'removed-text',
+  'wasDeleted': 'removed-text',
+}
+
+const DIFF_TAG_TO_EXPLANATION : {[K in DiffTag] : string} = {
```
Vtec234, line 104: This is not used. Did you mean to put it in as an ARIA hint?

Vtec234 (changes requested):
It looks great! My only change request would be to add a way to toggle the diff view, as it seems like a personal preference, and even then proof-specific, whether it should be shown. Maybe we could have a button right next to 'filters' and 'reverse list'?

Vtec234:
Hm, hold your horses. I have been testing this today and am observing a serious slowdown when typing in expressions into a proof. With the current release, I can type a `let` statement in a proof and, while the state updates to something meaningless during the process, after I finish typing it is instantly the right goal. With this branch, it takes something like 5s to go from processing (yellow filename) to showing the goal state after I finish typing. I am assuming this is a client-side problem because the diffs are computed in the server regardless of whether we show them (right?). This is on `nightly-2022-10-09`. It might be an issue with our general lack of goalview-update debouncing (we do not wait to fetch the goal state until the user is finished typing or more generally finished moving the cursor), but even if so, it is worrying if a single goal state draw has become significantly more expensive.

EdAyers:
That's a little worrying, I don't know what could be causing an overhead in client which makes me worry it's on the server.

The diffs are computed when they are requested by the client, at the same time the goals are converted to `InteractiveGoals`.

Vtec234:
@EdAyers okay, so this turned out to be #257 which is observable on a debug build of the current release as well. It is orthogonal to this PR, so not a blocker. Sorry for worrying unnecessarily!

gebner:
@Vtec234 Do you have a repro for the slowdown?  I see absolutely nothing in this PR that would explain that.  We don't change any of the requests to the Lean server, or trigger any additional updates, etc. (AFAICT)

The color scheme is a bit strong (particularly the red background is really in-your-face).  I expect people to get tired of it pretty quickly, but we can tone it down iteratively once it's deployed.

With this PR I can see quite a few bugs (or unexpected behavior) in the server implementation.  For example you get almost opposite highlighting results if you have your cursor on the first character of the tactic vs. on the space before it (which used to be the same).

gebner:
Are we inheriting some animation setting from the popup highlights?  Maybe my eyes are deceiving me, but I see a tiny fade-in...

Vtec234:
> @Vtec234 Do you have a repro for the slowdown?

Please see the linked issue. I was mistaken, this is not a problem with this PR, it is also present in the current extension.

Vtec234:
Hey @gebner, did you mean to merge? We should definitely add the on/off toggle for diffs before a new release at least.

gebner:
My plan was to wait for the inevitable user feedback first. 😄  But yeah, we should probably at least add a config option for the diff highlighting.

PR #262, opened 2022-10-24
avoid returning promises in `useEffect`

abentkamp:
I am working on a web version of Lean 4 and ran into the issue that some of the `useEffect` functions accidentally return a promise. The issue is described in some detail here: https://medium.com/geekculture/react-uncaught-typeerror-destroy-is-not-a-function-192738a6e79b. I am not sure why this is not an issue in the VSCode webview, but it is an issue in the browser.


gebner:
Thank you!  I wonder if this is an issue with gitpod/codespaces as well..

PR #263, opened 2022-10-27
update lakefile.lean defaultTarget and improve test reliability

lovettchris:
fix: update lakefile.lean defaultTarget.
fix: attempt to make bootstrap more reliable.
fix: windows bootstrap so workflow no longer needs to install Lean.


vscode-lean4/src/leanclient.ts
```diff
@@ -478,7 +480,10 @@ export class LeanClient implements Disposable {
     configChanged(e : ConfigurationChangeEvent): void {
         let newToolchainPath = this.storageManager.getLeanPath();
         if (!newToolchainPath) newToolchainPath = toolchainPath();
-        if (this.toolchainPath !== newToolchainPath){
+        if (this.toolchainPath === undefined){
+            this.toolchainPath = newToolchainPath;
```
Vtec234, line 484: Could you put a comment explaining why we don't need to restart in this case?

vscode-lean4/test/suite/bootstrap/bootstrap.test.ts
```diff
@@ -17,19 +13,27 @@ suite('Lean4 Bootstrap Test Suite', () => {
...
-        // and if it times out at 300 seconds then waitForInfoviewHtml prints the contents of the InfoView so we can see what happened.
-		await waitForInfoviewHtml(info, '4.0.0-nightly-', 5, 60000, false, () => restartLeanServer(client));
+        await waitForActiveClient(lean.exports.clientProvider, 120);
+        await waitForActiveClientRunning(lean.exports.clientProvider, 300);
+
+        // if it times out at 300 seconds then waitForInfoviewHtml prints the contents of the InfoView so we can see what happened.
+		await waitForInfoviewHtml(info, expected, 10, 10000, true, async () => {
+            // 60 seconds elapsed, and infoview is not updating, try and re-edit
+            // the file to force the LSP to update.
+            await deleteAllText();
```
Vtec234, line 34: I am not sure about this part. I feel like "reedit if the infoview is stuck" is a workaround that users may choose for an infoview-not-updating bug, but our tests should not tolerate such bugs.

vscode-lean4/src/config.ts
```diff
@@ -201,6 +201,10 @@ export function getTestFolder() : string {
     return typeof(process.env.TEST_FOLDER) === 'string' ? process.env.TEST_FOLDER : '';
 }
 
+export function getTestVersion(): string | undefined {
```
Vtec234, line 204: Should we just call this `DEFAULT_LEAN_TOOLCHAIN`? We do use it for tests but the implemented behaviour is to set `defaultToolchain` *in general*.

.github/workflows/on-push.yml
```diff
@@ -59,12 +59,6 @@ jobs:
         if: matrix.os == 'windows-latest'
         run: |
           echo "$HOME\.elan\bin" >> $env:GITHUB_PATH
-      - name: Setup elan toolchain on Windows
```
lovettchris, line 62: this was a hack since windows bootstrap test was not reliable, but this hack is no longer needed.

Vtec234 (changes requested):
I am not familiar enough with the test code to review it in detail but the changes seem sensible. My one comment is that I am not sure about the "try again" part I pointed out because while it is something users may do to move with their work, I feel like the tests should simulate a "very strict" user who complains and exits the moment something goes wrong rather than being more tolerant, at least for workflows which should not go wrong (obviously we must tolerate network failure and things like that). The tests will catch more issues this way.

vscode-lean4/src/utils/leanInstaller.ts
```diff
@@ -62,9 +58,6 @@ export class LeanInstaller implements Disposable {
                 // Ah, there is no elan, but what if Lean is in the PATH due to custom install?
                 const found = await this.checkLeanVersion(packageUri, leanVersion);
                 if (found.error) {
-                    // Ah, then we need to install elan so prompt the user, but don't wait
-                    // here for the answer because that could be forever.
-                    void this.showInstallOptions(packageUri);
```
lovettchris, line 67: not waiting on a window.showErrorMessage is bogus and should never be allowed, because it means the popup window can randomly disappear such that the user might not ever see it.

vscode-lean4/src/utils/leanInstaller.ts
```diff
@@ -161,24 +156,18 @@ export class LeanInstaller implements Disposable {
             prompt += ` from ${path}`
         }
 
-        this.promptingInstallEmitter.fire(uri);
         if (shouldAutofocusOutput()) {
             this.outputChannel.show(true);
-        } else {
-            const outputItem = 'Go to output';
```
lovettchris, line 168: this block was removed because it pops up an error message and blocks - this stops the real show install prompt message below on line 175, which we really don't want to block, that prompt takes precedence here.

vscode-lean4/test/suite/bootstrap/bootstrap.test.ts
```diff
@@ -70,43 +74,4 @@ suite('Lean4 Bootstrap Test Suite', () => {
 
     }).timeout(600000);
 
-    test('Create linked toolchain named master', async () => {
```
lovettchris, line 73: this test was always a bit bogus because it is literally copying a known good install, this is very different from actually using a locally build binaries from the lean4 repo, unfortunately testing that using a github artifact turned out to be unreliable, so we'll leave that one to manual testing for now.

vscode-lean4/test/suite/bootstrap/bootstrap.test.ts
lovettchris: Sure I'll comment it out, if tests remain reliable we can leave it out, but we also need reliable tests, so we shall see, sounds good?

Vtec234: If this breaks the tests, either something in the test's assumptions is wrong or there is another extension bug to fix.

vscode-lean4/src/leanclient.ts
lovettchris: I pushed a different fix, I was seeing vs raise the configuration change before anything else, before restart had ever been called once, so this.toolchainPath was uninitialized, but I pushed a change now that initializes it in the constructor which is a better way to solve this problem.

vscode-lean4/src/config.ts
lovettchris: Good point, I renamed it getDefaultLeanVersion and moved the default version here also.

vscode-lean4/src/config.ts
```diff
@@ -201,8 +201,8 @@ export function getTestFolder() : string {
     return typeof(process.env.TEST_FOLDER) === 'string' ? process.env.TEST_FOLDER : '';
 }
 
-export function getTestVersion(): string | undefined {
-    return typeof(process.env.TEST_VERSION) === 'string' ? process.env.TEST_VERSION : undefined;
+export function getDefaultLeanVersion(): string {
+    return typeof(process.env.TEST_VERSION) === 'string' ? process.env.TEST_VERSION : 'leanprover/lean4:nightly';
```
Vtec234, line 205: Let's also rename the environment var.

lovettchris: Oh, yeah, I see what you mean now, good idea, done. See 2f438a787d770623b749f840e0e2e3400fd56094

lovettchris:
Argh, failed on the 4th re-run !! 

image

This time in `=================== Test worker crashed and client running` ... harrumph, sometimes I think testing is harder than building the product in the first place.

lovettchris:
Ok, @Vtec234 5 greens, I think it's good to go now:
image

Vtec234:
Okay, the tests pass locally for me as well.

PR #267, opened 2022-11-15
feat: update infoview to React 18

### LINKED ISSUE #234: get es-module-shims from NPM
See discussion under https://github.com/leanprover/vscode-lean4/pull/167#issuecomment-1171741061,

specifically:

> I don't remember why I wrote 'it's not on NPM' even though it actually is. I may have just missed it. Note that the [version](https://ga.jspm.io/npm:es-module-shims@1.5.8/dist/es-module-shims.js) recommended in their README is different than [the NPM version](https://unpkg.com/es-module-shims@1.5.8/dist/es-module-shims.js), but it might just be minified. If you can get the NPM one working, it would be better. Since the point is to make Gitpod work, that would be the thing to test.

Vtec234:
- Update to React 18. We needed to do this eventually as more and more packages break compatibility with React 17 which causes issues when building widgets; so I took the plunge now.
- Update Rollup to v3.
- Drop dependency on external ES module bundlers and bundle `react/react-dom/react-popper` as part of the `lean4-infoview` build. It seems literally impossible to do this any other way, see the comment in `rollup.config.js` (not that this is really an issue, just a bit of a pain that we have to bundle manually).
- Closes #234.


gebner:
I think I was too quick with merging.  CI says it builds, but when I hit F5 locally I get:
```
Uncaught Error Error: 404  https://file+.vscode-resource.vscode-cdn.net/home/gebner/vscode-lean4/vscode-lean4/dist/lean4-infoview/react.development.js imported from https://file+.vscode-resource.vscode-cdn.net/home/gebner/vscode-lean4/vscode-lean4/dist/lean4-infoview/index.development.js
```

Vtec234:
@gebner it may simply be a matter of rerunning `npm run watch`. The build script doesn't quite keep track of the dependencies across the packages in the monorepo so sometimes it needs to be run twice before changes propagate. I just did a fresh clone, ran that, and F5 worked.

PR #272, opened 2022-11-28
fix: RPC race condition

Vtec234:
It looks like we had a long-standing bug in the infoview RPC channel wherein method calls could start coming in from the other side before we get a chance to call `rpc.register`. This became much more observable when I messed with some other code but it's a potential source of some occurrences of the `Waiting for Lean server to start..` bug.


Vtec234:
This may actually make the whole `initialize`/`initialized` sending redundant, but it doesn't hurt to impose some extra ordering.

Issue #79, opened 2021-11-25
Local toolchain setting is intransparent and can't be reset


gebner:
So I've just tried to debug why the Lean 4 extension used an old Lean 4 nightly for my Lean 3 checkout.  Looking through the code, either localStorage.getLeanPath or storage.getLeanVersion must return some override, but I can't tell which.

This is honestly really bad and confusing.  If this happened in a Lean 4 project then I would've gone crazy if a file broke in the editor but worked just fine on the command-line with no indication that the editor used some random lean version.

With vscode settings, I can look in the settings tab for the current setting (and it shows the value that I want.....).  With elan overrides, I can see what's going on in the command line (I can also query the override with `elan override list`).  But with these local storage settings, I have no idea what they're set to.  They're stored in a sqlite database with the wonderfully descriptive name `fdbe00b2e2eb8f3841860e8a590c4788/state.vscdb` (where of course the json is stored in a string column.....), with no good way to view the contents.

If we want to keep the toolchain selection feature, then the chosen version needs to be transparent to the user and it must be possible to reset the override:
1. There should be a status bar element (like in the python extension) that shows the effective lean version and executable path.
2. The configuration setting description (lean.executablePath) should mention that it might be overridden.
3. The version selection command must be available even if the Lean 3 extension is active.
4. It should be possible to reset the override.

Vtec234:
Why don't we just use a standard VSCode setting for the toolchain version? Local storage is useful for persisting internal objects, but it doesn't sound appropriate for user-controllable settings.

lovettchris:
I agree we need to make it more clear when you have chosen something in Select Toolchain (this is why originally I wanted to edit your lean-toolchain file to make it very explicit what you just did.  I stored it in the Workspace settings because you didn't want me to edit the lean-toolchain file.

tydeu:
> But note that a Workspace settings ONLY applies to that work space, if you opened a different Lean 3 folder it will NOT apply the result of your previous Select Toolchain choice because it is a different workspace.

I would think this would be the desired behavior? Support for different toolchain settings for different workspaces seems ideal.

lovettchris:
Here's my proposed fix for this issue.  First, the PR# https://github.com/leanprover/vscode-lean4/pull/83 makes select toolchain less common by setting a `(default)` toolchain when there is no elan installed.  

Then I added the following UI toi show "(workspace override)" on the toolchain that is currently overriding the version in the current workspace:

[...]

PR #284, opened 2023-02-04
chore: remove select toolchain command

gebner:
The select toolchain command is an extremely confusing feature for users.  It allows you to set a different version in vscode than on the command line, which is almost never what you want.  This mismatch also makes it hard to debug the error.  Additionally the current state of the vscode toolchain override is not visible, not even an experienced user can tell if anything's wrong.  The only way to debug it is `#eval Lean.versionString`.

Today, [`Thomas ran into this issue`](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.E2.9C.94.20.60lake.20build.60.20builds.20independently.20of.20lean4.20VS.20code.20exte.2E.2E.2E/near/325744138) believing that he did everything right (even selecting the "default" toolchain!).  He won't be the last one to run into this.  Since the feature is causing issues, I'm going to remove it.


gebner:
See also #79.

Issue #210, opened 2022-07-06
InfoView popups need scrollability


lovettchris:
Some large markdown comment blocks require scrolling like this, which you will see in the editor tips:

![image](https://user-images.githubusercontent.com/18707114/177438498-68341248-dd7f-46e5-b613-c3c77c8ed312.png)

Today the infoview will render large popups off the bottom or off the top of the infoview:

![image](https://user-images.githubusercontent.com/18707114/177457743-14282443-5859-4385-95e8-8a6f85940368.png)

or

![image](https://user-images.githubusercontent.com/18707114/177438861-188925f2-b63a-4432-8693-44a988b25183.png)

and the vertical size will be infinite with no nested scrollbar inside the tip.  I think a fixed max vertical height  with scrollbars when needed would be nicer especially because of the flip() limitations in Popper v2.x mentioned in https://github.com/leanprover/vscode-lean4/pull/207

lovettchris:
The flip logic is completely lacking any concept of "best" orientation based on "minimal offscreen" amounts. Instead it is a binary switch, if "bottom" results in any amount off the bottom then don't use it which is a not the best algorithm for large hover tips, and turns out off the bottom in the InfoView is fine because the InfoView itself is scrollable, so it's easy to read the hover tip with some vertical scrolling, off the top is much worse because the InfoView scrollbar won't go negative.

The flip() implementation in Popper v2 is also impossible to replace as far as I can tell, but I've noticed Popper seems to be replaced now by https://github.com/floating-ui/floating-ui which looks much better because it gives you much more control over the middleware (e.g. flip function etc). It also claims to be smaller and lighter so we should be able to implement a more InfoView friendly flip algorithm very easily.

Issue #281, opened 2023-01-20
Popups for huge expressions are shown outside of viewport


gebner:
```lean
def N : Nat := Nat.add 37 50

example : N ≠ N := by
  simp (config := {decide := false}) only [N, Nat.add]
  /- Put cursor on `≠` in infoview -/
  sorry
```

When highlighting a subterm that is multiple pages long, the popup gets shown at the beginning of the subterm, which is outside of the viewport.

![image](https://user-images.githubusercontent.com/313929/213806861-18d1a068-502d-4a1c-8656-6175581b680c.png)

PR #289, opened 2023-03-14
fix: Ensure large tooltip popups are not clipped and stay within the inforview viewport

### LINKED ISSUE #280: Large popups are clipped
```lean
def N : Nat := Nat.add 37 50

example : N ≠ N := by
  simp (config := {decide := false}) only [N, Nat.add]
  /- Put cursor on first `Nat.succ` in infoview -/
  sorry
```

The popup will start above and outside of the page, so you can't see where it begins.

image

tnaoi:
# Summary
This fix resolves #280, and #210.

# Changes
- Replaced popper 2 with floating-ui.
- Update CSS for tooltip to allow for markdown comment scrolling when tooltip exceeds a certain height.
- Ensure tooltip stays within the infoview viewport by implementing custom flip logic as a floating-ui middleware. Tooltip position is based on pointer position instead of the size and position of the reference element - the element that is being hovered over. This allows for proper tooltip placement, regardless of the shape and size of the reference element. The default position of the tooltip is above and to the right of the pointer. The direction of the flip and the level of offset is based on the quadrant - within the viewport - in which the pointer was located when the hover event was triggered. Viewport quadrants are recalculated every time the hover event is triggered.

Below is a gif which exemplifies the typical behavior of the tooltip:
input mov

### DIFF
--- a/lean4-infoview/src/infoview/tooltips.tsx
+++ b/lean4-infoview/src/infoview/tooltips.tsx
@@ -3,13 +3,10 @@ import * as ReactDOM from 'react-dom'
 
 import {
   detectOverflow, useFloating, MiddlewareState, Side, Coords, SideObject,
-  flip, shift, offset, arrow, FloatingArrow
+  shift, offset, arrow, autoPlacement, autoUpdate, size, FloatingArrow
 } from '@floating-ui/react';
 
 import { forwardAndUseRef, forwardAndUseStateRef, LogicalDomContext, useLogicalDomObserver, useOnClickOutside } from './util'
-
-/** Tooltip contents should call `redrawTooltip` whenever their layout changes. */
-export type MkTooltipContentFn = (redrawTooltip: () => void) => React.ReactNode
 
 // Pointer coordinates used for tooltip placement.
 type PointerCoords = {pageX: number, pageY: number, clientX: number, clientY: number}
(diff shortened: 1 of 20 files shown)


Vtec234:
Thank you for the contribution @tnaoi! I like the idea, but I think it needs some improvement to be workable:
- Pointer-based positioning works well when the reference element is huge (like in your gif), but it's not great when the reference is small, i.e. in the majority of cases. In those cases the popup tends to obscure the reference; it would be better if it appeared next to the reference, as it does now.
  - I would suggest the following approach. Use the standard [visibility optimizers](https://floating-ui.com/docs/computeposition#visibility-optimizers) like in the [tutorial](https://floating-ui.com/docs/tooltip), and a standard placement strategy. I don't think we should be computing flips/quadrants/positioning (i.e. most of `pointer` ourselves) -- `floating-ui` can already do it for us! Instead, we should detect whether the tooltip is positioned "well" (this is intentionally fuzzy as I'm not certain what the check should be; perhaps whether the anchor point is within the viewport?), and only use pointer-based positioning when it is not. To implement pointer-based positioning, rather than writing the logic ourselves we should use a [virtual element](https://codesandbox.io/s/fancy-worker-xkr8xl?file=/src/index.js) set to the pointer coordinates as the reference.
- Good riddance to `TooltipPlacementContext`, it was not that useful.
- But let's not remove the `arrow` and `offset` middlewares.

Are you okay with making those changes?

lean4-infoview/src/infoview/tooltips.tsx
```diff
@@ -1,73 +1,167 @@
...
+      }
+    }
+
+    // [subpixel accelerated positioning]
+    // (https://floating-ui.com/docs/misc#subpixel-and-accelerated-positioning).
+    Object.assign(floatingEl.style, {
+      top: '0',
+      left: '0',
+      transform: `translate(${roundByDPR(coords.x)}px,${roundByDPR(coords.y)}px)`,
+    });
```
Vtec234, line 125: Rather than making an imperative update, could the `transform` be passed to the `div` returned from `Tooltip` directly? You can grab `x/y` from the return values of `useFloating`.

Vtec234 (changes requested):

tnaoi:
> Thank you for the contribution @tnaoi! I like the idea, but I think it needs some improvement to be workable:
> 
> * Pointer-based positioning works well when the reference element is huge (like in your gif), but it's not great when the reference is small, i.e. in the majority of cases. In those cases the popup tends to obscure the reference; it would be better if it appeared next to the reference, as it does now.
>   
>   * I would suggest the following approach. Use the standard [visibility optimizers](https://floating-ui.com/docs/computeposition#visibility-optimizers) like in the [tutorial](https://floating-ui.com/docs/tooltip), and a standard placement strategy. I don't think we should be computing flips/quadrants/positioning (i.e. most of `pointer` ourselves) -- `floating-ui` can already do it for us! Instead, we should detect whether the tooltip is positioned "well" (this is intentionally fuzzy as I'm not certain what the check should be; perhaps whether the anchor point is within the viewport?), and only use pointer-based positioning when it is not. To implement pointer-based positioning, rather than writing the logic ourselves we should use a [virtual element](https://codesandbox.io/s/fancy-worker-xkr8xl?file=/src/index.js) set to the pointer coordinates as the reference.
> * Good riddance to `TooltipPlacementContext`, it was not that useful.
> * But let's not remove the `arrow` and `offset` middlewares.
> 
> Are you okay with making those changes?

Thank you for the feedback @Vtec234, I'll try to implement the changes.

Vtec234:
Hi @tnaoi, are you still planning to make the changes? If not, I am happy to do them.

tnaoi:
@Vtec234 Sorry I've been neglecting this issue. Not sure if recent changes and requests require different solutions to the problem, so please make whatever changes necessary to address the issue.

Vtec234:
@tnaoi no problem at all, your contribution so far has already been helpful. I am building on it now with the new changes in order to address the issues I raised.

Vtec234:
@tnaoi in the end I wasn't sure how to incorporate the at-pointer positioning logic in a usable way. This leaves #281 unsolved, but is best to get the scrollability out now. If you figure out a nice way to incorporate it, a follow-up PR would be welcome!

Vtec234:
The Documentation View Example Test is failing consistently but none of that code changed in this PR, and it also failed in #1838.

PR #291, opened 2023-03-15
feat: Stop server command

semorrison:
This adds a "Lean 4: Stop server" command.

It's not ideal, as it generates a pop-up saying "Lean server printed an error: Watchdog error: Got `shutdown` request, expected an `exit` notification". If someone knows how to avoid this please let me know or make the change!


gebner:
Test failure is unrelated and the same as on the readme changes.

Issue #299, opened 2023-05-07
Infoview tweaks to clarify goal list


hrmacbeth:
I have noticed, watching novices, that their eyes often move right past the first goal and settle on a later one.  Honestly, I sometimes have this problem too -- I think the infoview is not yet as visually clear as the Lean 3 infoview was.  Here are several tweaks which could help with this problem:
- [X] #285, the suggestion to show a "number of goals" count (this was done in Lean 3)
- more whitespace between the goals
- an option to turn off the titles of the goals ("case: inl.inr" or whatever) -- the tradeoff between visual clutter and information for these titles is often not worth it
- an option to give visual emphasis to the first goal (coloured background? larger font?)
- an option to hide the "expected type" by default (this is a quick fix but a more intelligent solution would be https://github.com/leanprover/lean4/issues/2202, to suppress reporting "expected type" in the contexts when it's not useful)

From the [Zulip conversation](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/small.20.2237.20goals.22.20message) where this was discussed, here is a screenshot showing what the infoview currently looks like when there are many goals.
<img width="1440" alt="Screen-Shot-2023-04-25-at-10 38 15-AM" src="https://user-images.githubusercontent.com/25316162/236692917-6efc175b-7148-4b78-994e-de2469650d8b.png">

PatrickMassot:
@gebner @Vtec234 @EdAyers could you clarify whether all the items on Heather's list would indeed be welcomed PRs?

Vtec234:
@PatrickMassot as far as I am concerned, all of them are :) For the potential options, we have a `lean4.infoview` namespace in the settings declared in `vscode-lean4/package.json` that could be used.

Vtec234:
Done in #302.

PR #300, opened 2023-05-13
Bring back goals count in tactic state.

### LINKED ISSUE #285: Show number of goals
In Lean 3, we used to show the number of goals (i.e., "42 goals").  Heather would like to see that come back.

PatrickMassot:
Fixes #285

This is my naive attempt at fixing one of the issues with the current InfoView, but I'm ready to do it differently if this is not suitable.

### DIFF
--- a/lean4-infoview/src/infoview/goals.tsx
+++ b/lean4-infoview/src/infoview/goals.tsx
@@ -158,13 +158,15 @@ export const Goal = React.memo((props: GoalProps) => {
 interface GoalsProps {
     goals: InteractiveGoals
     filter: GoalFilterState
+    displayCount: boolean
 }
 
-function Goals({ goals, filter }: GoalsProps) {
+function Goals({ goals, filter, displayCount}: GoalsProps) {
     if (goals.goals.length === 0) {
         return <>No goals</>
     } else {
         return <>
+            {displayCount && <strong className="goal-goals">{goals.goals.length} goals</strong>}
             {goals.goals.map((g, i) => <Goal key={i} goal={g} filter={filter} />)}
         </>
     }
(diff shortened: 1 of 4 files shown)


Vtec234:
For reference, this is what it looks like now:
Screenshot 2023-05-14 at 10 52 45 PM

This is the same as in Lean 3. Having long forgotten how it looked there, I half-expected it to be writte `Tactic state (N goals)` which would some vertical space. Any preference between the current version and this alternative @PatrickMassot?

lean4-infoview/src/infoview/goals.tsx
```diff
@@ -158,13 +158,15 @@ export const Goal = React.memo((props: GoalProps) => {
...
+    displayCount: boolean
 }
 
-function Goals({ goals, filter }: GoalsProps) {
+function Goals({ goals, filter, displayCount}: GoalsProps) {
     if (goals.goals.length === 0) {
         return <>No goals</>
     } else {
         return <>
+            {displayCount && <strong className="goal-goals">{goals.goals.length} goals</strong>}
```
Vtec234, line 169:
```suggestion
    const nGoals = goals.goals.length
    if (nGoals === 0) {
        return <>No goals</>
    } else {
        return <>
            {displayCount &&
                <strong className="goal-goals">{nGoals} {1 < nGoals ? 'goals' : 'goal'}</strong>}

```
(If we keep this variant.)

PatrickMassot:
I don't have very strong opinion about the layout, I simply copied the Lean 3 version. But remember this is all part of #299 about clarifying the info view so my goal is really to have this count visible. I'm afraid that your suggestion will weaken this.

PR #301, opened 2023-05-15
ci: remove failing test

Vtec234:
This test appears to fail often, e.g. in #300 and #291. This might be preferrable in any case as I am not sure the extension should care about where on disk the Lean toolchain is stored.


gebner:
I think the test makes sense in general, we're checking that the extension knows where Lean is located.  And the elan naming scheme is really stable.

I'm not sure why it's failing here and I wonder what the actual value of `getWorkspaceFolder` is.  In the end it's probably some race condition as always so I'm fine with removing the test.

PR #302, opened 2023-05-26
Infoview tweaks

PatrickMassot:
This implements all suggestions from #299 to clarify the infoview display, especially in the presence of several goals.

---

This isn't yet fully working, I'll discuss the issue on Zulip.


Vtec234:
@PatrickMassot could you try it now?

PatrickMassot:
Yes, it seems to work, thanks!

PR #314, opened 2023-08-12
Escape "-" in wordPattern regexp

DenisGorbachev:
Since "-" was not escaped, the sequence ")-=" was treated as "a range from ) to =". 

Escaping "-" fixes this.

Examples:

```
const r = new RegExp("([^`~@$%^&*()\\-=+\\[{\\]}⟨⟩⦃⦄⟦⟧⟮⟯‹›\\\\|;:\",./\\s]+)")

'SomeName123'.match(r)[0] === 'SomeName123'
'Some-Name-123'.match(r)[0] === 'Some'
```

P.S. Two backslashes are needed because the "wordPattern" is a key in a JSON file, which requires its own escape sequence.

P.P.S. This fix is needed for "Rename symbol" functionality in https://github.com/DenisGorbachev/vscode-lean4-code-actions


DenisGorbachev:
Could you please take a look at this? (it's a fairly simple fixe)

mhuisi:
Closing this due to it being subsumed by #315.

PR #321, opened 2023-09-06
Use stable release for bootstrapping

mhuisi:
This PR adjusts the bootstrapping to load Lean 4 stable instead of Lean 4 nightly in preparation for the first release. It should not be merged before the release is out. I also need to do some additional QA after the release and before merging, as this is currently entirely untested.


mhuisi:
I did some manual testing and it seems to work as expected.

vscode-lean4/src/leanclient.ts
```diff
@@ -542,14 +542,17 @@ export class LeanClient implements Disposable {
...
-            const match = /^leanprover\/lean4:nightly-(\d+)-(\d+)-(\d+)$/.exec(toolchainVersion);
-            if (match) {
-                return new Date(parseInt(match[1]), parseInt(match[2]) - 1, parseInt(match[3]));
+            const nightly_match = /^leanprover\/lean4:nightly-(\d+)-(\d+)-(\d+)$/.exec(toolchainVersion);
+            if (nightly_match) {
+                return new Date(parseInt(nightly_match[1]), parseInt(nightly_match[2]) - 1, parseInt(nightly_match[3]));
+            }
+            const release_match = /^leanprover\/lean4:(\d+)-(\d+)-(\d+)$/.exec(toolchainVersion);
+            if (release_match) {
+                return new Date(2023, 9, 8);
```
Kha, line 555: Oof. Is it time to get rid of the Lake conditional?

Kha (approved):

vscode-lean4/src/leanclient.ts
mhuisi: I think it's a good idea to get rid of it eventually, though not in this PR. I'm not sure about the ways in which people might acquire very old setups, so perhaps it's better to leave it in for now.

PR #328, opened 2023-09-21
feat: indent on enter after starting a focus block

thorimur:
This PR adds the following to `language-configuration.json` to cause indentation after starting a focus block (`·`):
```json
    "onEnterRules" : [
        {
            // Indent focus blocks
            "beforeText" : "^\\s*(·|\\.)\\s.*$",
            "action" : { "indent" : "indent" }
        }
    ]
```
Now, hitting enter after `intro x` yields
```lean
constructor
· intro x
  | <-- cursor ends up here
```

See [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/VSCode.20indentation).

-----

Note: there are some cases where this will produce the incorrect indentation, e.g.
```
  · have : Foo := by
    | <-- cursor will be here

  · have : Foo := by
      | <-- cursor should be here
```
but handling EOL tokens like `by` is done in #329 .


thorimur:
awaiting-review

vscode-lean4/language-configuration.json
```diff
@@ -130,5 +130,12 @@
         ["\"", "\""],
         ["'", "'"]
     ],
-    "wordPattern" : "([^`~@$%^&*()\\-=+\\[{\\]}⟨⟩⦃⦄⟦⟧⟮⟯‹›<>\\\\|;:\",./\\s]+)"
+    "wordPattern" : "([^`~@$%^&*()\\-=+\\[{\\]}⟨⟩⦃⦄⟦⟧⟮⟯‹›<>\\\\|;:\",./\\s]+)",
+    "onEnterRules" : [
+        {
+            // Indent focus blocks
```
mhuisi, line 136: It seems that VSCode uses JSONC and can deal with these comments, but I would still prefer it if these kinds of comments were put somewhere else so that these files are valid JSON. How about creating a `language-configuration.md` in `docs` and moving this comment and the ones in the other PRs there?

thorimur: Sounds good! :) Done in de81d7.

PR #330, opened 2023-09-21
feat: provide more bracket support

thorimur:
Lean uses a lot of different brackets! This PR augments `language-configuration.json` to account for some more. This primarily affects
* highlighting
* select-to-bracket
* autoclosing

Specifically, this PR:

* Adds support for "modified standard brackets": <code>\`()</code>,<code>\`\`()</code>, `#[]`, `@[]`, and `%[]`. All of these are defined in Lean core. This primarily affects highlighting. Note that `$[]` is specifically not considered a bracket pair by this PR.
* Adds support for "modified string literals": `s!""`, `f!""`, and `m!""`. Previously, typing e.g. `m!"` would not autoclose the quotation and insert the second quote for you.
* Adds "surrounding" support for markdown enclosers: <code>\`\`</code>, `**`, and <code>__</code>. This means that you can select `text`, type e.g. a single <code>\`</code>, and the text will become <code>\`text\`</code>. This is useful when writing comments, but we don't want these to autoclose or be regarded as brackets, so we only add this to `surroundingPair`.

Note that `surroundingPair` only makes sense with a single keystroke (I believe), so we don't add modified standard brackets or modified string literals to `surroundingPair`.

-----

This PR currently takes the view that trailing tokens like `!`, `?` (for array access) etc., are not "part of the brackets". However, things might just look cleaner—and clearer—if we *did* consider them part of the brackets for the purposes of highlighting. Opinions welcome. :)


Kha:
I find it confusing that the different `$`s are highlighted differently. This is not how I think about antiquotations.

thorimur:
@Kha That's fair—my brain is happy to use the color here only for grouping (i.e. to indicate that these `[]` brackets are in some sense 'part of' the `$` process, while the `$` alone is what indicates antiquotationhood, independently of color), but I can see the importance of/intuition in having all `$`s in this context convey their equivalent nature.

In any case, seeing as you were involved in actually *making* the antiquotations in Lean in the first place, I'd say your opinion carries the most weight in how they ought to be highlighted...! :)

vscode-lean4/language-configuration.json
```diff
@@ -5,7 +5,13 @@
     },
     "brackets": [
         ["(", ")"],
+        ["`(", ")"],
+        ["``(", ")"],
         ["[", "]"],
+        ["#[", "]"],
+        ["@[", "]"],
+        ["%[", "]"],
+        ["$[", "]"],
```
mhuisi, line 14: I would also vote to leave this rule out.

vscode-lean4/language-configuration.json
```diff
@@ -5,7 +5,13 @@
     },
     "brackets": [
         ["(", ")"],
+        ["`(", ")"],
+        ["``(", ")"],
         ["[", "]"],
+        ["#[", "]"],
+        ["@[", "]"],
```
mhuisi, line 12: I don't have a strong opinion on this one, either.

vscode-lean4/language-configuration.json
thorimur: Removed in 5c13f5c . :)

mhuisi (approved):
Sorry for not getting back to you earlier - this PR looks great!

PR #343, opened 2023-10-27
fix: update semver

joneugster:
updating `semver` as used version reports vulnerability.


joneugster:
The deprecated version causes lean4web to report medium vulnerability issues, see [CI output](url).

joneugster:
See [CI output](https://github.com/leanprover-community/lean4web/actions/runs/6669443884/job/18127228266#step5:7) of lean4web, which depends on this node package.

joneugster:
> The package-lock for this PR needs to be generated differently. You can apply the following patch to fix it.
done.

PR #354, opened 2023-11-10
Update lean toolchain after `lake update`

mhuisi:
This PR adjusts the `Project: Update Dependency` command so that it updates the toolchain *after* running `lake update`. This was prompted by the introduction of a post-update hook in leanprover-community/mathlib4#8243 that automatically updates the toolchain.

With the post-update hook and asking to update the toolchain before running `lake update`, a user decision of "Do not update the toolchain" would be ignored for projects downstream of Mathlib, as the post-update hook will update the toolchain regardless of the user choice. Asking to update the toolchain (if it changed) after `lake update` ensures users do not see a misleading dialog.

One caveat is that this means that we will now run `lake update` with the Lean version of the project itself, not the Lean version of its dependency.

Another change induced by switching the order of operations is that we can now read the toolchain from `lake-packages` instead of fetching it from `raw.githubusercontent.com`.

### DIFF
--- a/vscode-lean4/src/projectoperations.ts
+++ b/vscode-lean4/src/projectoperations.ts
@@ -155,22 +155,7 @@ export class ProjectOperationProvider implements Disposable {
             return
         }
 
-        const localToolchainPath: string = join(activeClient.folderUri.fsPath, 'lean-toolchain')
-        const dependencyToolchainResult: string | 'DoNotUpdate' | 'Cancelled' = await this.determineDependencyToolchain(localToolchainPath, dependencyChoice)
-        if (dependencyToolchainResult === 'Cancelled') {
-            return
-        }
-
         await this.runOperation(async lakeRunner => {
-            if (dependencyToolchainResult !== 'DoNotUpdate') {
-                try {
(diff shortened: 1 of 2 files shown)


vscode-lean4/src/projectoperations.ts
```diff
@@ -181,6 +166,22 @@ export class ProjectOperationProvider implements Disposable {
             }
 
             await this.tryFetchingCache(lakeRunner)
+
+            const localToolchainPath: string = join(activeClient.folderUri.fsPath, 'lean-toolchain')
+            const dependencyToolchainPath: string = join(activeClient.folderUri.fsPath, 'lake-packages', dependencyChoice.name, 'lean-toolchain')
```
tydeu, line 171: Hard-coding this problematic. For one, [the Lake directory change](https://github.com/leanprover/lean4/pull/2749) will break it. Ideally, this would use the `packagesDir` setting from the root Lake manifest instead.

mhuisi: Good point! I'll adjust it.

PR #359, opened 2023-11-17
fix: update lake manifest parsing for manifest version 7

mhuisi:
"Update Dependency" was broken because of this since [v4.3.0-rc2](https://github.com/leanprover/lean4/releases/tag/v4.3.0-rc2).

### DIFF
--- a/vscode-lean4/src/utils/manifest.ts
+++ b/vscode-lean4/src/utils/manifest.ts
@@ -15,15 +15,8 @@ export interface Manifest {
     directGitDependencies: DirectGitDependency[]
 }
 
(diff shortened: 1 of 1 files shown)


tydeu:
Oops. Sorry! Should we have (and is it possible to have) some automation that checks the VS Code extension for breakages like we do for mathlib? /cc @semorrison

mhuisi:
I don't think maintaining this kind of automation is worth it yet, especially since it is more difficult to detect breakages here than in Mathlib.

tydeu:
@mhuisi Makes sense! In that case, I will try to keep in mind that I should notify you in advance of any manifest changes. 😃

PR #366, opened 2023-12-03
feat: Add option to open Mathlib API Documentation

hmonroe:


mhuisi:
The DocView is not capable of displaying the mathlib API documentation. The sidebar is broken and the search does not work. I will resolve this in the future by replacing the DocView entirely; see also [this post on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Call.20for.20help.3A.20Loogle.20in.20VS.20Code/near/404603083).

PR #367, opened 2023-12-03
Update dev.md lerna commands are outdated (and still not working)

hmonroe:


docs/dev.md
```diff
@@ -14,7 +14,7 @@ See the following design topics:
 
 ### Building
 - Make sure you have an up to date installation of `npm` and `node.js`.  For example `npm` version 8.1.3 and `node.js` version v16.13.0.
-- Run `npm install` in this folder. This installs the Lerna package manager.
+- Run `npm install lerna` in your root folder for development. This installs the Lerna package manager.
```
Vtec234, line 17:
```suggestion
- Run `npm install` in your workspace root folder. This installs the Lerna package manager.
```
You don't have to explicitly `install lerna` as it is in `package.json`.

PR #371, opened 2023-12-06
feat: customizable occurrence highlighting

david-christiansen:
Makes it possible to disable substring-based occurrence highlighting and use only Lean's compiler responses.

I personally found the string-based highlighting to be distracting and confusing. This change defaults to semantic-only highlighting at @mhuisi 's request, so it will be user-visible.

### DIFF
--- a/vscode-lean4/package.json
+++ b/vscode-lean4/package.json
@@ -64,6 +64,21 @@
 					"default": true,
 					"markdownDescription": "Enable eager replacement of abbreviations that uniquely identify a symbol."
 				},
+				"lean4.editor.occurrenceHighlightingMode": {
+					"type": "string",
+					"description": "Highlight occurrences using Lean's semantic information, by string matching, or both",
+					"default": "semantic",
+					"enum": [
+						"both",
+						"semantic",
+						"strings"
+					],
+					"enumDescriptions": [
+						"Use Lean's semantic information with string matching as fallback",
+						"Use only Lean's semantic information",
+						"Use only string matching"
+					]
+				},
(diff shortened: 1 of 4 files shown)


vscode-lean4/package.json
```diff
@@ -64,6 +64,21 @@
 					"default": true,
 					"markdownDescription": "Enable eager replacement of abbreviations that uniquely identify a symbol."
 				},
+				"lean4.editor.occurrenceHighlightingMode": {
+					"type": "string",
+					"description": "Highlight occurrences using Lean's semantic information, by string matching, or both",
```
Vtec234, line 69:
```suggestion
					"description": "Highlight occurrences of a symbol using Lean's semantic information, by string matching, or both",
```
Maybe slightly clearer as to what it is that's occurring?

david-christiansen:
I think #377 is strictly better than this, but I'll leave it up to a maintainer to decide what to merge and what to close.

vscode-lean4/package.json
david-christiansen: I intentionally avoided this phrasing because the string-based highlighting doesn't match symbols (which is why I want to turn it off - I find it undermines the usefulness of the highlights if I can't trust them). But perhaps I'm being too pedantic here.

How do you feel about the phrasing in #377?

Vtec234: Gotcha! The other PR is worded well.

david-christiansen: Thanks! I tried hard to thread that needle, glad it succeeded

mhuisi:
Subsumed by the now merged #377.

PR #377, opened 2023-12-14
feat: customizable occurrence highlighting

david-christiansen:
Makes it possible to disable substring-based occurrence highlighting and use only Lean's compiler responses.

This is an alternative version of #371 that simplifies the settings interface on the assumption that everyone wants the correct highlights. @mhuisi - this is what we talked about the other day.

### DIFF
--- a/vscode-lean4/package.json
+++ b/vscode-lean4/package.json
@@ -64,6 +64,11 @@
 					"default": true,
 					"markdownDescription": "Enable eager replacement of abbreviations that uniquely identify a symbol."
 				},
+				"lean4.fallBackToStringOccurrenceHighlighting": {
+					"type": "boolean",
+					"description": "Fall back to string-based occurrence highlighting when there are no semantic symbol occurrences from the Lean language server to highlight.",
+					"default": false
+				},
 				"lean4.automaticallyBuildDependencies": {
 					"type": "boolean",
 					"default": false,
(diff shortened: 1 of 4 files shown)


david-christiansen:
It looks like the Windows CI failure is a repo-wide issue rather than something from this PR in particular, but I don't have a Windows machine handy to manually test on.

mhuisi:
No worries, the tests are somewhat flaky.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Read "lean-toolchain" files and do not set --default-toolchain to something specific.
 - elan should not install multiple toolchains.
+- Occurrence highlighting can be configured to ignore string matches and use only Lean's LSP responses
```
mhuisi, line 7: As previously discussed, you can leave this out because we don't really update `CHANGELOG.md` anymore.

vscode-lean4/package.json
```diff
@@ -64,6 +64,11 @@
 					"default": true,
 					"markdownDescription": "Enable eager replacement of abbreviations that uniquely identify a symbol."
 				},
+				"lean4.editor.occurrenceHighlightWithStringFallback": {
```
mhuisi, line 67: How about `fallBackToStringOccurrenceHighlighting`? Otherwise it is not as clear from the name alone that setting this to `false` does not disable all occurrence highlighting. I would also leave out the `editor` prefix.

vscode-lean4/package.json
```diff
@@ -64,6 +64,11 @@
...
+					"type": "boolean",
+					"description": "Fall back to string occurrence matching when there are no symbol occurrences from Lean to highlight.",
```
mhuisi, line 69:
```suggestion
					"description": "Fall back to string-based occurrence highlighting when there are no semantic symbol occurrences from the Lean language server to highlight.",
```

vscode-lean4/src/config.ts
```diff
@@ -227,6 +227,10 @@ export function shouldShowInvalidProjectWarnings(): boolean {
     return workspace.getConfiguration('lean4').get('showInvalidProjectWarnings', true)
 }
 
+export function getOccurrenceHighlightStringFallback(): boolean {
+    return workspace.getConfiguration('lean4.editor').get('occurrenceHighlightWithStringFallback', true)
```
mhuisi, line 231: The default for this should be `false`, not `true`.

vscode-lean4/src/leanclient.ts
```diff
@@ -634,29 +634,30 @@ export class LeanClient implements Disposable {
...
-                    const text = doc.getText();
-                    const nonWordPattern = '[`~@$%^&*()-=+\\[{\\]}⟨⟩⦃⦄⟦⟧⟮⟯‹›\\\\|;:\",./\\s]|^|$'
-                    const regexp = new RegExp(`(?<=${nonWordPattern})${escapeRegExp(word)}(?=${nonWordPattern})`, 'g')
-                    for (const match of text.matchAll(regexp)) {
-                        const start = doc.positionAt(match.index ?? 0)
-                        highlights.push({
-                            range: new Range(start, start.translate(0, match[0].length)),
-                            kind: DocumentHighlightKind.Text,
-                        })
+                    if (getOccurrenceHighlightStringFallback()) {
```
mhuisi, line 637: I would prefer an early return here :-)

CHANGELOG.md
david-christiansen: Done. Should we delete the file?

vscode-lean4/package.json
david-christiansen: OK, I made the change.

vscode-lean4/src/config.ts
david-christiansen: Indeed!

vscode-lean4/src/leanclient.ts
david-christiansen: Is the latest modification what you were looking for?

david-christiansen:
Thanks for the feedback!

CHANGELOG.md
mhuisi: Yes, I think it's fine to delete it. I don't think the VS Code marketplace uses it and given how incomplete it is, it can't be particularly helpful to anyone.

vscode-lean4/src/leanclient.ts
mhuisi: Yes!

mhuisi (approved):

PR #379, opened 2023-12-22
Export `LocationsContext` in `@leanprover/infoview`

0art0:
The `LocationsContext` is exported in `infoview/goalLocation.tsx` but not in the `.index.tsx` file. It is sometimes useful to be able to use the `LocationsContext` while writing widgets.

### DIFF
--- a/lean4-infoview/src/index.tsx
+++ b/lean4-infoview/src/index.tsx
@@ -12,7 +12,7 @@ export { EditorContext, VersionContext } from './infoview/contexts';
 export { EditorConnection } from './infoview/editorConnection';
 export { RpcContext } from './infoview/rpcSessions';
 export { ServerVersion } from './infoview/serverVersion';
-export { GoalLocation, GoalsLocation } from './infoview/goalLocation';
+export { LocationsContext, GoalLocation, GoalsLocation } from './infoview/goalLocation';
 export { importWidgetModule, DynamicComponent, DynamicComponentProps,
     PanelWidgetProps } from './infoview/userWidget';
 


0art0:
I'd like to bump this PR. I couldn't find an `easy` label for PRs like in `Mathlib`.

Vtec234:
Sorry for the slow turnaround: I don't get notifications for this repo. Please feel free to ping me if you are trying to do something with `lean4-infoview` or `lean4-infoview-api`.

As for the issue: we could definitely expose this, but have you actually tried using it, for instance by building a local copy of `vscode-lean4` with the `LocationsContext` exported? As far as I can remember, `LocationsContext` is set in the goal part of the infoview, but not in the widget part. Check out [this snippet](https://github.com/leanprover/vscode-lean4/blob/82464991f7e221411248c18f49d08cc4fc13790f/lean4-infoview/src/infoview/info.tsx#L118-L148). If I am not mistaken, then, this context would be useless in a widget (but please do double-check).

In fact, ideally we would just allow widgets to modify the selection context by exposing it for widgets. I don't remember if we have an issue for it, but I would be happy to review a PR implementing this functionality. Note that it *is* quite complicated: widgets decide which goals they want to display and when, and correctness of all that would have to be accounted for in the PR.

0art0:
Thanks for your detailed reply. 

At the moment, I am exporting `LocationsContext` by patching the local copy of the `@leanprover/infoview` package. 

In my use-case, I am setting the `LocationsContext` with a new value: https://github.com/Human-Oriented-ATP/tree-rewriting-game/blob/217660b9479b929af5804af73e29658dba1cee92/widget/src/treeRewritingGame.tsx#L24. This is used to allow shift-clicking of sub-expressions in a widget rendered in this context and to convey the list of selections back to Lean.

Currently, testing this code out requires a modified VS Code extension that has `LocationsContext` exported -- having this context exported by default would allow this code to be run with the regular `lean4` VS Code extension.

Modifying the selection context associated with the actual goal is not necessary for my use-case, though it would certainly be an interesting feature that I might like to work on in the future.

Vtec234:
Ah, so you set up a `LocationsContext` yourself, and then the integration with `InteractiveCode` and `Expr`ession displays selects things in it for you using the standard shift-click mechanism? I didn't think of that use case!

0art0:
Yes, we're using the `InteractiveExpr` and `LocationsContext` machinery to allow shift-click selections within custom widgets.

Vtec234:
@0art0 I made a new release, it should contain the exported symbol.

0art0:
Thank you!

PR #399, opened 2024-02-23
fix: More arrows

bustercopley:
Add "⇰", and also "|->" as an alias for "mapsto".


bustercopley:
Suggested by @philnguyen [here](https://github.com/leanprover/lean4-mode/pull/52/)

philnguyen:
Thanks @bustercopley so much!! I was going to do this myself

PR #408, opened 2024-03-15
feat: add "go to definition" entry to infoview context menu

mhuisi:
This PR adds the "Go to definition" feature that was already available using Ctrl+Click to the context menu of the InfoView as well, hopefully making it much more discoverable.

- [x] Update infoview and infoview-api versions


vscode-lean4/package.json
```diff
@@ -278,6 +278,12 @@
 				"title": "Infoview: Toggle Expected Type",
 				"description": "Toggle the display of expected type."
 			},
+			{
+				"command": "lean4.infoview.goToDefinition",
```
Vtec234, line 282: I hate how roundabout this implementation has to be because of VSCode's limitations. Is there a way to hide this command from users' views? If you run it from the command palette right now it just crashes, and even if it didn't it is guaranteed to be confusing (since, for example, it will not go-to-def for the token under the cursor which is a plausible interpretation of what it might do based on just the name).

lean4-infoview/src/infoview/event.ts
```diff
@@ -5,20 +5,39 @@ import type { Disposable } from 'vscode-languageserver-protocol';
  * to all registered handlers. Handlers can be registered using `on`. */
 export class EventEmitter<E> {
     private handlers: ((_: E) => void)[] = [];
+    private handlersWithKey = new Map<string, ((_: E) => void)[]>();
```
Vtec234, line 8: What is the purpose of keys? Instead of having keys, one can alternatively fire all the handlers for a given event and check whether the key-parameter matches the expected one in specific handlers, right? In this case, something like
```typescript
  useEvent(ec.events.goToDefinition, key => { if (key === interactiveCodeTagId) { execGoToLoc() } })
```
since, if I understand correctly, this doesn't add any new functionality, is it an optimization? Have you measured it to have an observable impact? If not, I would just do the check in the handler in order to have less code.

lean4-infoview-api/src/infoviewApi.ts
```diff
@@ -140,6 +140,13 @@ export interface InfoviewApi {
    */
   requestedAction(action: InfoviewAction): Promise<void>;
 
+  /**
+   * Must fire whenever the user requests to go to a definition using the infoview context menu.
+   * `interactiveCodeTagId` is obtainable from the `data-vscode-context` field in
```
Vtec234, line 145: Could we point to the VSCode context menu docs here? It's probably not really possible to understand what is going on here without that context.

lean4-infoview/src/infoview/interactiveCode.tsx
```diff
@@ -153,17 +158,23 @@ function InteractiveCodeTag({tag: ct, fmt}: InteractiveTagProps<SubexprInfo>) {
...
+  // When selecting a custom context menu entry, VS Code will execute a VS Code command
+  // parameterized with `data-vscode-context`. We then use this context to execute the
+  // command in the context of the correct interactive code tag in the InfoView.
+  const interactiveCodeTagId = React.useId()
+  const vscodeContext = { interactiveCodeTagId }
+  useEvent(ec.events.goToDefinition, _ => execGoToLoc(), undefined, interactiveCodeTagId)
+
   return (
     <WithTooltipOnHover
+      data-vscode-context={JSON.stringify(vscodeContext)}
```
Vtec234, line 171: I am wondering how difficult it would be to store enough information here for the editor extension to be able to go-to-def without sending a message back to the infoview. The main issue appears to be that we don't eagerly know the location one has to go to (`fetchGoToLoc` does that). We could in principle only set this context after `fetchGoToLoc` has replied, but that might make for some bad UX wherein the 'Go to Definition' option is sometimes there and sometimes not (or is always there, but doesn't always work).

Vtec234 (changes requested):
It will be great to have this functionality; I have some questions about the implementation.

vscode-lean4/package.json
mhuisi: Yes, we can hide it with `"when": "false"`. I initially thought that not adding it to the command palette menu would be sufficient, but apparently VS Code includes all commands there regardless.

lean4-infoview/src/infoview/event.ts
mhuisi: I haven't measured it, but the main issue I see with not using a key parameter is that [React will run both the code that adds the handler as well as the code that removes it on every re-render](https://react.dev/reference/react/useEffect#connecting-to-an-external-system).

Since removing the handler without the key is linear in the amount of handlers and an amount of handlers that is linear in the size of the tree is being removed, removing all handlers on every re-render is quadratic in the size of the tree (just the delaboration of which is already causing performance issues for some users).

This seems like a potential performance bug to me.

lean4-infoview-api/src/infoviewApi.ts
mhuisi: Sure!

lean4-infoview/src/infoview/interactiveCode.tsx
mhuisi: Yes, I didn't see a nice way to implement this differently, either. 

We could in principle invert the control flow: The 'Go to definition' option is always displayed but when clicked waits until the infoview fires an event for the given id either with a location or undefined. I don't think this is much better, though.

vscode-lean4/package.json
```diff
@@ -278,6 +278,12 @@
...
+				"category": "Lean 4",
+				"title": "Go to Definition",
+				"description": "Go to definition for the currently selected element in the InfoView."
```
Vtec234, line 285:
```suggestion
				"title": "(Internal) Go to Definition",
				"description": "This command is an implementation detail of the 'Go to Definition' context menu option in the infoview."
```
I would make it super explicit that this is a impl detail, and one should look at the infoview code to understand what's actually happening.

lean4-infoview/src/infoview/interactiveCode.tsx
Vtec234: There are no good solutions here 😅

vscode-lean4/package.json
mhuisi: The context menu displays the title. I believe the description is not user-facing, though.

Vtec234: I missed that, let's just change `description` then.

lean4-infoview/src/infoview/event.ts
Vtec234: I see what you mean! A couple of comments:
- The event handler will not be re-added on every render, but rather only on those renders on which the dependencies have changed. I defined `deps` in all occurrences of `useEvent` now.
- There is still an issue you pointed out in the linear complexity of `filter` invoked during handler removal; I changed it to use handler IDs instead, which should be amortized constant time.
- The issue of linear complexity remains in the case of an event firing, at which point all the keyless handlers would be invoked; so let's keep keyed handlers.

Vtec234 (approved):
I made use of keyed handlers in other locations (may as well, if we are adding them). LGTM now, though it would be good to test it more; I may have introduced bugs in the changes.

mhuisi:
Oops. It should be resolved now.

lean4-infoview/src/infoview/tooltips.tsx
```diff
@@ -99,19 +99,26 @@ export const WithFadingTooltip =
...
       onEntered={() => hideTooltip()}
     >
     {state =>
-      <Tooltip
-        reference={ref}
-        style={fadeOutStyles[state]}
-      >
-        {tooltipChildren}
-      </Tooltip>}
+      <div ref={nodeRef}>
```
Vtec234, line 114: It looks like this introduced a regression: the `div`s introduce linebreaks:
Screenshot 2024-03-25 at 11 47 05 PM

Changing the `div` to a `span` seems to fix it. However, I would advise against pulling in `react-transition-group`. The last commit was 2 years ago. This could mean it's very stable, but the presence of issues like the one you linked (and many others) suggests it's actually just abandoned. Do you think the "No definition found" popup is essential? It's nice to have for sure, but just having the action fail seems like a clear enough signal for users, and will definitely simplify maintenance.

mhuisi: Sorry for wasting your time with all those messy mistakes. Not sure why I didn't notice the broken formatting when testing the change.

I do think the popup is essential to reduce confusion and I agree that avoiding this dependency is preferable. I tried to make the fade-out animation work using `onTransitionEnd` for a while before including the dependency but gave up eventually. If you know of a nice way to implement this transition without the library, please go ahead :-)

Vtec234: There probably is a way, but I don't have the time currently to experiment. If you are willing to maintain this stuff (please note: the tooltip code is *the most complicated* part of the infoview), feel free to merge (but please squash the few 'fix' commits towards the end).

mhuisi: I always squash by default these days. I'm willing to maintain this change! If it breaks, I'll invest the time necessary to get rid of the dependency.

PR #415, opened 2024-03-20
Remove `_foo_` as an auto-closing delimiter

digama0:
This interferes with the behavior in regular lean code where you may want to select a subterm and replace it with `_`.


mhuisi:
Zulip thread related to this PR: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/vscode.20server.3A.20selection.20.2B.20_

PR #419, opened 2024-03-22
feat: replace stale dep notifs with "restart file" infoview button

mhuisi:
This PR adds a "Restart File" button to the bottom right of the InfoView and deletes stale dependency notifications entirely.
I've found the latter to be way too noisy, especially now that Lean 4 on master will inform us about stale dependencies as soon as the dependency changes (it's especially noisy when switching branches).
The central issue with the notifications is that they stick around even after you switch away from the file.

Now, after fixing the `fullRange` behavior of the extension, the stale dependency diagnostic will be displayed in "Messages" anywhere in the file. The button in the InfoView should also be sufficiently visible that after reading the message, users know what to do.

This PR also fixes a bug where the InfoView would not correctly display the "imports out of date" error in the InfoView because the server doesn't respond to requests when it doesn't keep going after processing the header.

### DIFF
--- a/lean4-infoview-api/package.json
+++ b/lean4-infoview-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@leanprover/infoview-api",
-  "version": "0.2.1",
+  "version": "0.3.0",
   "description": "Types and API for @leanprover/infoview.",
   "scripts": {
     "watch": "tsc --watch",
--- a/lean4-infoview-api/src/infoviewApi.ts
+++ b/lean4-infoview-api/src/infoviewApi.ts
@@ -57,6 +57,9 @@ export interface EditorApi {
   /** Highlight a range in a document in the editor. */
   showDocument(show: ShowDocumentParams): Promise<void>;
 
+  /** Restarts the given file. */
+  restartFile(uri: string): Promise<void>
+
   /**
    * Creates an RPC session for the given uri and returns the session id.
    * The extension takes care of keeping the RPC session alive.
--- a/lean4-infoview/package.json
+++ b/lean4-infoview/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@leanprover/infoview",
-  "version": "0.4.5",
+  "version": "0.5.0",
   "description": "An interactive display for the Lean 4 theorem prover.",
   "scripts": {
     "watch": "rollup --config --environment NODE_ENV:development --watch",
@@ -47,8 +47,9 @@
     "typescript": "^4.9.5"
   },
   "dependencies": {
-    "@leanprover/infoview-api": "~0.2.1",
+    "@leanprover/infoview-api": "~0.3.0",
     "@vscode/codicons": "^0.0.32",
+    "@vscode/webview-ui-toolkit": "^1.4.0",
     "es-module-shims": "^1.7.3",
     "marked": "^4.3.0",
     "react-fast-compare": "^3.2.2",
--- a/lean4-infoview/src/infoview/index.css
+++ b/lean4-infoview/src/infoview/index.css
@@ -108,6 +108,13 @@ html,body {
 
 /* Headers for infoview */
 
+.restart-file-button {
+    position: fixed;
+    bottom: 10px;
+    right: 10px;
+    z-index: 37;
+}
+
 select {
     background-color: var(--vscode-dropdown-background);
     color: var(--vscode-dropdown-foreground);
--- a/lean4-infoview/src/infoview/info.tsx
+++ b/lean4-infoview/src/infoview/info.tsx
@@ -300,9 +300,11 @@ function InfoAux(props: InfoProps) {
         // While `lake print-paths` is running, the output of Lake is shown as
         // info diagnostics on line 1.  However, all RPC requests block until
         // Lake is finished, so we don't see these diagnostics while Lake is
-        // building.  Therefore we show the LSP diagnostics on line 1 if the
+        // building. Therefore we show the LSP diagnostics on line 1 if the
         // server does not respond within half a second.
-        if (pos.line === 0 && lspDiagsHere.length) {
+        // The same is true for fatal header diagnostics like the stale dependency notification.
+        const isAllHeaderDiags = lspDiagsHere.length > 0 && lspDiagsHere.every(diag => diag.range.start.line === 0)
+        if (isAllHeaderDiags) {
             setTimeout(() => resolve({
                 pos,
                 status: 'updating',
--- a/lean4-infoview/src/infoview/main.tsx
+++ b/lean4-infoview/src/infoview/main.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import * as ReactDOM from 'react-dom/client';
 import type { DidCloseTextDocumentParams, Location, DocumentUri } from 'vscode-languageserver-protocol';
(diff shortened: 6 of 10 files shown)


Vtec234:
I can confirm this works, but I don't love the look of the button. It is really distracting and visually distinct from all the other buttons (which use codicons instead of the webview-ui-toolkit). It makes it look like you're supposed to click it as soon as it shows up, given how prominent it is. Is an infoview button really needed for this, i.e., why is having access to the 'restart file' command not sufficient? If users aren't aware of it, could we simply change the message to point them to it?

Screenshot 2024-03-23 at 8 58 54 AM

mhuisi:
> It is really distracting and visually distinct from all the other buttons (which use codicons instead of the webview-ui-toolkit).

I don't consider all the other buttons in the InfoView integral for working with Lean - you can get along fine without knowing about them, and I'm sure that most users never use any of the codicon buttons specifically because they don't stick out much. "Restart File" however is central to working with Lean and you often cannot even get started with working on a file without using it. 

This kind of button shouldn't be hidden away behind an obscure keybinding, a menu with an obscure icon or an obscure feature like the command palette. It should be immediately visible and immediately accessible within a single click. 

> Is an infoview button really needed for this, i.e., why is having access to the 'restart file' command not sufficient?

For one, mouse-heavy users may find it annoying to use the command palette (you have to take your hand off the mouse to type in the command you want). This is important because "Restart File" is typically something you use while navigating code, not while writing code. The forall menu is easier to access, but it still requires multiple clicks to run the command, which I find especially annoying when using Lean with a touchpad. The keybinding is the best option even for mouse-heavy users (at least as long as your mouse is on the right side of the keyboard), but I suspect certain user groups of Lean (e.g. math undergrads) aren't used to heavy use of keybindings and may avoid it due to a general unfamiliarity with keybindings.

> If users aren't aware of it, could we simply change the message to point them to it?

I'm hesitant to have the language server emit diagnostics that are specific to the VS Code extension. Another option is to hack the InfoView to display a more specific message for the stale dependency diagnostics (perhaps even with a link to run the command), but this is very ugly to maintain, especially as Lean 4 evolves, so I decided against it. The stale dependency notifications do this, and I'm very unhappy with it. 

> [Screenshot]

I don't think the screenshot you posted is fair. Here's what it looks like on a regular 1920x1080 screen at 125% scaling (i.e. for use on a small 13 inch laptop screen):
image

I find the whole "Restart File" situation to be generally unfortunate from a UX standpoint especially for less technically inclined users, but there's sadly nothing we can do about it at this point.

Vtec234 (approved):
You mentioned that this will be urgently needed for the next Lean RC, and I don't want to block you on that. The implementation looks fine at a glance (I did not review in detail). I would midly prefer not to pull in a dependency on `webview-ui-toolkit` and just use plain CSS to style the button, but I will leave that decision to you.

> "Restart File" however is central to working with Lean and you often cannot even get started with working on a file without using it.
> I find the whole "Restart File" situation to be generally unfortunate from a UX standpoint especially for less technically inclined users, but there's sadly nothing we can do about it at this point.

Maybe the central issue that has to be addressed is not the button or lack thereof, but this point about the system's design. At various points we discussed re-adding an option for users to configure the server to behave a bit more like Lean 3 and auto-reload any file whose dependencies have changed. It is already computing stale dep notifications, so it sounds like that might not be too complex implementation-wise. The original issue in Lean 3, [noted on Lean.Server](https://github.com/leanprover/lean4/tree/master/src/Lean/Server#recompilation-of-opened-files), was that given a dep chain A->B, Lean 3 would refresh A as soon as B changed *even if B wasn't saved*. I think it would be completely okay to refresh only if the user saves B; this way accidental changes don't trigger recompilation, but the 'restart file' notifications/button is gone. What do you think? Why do you say nothing can be done?

mhuisi:
> It is already computing stale dep notifications, so it sounds like that might not be too complex implementation-wise.

It's worth pointing out that this is not perfect, since we use file watchers and `didSave` notifications to issue the diagnostic, so triggering a rebuild whenever users see the diagnostic will be too frequent. Only when opening the file do we actually query Lake to detect stale dependencies, since we don't want to poll Lake for every file worker. There are several ways to improve this, but each is currently too expensive to implement.

> I think it would be completely okay to refresh only if the user saves B; this way accidental changes don't trigger recompilation, but the 'restart file' notifications/button become less frequently needed.

As Sebastian mentioned above, we've seen a lot of feedback from Mathlib users that manually indicating that they want to trigger a rebuild is what they want, even when initially opening a file and well after saving a change.

> Why do you say nothing can be done?

Because we've tried more automatic approaches and people strongly expressed that they were unhappy with them. 

The fundamental problem is that getting the semantic information back in sync after a change is so expensive in Lean and that so many files need to be rebuilt entirely. If this could be improved and both cost and latency to get the language server back in sync were acceptable, then we could move to an interaction model that is closer to other language servers.

Vtec234:
> Automatically rebuilding saved files _was_ the previous Lean 4 behavior, people did _not_ like it. Thus the status quo.
> 
> edit: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Imports/near/268441788

> As Sebastian mentioned above, we've seen a lot of feedback from Mathlib users that manually indicating that they want to trigger a rebuild is what they want

I can see Gabriel expressing a preference there because he has `autosave` on, so that rebuild-saved-files is equivalent to rebuild-changed-files. Other [strongly prefer](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/import.20must.20be.20rebuilt/near/398728034) the other behaviour. One thing I missed in our discussion here is that there already is an option to auto-rebuild when opening a file! *But* there is no option to also auto-rebuild when a dependency changes and the file is already open. As you mention Marc, this could result in unnecessarily eager rebuilds, but I still think it would be great to have such an option (off by default).

Also, I am realizing now (after merge; sorry) that there is a regression in the UI: Lean no longer indicates that there exist dependencies that have changed. Could we have a followup PR making the `Restart File` button become yellow when this happens? This would be analogous to how other UI elements are yellow when something is loading.

mhuisi:
> Also, I am realizing now (after merge; sorry) that there is a regression in the UI: Lean no longer indicates that there exist dependencies that have changed. Could we have a followup PR making the Restart File button become yellow when this happens? This would be analogous to how other UI elements are yellow when something is loading.

The diagnostic should be visible under 'Messages' (modulo #424).

Vtec234:
Yep, it's just not clearly correlated with the button. But I remember you were reluctant to special-case the stale dependencies message, so let's leave the UI as-is for now.

Vtec234:
You're right, my bad.

mhuisi:
> I can see Gabriel expressing a preference there because he has autosave on, so that rebuild-saved-files is equivalent to rebuild-changed-files. Other [strongly prefer](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/import.20must.20be.20rebuilt/near/398728034) the other behaviour. One thing I missed in our discussion here is that there already is an option to auto-rebuild when opening a file! But there is no option to also auto-rebuild when a dependency changes and the file is already open. As you mention Marc, this could result in unnecessarily eager rebuilds, but I still think it would be great to have such an option (off by default).

I don't mind supporting an option for this. I'll put it on my to-do list :-)

PR #422, opened 2024-03-26
feat(abbreviations.json): add and change some abbreviations

fpvandoorn:
These are either the LaTeX name, the Mathlib name or a description of the symbol.

Note: this *changes* the following abbreviations:
* `\setminus` from `∖` to `\` (the latter is the ASCII symbol). The former is unused in Mathlib, except in docstrings using the symbol incorrectly to refer to the latter. The former is still available as `\smallsetminus`.
* `\subset` from `⊂` to `⊆`. The former is almost unused compared to the latter in Mathlib (240 vs 6900 hits). The former symbol is still available as `\ssub` or `\ssubset`. This should also make `\subs` autocomplete conveniently to `⊆` instead of `⊂`.
* `\supset` similarly from `⊃` to `⊇`, which are rarely used (former: 13 hits, but only in tests/delaborators, latter: 80 hits). 
* `\lf` is renamed to `\lfuzzy`, so that `\lf` expands to `\lfloor`.
* Add by symmetry:
```
    "rfuzzy": "⧐",
    "|>": "⧐",
``` 

Also reorder some abbreviations, to make `mapsto` have priority over other partial matches.


fpvandoorn:
Can we add a comment somewhere what algorithm is used to complete a partial match? Is it the first completion in the json file? 
I moved "mapsto" above "male", which will make that a higher precedence if that is indeed the heuristic.

vscode-lean4/src/abbreviation/abbreviations.json
```diff
@@ -1361,6 +1381,7 @@
     "P0": "Π₀",
     "Pi_0": "Π₀",
     "P_0": "Π₀",
+    "directsum": "Π₀",
```
eric-wieser, line 1384: This one is at odds with mathlib, which uses a different symbol for direct sums.

fpvandoorn: Oops, I copied from the definition of `DirectSum` instead of the notation line.

vscode-lean4/src/abbreviation/abbreviations.json
```diff
@@ -14,6 +14,8 @@
     "[--]": "⁅$CURSOR⁆",
     "nnnorm": "‖$CURSOR‖₊",
     "norm": "‖$CURSOR‖",
+    "floor": "⌊$CURSOR⌋",
+    "ceil": "⌈$CURSOR⌉",
```
eric-wieser, line 18: Should we have `nfloor` and `nceil` for the `₊` versions (`Nat.floor` and `Nat.ceil`)?

fpvandoorn: That is a good idea.

mhuisi:
I can't say whether the Mathlib community in general agrees with these changes to the abbreviations. Could you make a thread on the Lean Zulip with a summary of the symbols with changed abbreviations to ensure that everyone agrees with them?

vscode-lean4/src/abbreviation/abbreviations.json
```diff
@@ -761,7 +781,7 @@
     "sphericalangle": "∢",
     "section": "§",
     "searrow": "↘",
-    "setminus": "∖",
+    "setminus": "\\",
```
eric-wieser, line 784: This one feels slightly weird to me; we may not (currently) use this character in mathlib, but this really is "setminus"!

fpvandoorn: I think it's most important that `\setminus` expands to Mathlib's character for set differences, even if that character is "wrong". I don't actually care which symbol is used for set differences. If you're willing to change the character used in Mathlib, I'm happy to revert this (in a follow-up PR).

fpvandoorn:
@mhuisi I should have linked that thread in this PR, I forgot. I think the discussion in that thread is positive (or suggestions that I've incorporated). I'm happy to write a follow-up PR if there is unexpected consequences/outcry (which I don't expect).

vscode-lean4/src/abbreviation/README.md
```diff
@@ -8,6 +8,11 @@ Use `$CURSOR` to set the new location of the cursor after replacement.
 If VS Code adds certain characters automatically (like `]` after typing `[`),
 ensure that each such subword is a strict prefix of another abbreviation.
 
+## Manual replacements
+
+When manually replacing an abbreviation early (using `Tab` by default), the abbreviation feature will pick the shortest abbreviation with a matching prefix.
```
fpvandoorn, line 13: I didn't know that `Tab` replaces an abbreviation! I use space, and regularly use space-backspace if I don't want the space.

mhuisi: Early completions using space should apply the same logic, I believe.

fpvandoorn: Yeah, that sounds plausible. Tab is really nice though.

mhuisi:
Is this PR ready to go?

fpvandoorn:
Yes

PR #428, opened 2024-04-05
feat: support for multi-toolchain workspaces

### LINKED ISSUE #138: Toolchain version is not properly detected within subfolders
To reproduce, set your default version of Lean in `elan` to A. For example:
```shell
> elan toolchain list
leanprover/lean4:nightly-2021-12-28 (default)
leanprover/lean4:nightly-2022-01-23
```

Then make a project with the following structure:
```shell
> tree foo/
foo
└── bar
    ├── Bar.lean
    ├── lakefile.lean
    └── lean-toolchain

3 directories, 5 files
```
where `lean-toolchain` specifies version B (e.g. `leanprover/lean4:nightly-2022-01-23`) and `Bar.lean` contains
```lean
#eval Lean.versionString
```

Finally open VSCode with `foo/` as the top-level folder. I would expect the interactive version in VSCode to be B as per `lean-toolchain`, but in fact the default version (A) is executed. This leads to discrepancy between command-line builds of `Bar` and VSCode.

A concrete reason for why I think we should handle this case properly is to support multi-language projects containing Lean packages together with perhaps ones in other languages (e.g. `foo/myJsProject`), where the top-level is opened as a workspace.

mhuisi:
This PR adds support for using multiple lean toolchains in the same workspace folder. It also enables users to use Lean 4 without first having to open a folder.

For nested workspaces, the client uses the outer-most Lean project. This is the correct behavior for projects in `.lake`, but not the correct behavior for actual nested projects. I unfortunately do not see a way to make the latter work because the language client library only supports globs to specify which files should be picked up by a language server, and hence cannot express something like "all Lean files in this directory, except those contained in inner directories with their own lean-toolchain". We could in principle set up middleware for every single notification and request, but this is likely to be overly fragile.

Closes #138.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ you to install lean via elan.  If the folder contains a `lean-toolchain` version
 otherwise it will install `leanprover/lean4:stable`.
 You can override this default version by setting an `DEFAULT_LEAN_TOOLCHAIN` environment variable.
 
-Otherwise, if there is a `lean-toolchain` (or `leanpkg.toml`) in the workspace folder or in a parent
+Otherwise, if there is a `lean-toolchain` in the workspace folder or in a parent
 folder then it will use the version specified in the specified version.
 
 Then with the selected version it runs `lean --version` to check if that version is installed yet.
--- a/vscode-lean4/src/docview.ts
+++ b/vscode-lean4/src/docview.ts
@@ -5,36 +5,12 @@ import { URL } from 'url';
 import { commands, Disposable, Uri, ViewColumn, WebviewPanel, window,
      workspace, WebviewOptions, WebviewPanelOptions, TextDocument, languages,
      Range, Position } from 'vscode';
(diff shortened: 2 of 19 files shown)


david-christiansen:
It would be extremely convenient for me to have support for nested folders with different toolchain versions.

Would it be a suitable workaround to symlink to the inner directory from somewhere else, then add that to the workspace?

mhuisi:
> It would be extremely convenient for me to have support for nested folders with different toolchain versions.

Do these different Lean projects depend on one another in some way?

mhuisi:
It looks like the language client library recently added support for general notification and request middleware. I may take another stab at nested toolchains with that in mind.

david-christiansen:
(for anyone following along, we spoke in person today and it'll be enough for the projects to be peer directories in the same Git repo, which will work for my use case)

PR #430, opened 2024-04-09
chore: add and run prettier

### LINKED ISSUE #38: Use Prettier for formatting
We should format the whole repository with [Prettier](https://prettier.io/) on pre-commit and forget about style from then on.

mhuisi:
- Adds Prettier and a pre-commit hook for ESLint & Prettier.
- Formats the entire code base with Prettier.
- Adds the formatting commit to a `.git-blame-ignore-revs` file so that `git blame` remains useful for the Git history before the formatting commit. GitHub should pick up this file automatically, whereas `git config blame.ignoreRevsFile .git-blame-ignore-revs` needs to be run once locally so that the file is picked up.
- Removes the legacy `old` directory that was a left-over from the Lean 3 VS Code extension.

Closes #38.


mhuisi:
@Vtec234 Please let me know what you think of the formatted InfoView code.

lean4-infoview/src/index.tsx
```diff
@@ -1,42 +1,51 @@
...
 
     if (interactive.state === 'resolved') {
         return <InteractiveMessage fmt={interactive.value} />
     } else if (interactive.state === 'loading') {
         return <>...</>
     } else {
-        return <div>Failed to display message:
-            {<span>{mapRpcError(interactive.error).message}</span>}
-        </div>
+        return (
```
Vtec234, line 44: Is there an option to start on the `return` line like we did before?

Vtec234 (approved):
One question but looks good either way.

lean4-infoview/src/index.tsx
mhuisi: I don't think so. The only relevant JSX setting I can see is https://prettier.io/docs/en/options#bracket-line

PR #431, opened 2024-04-10
Log errors if something goes wrong with the lake binary

eric-wieser:
When using a custom `lakePath`, it is helpful to be able to see the error messages if something goes wrong.

### DIFF
--- a/vscode-lean4/src/leanclient.ts
+++ b/vscode-lean4/src/leanclient.ts
@@ -487,6 +487,7 @@ export class LeanClient implements Disposable {
         const start = Date.now()
         const result: ExecutionResult = await batchExecute(executable, versionOptions, this.folderUri?.fsPath);
         if (result.exitCode !== ExecutionExitCode.Success) {
+            logger.error(`[LeanClient] Ran '${executable} ${versionOptions.join(' ')}', got error:\n{result.stderr}`);
             return false
         }
         logger.log(`[LeanClient] Ran '${executable} ${versionOptions.join(' ')}' in ${Date.now() - start} ms`);


vscode-lean4/src/leanclient.ts
```diff
@@ -487,6 +487,7 @@ export class LeanClient implements Disposable {
         const start = Date.now()
         const result: ExecutionResult = await batchExecute(executable, versionOptions, this.folderUri?.fsPath);
         if (result.exitCode !== ExecutionExitCode.Success) {
+            logger.error(`[LeanClient] Ran '${executable} ${versionOptions.join(' ')}', got error:\n{result.stderr}`);
```
mhuisi, line 490: Are you sure that this is what you want? Where do you expect to see this error?

The VS Code extension uses `logger` calls purely for debugging purposes. They show up in the VS Code debug console when debugging the extension.

eric-wieser: > Where do you expect to see this error?

In the debug console. I already am able to see the log message below about things succeeding. If we log when they succeed, then surely we should also log when they fail!

I don't really care where the error appears; just that it appears somewhere so that it can be found. The context here is that my custom `lakePath` may have failed to startup with a message, and I want to be able to see that.
I think making this a popup message is overkill, and is likely to be confusing for users who haven't installed lake at all.

eric-wieser: > They show up in the VS Code debug console when debugging the extension.

Looking again, I think these used to appear in a the "output pane" under "Server"; but looking again, the "Server" option seems to have disappeared from the dropdown, presumably due to a vscode update.

mhuisi: I'm happy to merge this, but it may take a while until it is released in full since we are currently doing some pre-release testing.

eric-wieser: Here's where this message would appear:

image

PR #432, opened 2024-04-18
fix: slow infoview

mhuisi:
Fixes a performance regression reported at https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Infoview.20lag/near/433818006 and introduced in #408.

In doing so, it replaces the fading error message tooltip with one that can be closed by clicking anywhere.


lean4-infoview/src/infoview/tooltips.tsx
```diff
@@ -71,57 +70,43 @@ export function Tooltip(props_: TooltipProps) {
...
-          {tooltipChildren}
-        </Tooltip>
-      </span>}
-    </Transition>
-    {props.children}
-  </span>)
+  return <LogicalDomContext.Provider value={logicalDomStorage}>
+      <span
+        ref={setRef}
+        onClick={() => { if (!window.getSelection()?.toString()) hideTooltip() } }
```
Vtec234, line 100: This has the effect that a selection within the error tooltip keeps it open, but a click there hides it. If that's intended, great; otherwise you can use `isWithinHoverable` as in `WithTooltipOnHover`.

Vtec234 (approved):
There is some duplication that could perhaps be refactored into a common tooltip component, but it's not too bad. Thanks!

lean4-infoview/src/infoview/tooltips.tsx
mhuisi: Yes, the intended behavior is that you can still copy and paste the error message :-)

Issue #197, opened 2022-06-17
Infoview DocumentPosition should be a context


EdAyers:
Currently in the infoview code, every component is passing the `pos : DocumentPosition` prop up through the component hierarchy. I think it would make a lot of sense to make this a React context instead.

gebner:
I'm not sure if this is such a good idea.  We have a lot of different positions floating around.
1. The cursor position in the editor.
2. The position of the current info (which may be different from the cursor position if it's pinned).
3. The position of the data shown in the info (which may lag behind the info position when the cursor is moved).
4. The position used for RPC queries.
5. The position of the diagnostic message / term goal / widget.

It goes without saying that all of the can be different.  I vaguely recall that we've had some issues mixing up 2 and 3 but I can't find it right now.

lovettchris:
I agree with Gabriel, I think "pinning" shows an example where a react component needs it's own local "pos" that is not in a shared context.  You see this in Infos.tsx where the pos being passed down to the <Info> component is conditionally coming from  the `curLoc` state or the `pinnedPoss` state. Infos is also listening to the `ec.events.changedCursorLocation` event so it can update when the current location is changed.  `AllMessages` is also an example of a React component in the info view that does *not* use the cursor location at all.

![image](https://user-images.githubusercontent.com/18707114/174671589-9b40ba61-22af-4f89-88ff-fb21e065e8a3.png)

I am curious about the naming of `pinnedPoss` and setPinnedPoss`.  I would have named these `pinnedPos` and setPinnedPos`.  I'm not sure what the second `s` at the and of those names represents exactly... it's not like it is plural positions or anything... so a bit confusing to me?  Or is it plural because of the `pinKey.current` refcount?

EdAyers:
Yes there are different `pos`es flying around. I still think it's easier to manage these with contexts. Once you are inside the body of `InfoDisplay` there is only one pos. so the proposal would be to add pos as a context there.

gebner:
> once you are inside the body of `InfoDisplay` there is only one pos

Inside of `InfoDisplay` you still have the positions (2)-(5) from my list.

PR #449, opened 2024-05-22
Widget messages

Vtec234:
Implementation of leanprover/lean4#2064.

Besides adding support for widget messages, this adds an `EnvPosContext` that stores the position that a component is "about" (see the docstring for more detail). This was originally proposed in #197 and eventually dismissed (see #226 for a detailed discussion), but I think it warrants a reconsideration. While Gabriel's point that there are many positions around remains true, the `EnvPosContext` stores a very specific one. This position is only used to retrieve a snapshot, and from that a `Lean.Environment`, on the server-side. Widgets currently have to pass this position around manually; this has led to [errors](https://github.com/leanprover-community/ProofWidgets4/pull/60/commits/ca66ebb68ab091469770b0dc80a993932c915dbf) in user code. Furthermore, so far I have not encountered any situation where it makes sense to modify the position; thus a context seems like a preferable approach. If needed, the context can still be set by users.

I also had to update some package versions to have tests building again. I hope that didn't break anything!


lean4-infoview/src/infoview/rpcSessions.tsx
```diff
@@ -43,18 +46,38 @@ export function WithRpcSessions({ children }: { children: React.ReactNode }) {
...
+    return React.useContext(RpcSessionsContext)?.connect(pos) || noCtxRpcSession
 }
 
 export function useRpcSessionAtPos(pos: DocumentPosition): RpcSessionAtPos {
     return useRpcSessionAtTdpp(DocumentPosition.toTdpp(pos))
 }
 
-export const RpcContext = React.createContext<RpcSessionAtPos>(fakeRpcSession)
+/** @deprecated use {@link useRpcSession} instead */
+export const RpcContext = React.createContext<RpcSessionAtPos>(noCtxRpcSession)
```
mhuisi, line 70: What is the path to fully replacing this with `useRpcSession`? Why is it still necessary to use `RpcContext` in the code base?

vscode-lean4/package.json
```diff
@@ -892,31 +892,31 @@
         "@leanprover/infoview-api": "~0.4.0",
         "axios": "^1.6.7",
         "cheerio": "^1.0.0-rc.12",
-        "mobx": "^5.15.7",
+        "mobx": "^6.12.3",
```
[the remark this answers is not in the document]
Vtec234: Thanks for testing. I restored the previous version.

vscode-lean4/test/suite/index/index.ts
```diff
@@ -21,23 +21,19 @@ export function run(testsRoot: string, cb: (error: any, failures?: number) => vo
 
     logger.log('>>>>>>>>> testsRoot=' + testsRoot)
 
-    glob('**/**.test.js', { cwd: testsRoot }, (err, files) => {
-        if (err) {
-            return cb(err)
-        }
+    const files = glob.sync('**/**.test.js', { cwd: testsRoot })
```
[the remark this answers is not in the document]
Vtec234: I have no clue! The `glob` docs don't say anything about exceptions, and I don't understand the test code. I moved this into the `try/catch` block below.

lean4-infoview/src/infoview/rpcSessions.tsx
Vtec234: Existing widgets use the `RpcContext`, so for backwards compatibility we must set it to the correct value. It could potentially be removed later on in the future.

mhuisi: Ah, that makes sense. Thanks!

mhuisi (approved):
Over all, these changes look good to me now. I wasn't involved in the original conversation, but I do wonder whether it would make sense to have separate contexts for the different use cases of `PosContext` in case these won't overlap as nicely as they currently do in the future, or just to make the type name better describe which position is actually provided by the context.

Vtec234:
> I do wonder whether it would make sense to have separate contexts for the different use cases of `PosContext` in case these won't overlap as nicely as they currently do in the future

`PosContext` is supposed to store positions used for the specific purpose of resolving a server snapshot, from which an environment, from which and RPC method/widget module/any other kind of annotated definitions. The context is not meant to store positions aimed to be used for anything else (e.g. editor positions at which we insert a tactic script in "Try this"). My reason for introducing `PosContext` is that those environment-resolving positions do not tend to vary across component hierarchies in widgets (they do vary in the infoview  itself), don't have to be super precise (in contrast to editor positions at which we insert a tactic script, say), and having to manage them manually has led to user error.

> make the type name better describe which position is actually provided by the context

This could be a good idea; any suggestions?

mhuisi:
Hmm, the only reasonable name I can think of is something like `AnchorPosContext` since the position describes where a specific component is anchored in the source code.

Vtec234:
Okay, renamed to `EnvPosContext` and simplified docstring.

PR #453, opened 2024-05-23
feat: manual

mhuisi:
This PR adds an extensive manual for the Lean 4 VS Code extension and central parts of the Lean 4 language server. It also covers general VS Code features that Lean users who have not used VS Code before may find to be helpful when working with Lean.

There are still some obvious improvements to the manual view that would be nice to have, but I hope that this is at least a good start.

ToC


vscode-lean4/media/manual.md
```diff
@@ -0,0 +1,631 @@
...
+When making a change to a file, Lean needs to re-process all declarations that can depend on the declaration that the change was made to. Currently, this means that all code below the point where the change occured needs to be re-processed.
+
+For parts of the file that are currently still being processed, Lean cannot provide any interactive features that require file processing information yet. For example, in unprocessed parts of the file, [errors, warnings and information](#errors-warnings-and-information) are not displayed, the [InfoView](#infoview) will not show the current goal state and [hovers](#hovers) will not display a popup until file processing has progressed to the respective part of the file.
+
+| &nbsp; |
+| :--: | 
+| ![](media/file-progress.png) | 
+| *File processing progress. The orange bar on the right displays the progress for the whole file, whereas the one on the left displays the progress for the currently visible source code lines.* |
+
+### Errors, warnings and information
```
TwoFX, line 136: The [Error Lens](https://marketplace.visualstudio.com/items?itemName=usernamehw.errorlens) extension is quite popular with Lean users. Maybe it makes sense to mention it in this section?

TwoFX:
There is a section "Using Lean 4 in VS Code" in the setup guide, which is a bit like an extremely condensed version of this manual. I think that the setup guide should contain a link to the manual somewhere.

vscode-lean4/media/manual.md
```diff
@@ -0,0 +1,631 @@
...
+1. **['Project: Fetch Mathlib Build Cache'](command:lean4.project.fetchCache)**. Downloads and installs the current Mathlib build artifact cache if the project is [Mathlib](https://github.com/leanprover-community/mathlib4) or depends on it.
+
+| &nbsp; |
+| :--: | 
+| ![](media/update-dependency.png) | 
+| *'Project: Update Dependency…' selection dialog* |
+
+### Terminal
+
+If the commands provided by the Lean 4 VS Code extension are not sufficient to manage the Lean project, then a built-in VS Code terminal can be launched by calling the ['Terminal: Create New Terminal'](command:workbench.action.terminal.new) command or by pressing ```Ctrl+Shift+```` (```Cmd+Shift+````). In this terminal, all [Lake](https://github.com/leanprover/lean4/blob/master/src/lake/README.md) commands can be executed.
```
TwoFX, line 556: This does not render correctly: ```Ctrl+Shift+```` (```Cmd+Shift+````)

vscode-lean4/media/manual.md
```diff
@@ -0,0 +1,631 @@
...
+In order to automatically trigger rebuilds of all changed dependencies when opening a file instead of issuing an error, the 'Lean 4: Automatically Build Dependency' setting can be enabled.
+
+| &nbsp; |
+| :--: | 
+| ![](media/restart-file-infoview-button.png) | 
+| *'Restart File' button in InfoView* |
+
+### Unicode input
+
+Lean code uses lots of [Unicode symbols](https://home.unicode.org/) to aid readability. In Lean files, these Unicode symbols can be entered by typing a backslash followed by an abbreviation identifier referring to the symbol. As soon as the abbreviation identifier is unique, the abbreviation will be automatically replaced with the corresponding Unicode symbol - for example, `\forall` will yield `∀`.
```
TwoFX, line 178: Is the last sentence correct? Isn't the prefix `foral` already unique? To me, it seems like the automatic replacement triggers only once the full identifier has been entered.

vscode-lean4/media/manual.md
```diff
@@ -0,0 +1,631 @@
...
+| &nbsp; |
+| :--: | 
+| ![](media/restart-file-infoview-button.png) | 
+| *'Restart File' button in InfoView* |
+
+### Unicode input
+
+Lean code uses lots of [Unicode symbols](https://home.unicode.org/) to aid readability. In Lean files, these Unicode symbols can be entered by typing a backslash followed by an abbreviation identifier referring to the symbol. As soon as the abbreviation identifier is unique, the abbreviation will be automatically replaced with the corresponding Unicode symbol - for example, `\forall` will yield `∀`. 
+
+To replace an abbreviation early, before it is unique, `Tab` can be pressed to trigger the ['Lean 4: Input: Convert Current Abbreviation'](command:lean4.input.convert) command. This will yield the Unicode symbol with the shortest abbreviation identifier.
```
TwoFX, line 180: I think it should also be mentioned that space and enter (and also right-arrow, and maybe some other keys I don't know about) also cause the identifier to be replaced.

mhuisi:
> There is a section "Using Lean 4 in VS Code" in the setup guide, which is a bit like an extremely condensed version of this manual. I think that the setup guide should contain a link to the manual somewhere.

This PR already changed that, though?

TwoFX:
> This PR already changed that, though?

Oops, sorry, it looks like I looked at a VS Code window that still had the previous version of the extension.

vscode-lean4/media/manual.md
mhuisi: I've tested it out for a couple of months, but the main drawback I found was that most Lean errors are typically formatted to span multiple lines, so Error Lens is rarely helpful to actually understand the issue. I can still add it, though.

vscode-lean4/media/manual.md
mhuisi: Oops.

PR #458, opened 2024-05-27
feat: auto-publish infoview

joneugster:
Automatically publish infoview and infoview-api on npmjs if the version changed.


joneugster:
This is a first draft, containing the modifications to publish the infoview from CI.

One needs to create an NPM access token and add it as github secret `NPM_TOKEN`.

This does not yet contain any version number management.

I have a hard time testing this, only did so for `@leanprover/infoview` but not for the `infoview-api` yet.

joneugster:
@mhuisi I think this is now one of the simplest versions I could think of. on a new release, if the package version of `infoview` or `infoview-api` changed, it will publish the new version to npmjs.

Using e.g. `npm --workspace=lean4-infoview version patch` to bump the version is still required to happen manually.

It does not necessarily need to be coupled to the release schedule of the VSCode-extension, but I figured doing so ensures that only maintainers will trigger the publishing (at defined steps of the procedure) and not by accident.

.github/workflows/on-push.yml
```diff
@@ -43,6 +43,18 @@ jobs:
           npm ci
           npm run build
 
+      - name: Publish infoview-api (if new version)
+        if: ${{ startsWith(github.ref, 'refs/tags/v') && !endsWith(github.ref, '-pre') && matrix.os == 'ubuntu-latest' }}
+        continue-on-error: true
+        run: |
+          [ "$(npm view @leanprover/infoview-api version)" != "$(sed -n 's/^\s*"version":\s*"\(.*\)",\s*/\1/p' lean4-infoview-api/package.json)" ] && (npm publish --workspace=lean4-infoview-api --access=public) || (echo "infoview-api version unchanged")
```
mhuisi, line 50: Is this `sed` approach really necessary? Won't publishing the InfoView simply fail if the version is unchanged?

joneugster: It does fail yes, so this is superfluous. It only shortens the CI output on failure.

mhuisi: In that case, I would prefer a solution without the `sed` invocation :-)

joneugster: just removed it :)

lean4-infoview/package.json
```diff
@@ -6,7 +6,7 @@
         "watch": "rollup --config --environment NODE_ENV:development --watch",
         "build": "rollup --config --environment NODE_ENV:production",
         "watchTest": "npm run watch",
-        "test": "tsc -p test/tsconfig.json"
+        "test": "./test/run-tests.sh"
```
mhuisi: It should be possible to run the tests on Windows.

joneugster: I moved that to an independent PR

Vtec234 (approved):
Haven't tested it but looks good in theory!

PR #464, opened 2024-06-05
Widget messages but hovers also work now

Vtec234:


package-lock.json
```diff
@@ -58,52 +57,259 @@
             }
         },
         "lean4-infoview-api": {
-            "name": "@leanprover/infoview-api",
```
mhuisi, line 61: This probably shouldn't be removed.

package-lock.json
```diff
@@ -11,21 +11,20 @@
...
                 "lerna": "^8.1.2",
                 "lint-staged": "^15.2.2",
-                "prettier": "3.2.5",
+                "prettier": "^3.2.5",
                 "prettier-plugin-organize-imports": "^3.2.4",
                 "typescript": "^5.4.5"
             }
         },
         "lean4-infoview": {
-            "name": "@leanprover/infoview",
```
mhuisi, line 28: This probably shouldn't be removed.

package-lock.json
llllvvuu: If I run `npm install` on your branch it adds them back. My setup:

```
; node --version                                                         
v20.12.2
; npm --version          
10.5.0
```

mhuisi (approved):

package-lock.json
Vtec234: Yeah, not sure why my NPM keeps removing these.

Vtec234: It seems that `npm update` removes them, and `npm install` readds them.

PR #483, opened 2024-06-22
feat(abbreviations.json): Add abbreviation for Spec -> `𝖲𝗉𝖾𝖼`

erdOne:
Corresponding to the new notation introduced in https://github.com/leanprover-community/mathlib4/pull/14025


mhuisi:
Please discuss this notation on Zulip first. I'm not sure if we really want to add abbreviations that are almost indistinguishable from regular editor text, but I'm happy to merge this if the Mathlib community agrees that this is the notation they want.

erdOne:
Closing this as the notation is removed.

PR #495, opened 2024-07-06
feat: [lean4web] abstract the implementation of infoview as a webview

abentkamp:
This is a pure refactoring. The behavior of the code has not changed. But this setup allows us to inject an iframe infoview implementation instead of the webview implementation.


joneugster:
recreated as #567

PR #511, opened 2024-07-26
fix: filter menu button collapsing goal state

### LINKED ISSUE #499: Pinned locations are tracked incorrectly
### Description

Adding a newline in the source should move pinned locations below that line down by 1. Under certain circumstances (which I haven't attempted to identify but which seem to occur fairly often), the pinned locations move by 2 instead. This completely defeats the purpose of pinning to see the results of edits made above in realtime. Note that this used to work correctly, so it must have been a recent regression.

### Steps to Reproduce

1. Pin a location, for example `Test.lean:100:2`.
2. Add a newline above it.
3. Pinned location is now `Test.lean:102:2`.

**Expected behavior:** Pinned location updates to `Test.lean:101:2`.

**Actual behavior:** As above.

### Versions

`vscode-lean4 v0.0.171`
`leanprover/lean4:4.10.0-rc1`

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/vscode-lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

### LINKED ISSUE #505: Tactic state collapses when filters are opened
### Description

As seen in the recording, opening the filters tooltip collapses the tactic state.

https://github.com/user-attachments/assets/ac35b200-8692-44b1-8928-b91790b763fc

**Expected behavior:** Opening the tooltip does not (un)collapse the tactic state.

**Actual behavior:** As above.

### Versions

`vscode-lean4 v0.0.174`

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/vscode-lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

mhuisi:
Closes #505.

### DIFF
--- a/lean4-infoview/src/infoview/collapsing.tsx
+++ b/lean4-infoview/src/infoview/collapsing.tsx
@@ -32,33 +32,23 @@ interface DetailsProps {
     setOpenRef?: (_: React.Dispatch<React.SetStateAction<boolean>>) => void
 }
 
-/** Like `<details>` but can be programatically revealed using `setOpenRef`. */
+/** Like `<details>` but can be programatically revealed using `setOpenRef`.
+ * The first child is placed inside the `<summary>` node. */
 export function Details({ initiallyOpen, children: [summary, ...children], setOpenRef }: DetailsProps): JSX.Element {
     const [isOpen, setOpen] = React.useState<boolean>(initiallyOpen === undefined ? false : initiallyOpen)
-    const setupEventListener = React.useCallback((node: HTMLDetailsElement | null) => {
-        if (node !== undefined && node !== null) {
-            // Prevents the native click event from firing and opening/closing the tag.
-            // This is necessary because we do not want the `details` tag to react to
-            // clicks when we call `e.stopPropagation()` in a synthetic React click event further down in the DOM,
-            // since synthetic React events are only executed after the corresponding native event
-            // has already fully bubbled up the DOM, and so a synthetic React event cannot stop the propagation
-            // of the corresponding native event.
-            node.addEventListener('click', e => {
-                e.preventDefault()
-            })
-        }
-    }, [])
     if (setOpenRef) setOpenRef(setOpen)
     return (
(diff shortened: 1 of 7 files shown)


Vtec234:
As far as I can tell, the underlying issue is not really about native vs synthetic React events (BTW, do you have a source for docs about this?). It is more simply that there were a couple of places where we call `stopPropagation` but not `preventDefault`. This means that handlers higher up do not get called, but the default event behaviours still occur. You can see it in this example app:
```typescript
export default function App() {
  return (
    <details>
      <summary onClick={e => e.preventDefault()}>
        <a onClick={e => e.stopPropagation()}>EFFECT</a>
        <span> - </span>
        <a>NO EFFECT</a>
      </summary>
      Content
    </details>
  )
}
```

So, I propose an alternative fix where we make sure that whenever `stopPropagation` is called, so is `preventDefault`. I made this change systematically, and it seems to work (but more testing would be good). I also changed the semantics of `Details` a bit to inject the `<summary>` because we should really install the `onClick` handler there rather than on `<details>`. I pushed the changes here, but feel free to revert if something is not working.

mhuisi:
I'm okay with you taking over this PR. Please proceed as you see fit. Your changes look good to me but I haven't tested them in detail.

> BTW, do you have a source for docs about this?

https://stackoverflow.com/questions/52444941/when-does-react-create-syntheticevents (and [several others](https://stackoverflow.com/a/24421834), though I can't find documentation of it in the React docs). 

I agree that this is due to a lack of `preventDefault` and not a native browser event firing - I had a misconception about the way the `<details>` toggle works here and how `preventDefault` and `stopPropagation` interact. I think my version worked because it uses a native browser event to call `preventDefault` and is hence not stopped by the `stopPropagation` calls in the synthetic events that we use everywhere else.

PR #519, opened 2024-08-26
chore: use emoji-variant for ❌️

joneugster:
Based on leanprover/lean4#5015, use emoji-variant selector for ❌️ across the Lean universe.


joneugster:
Note that this only replaces `\u{274C}` with `\u{274C}\u{FE0F}` (❌️)

PR #520, opened 2024-08-27
chore: use text variant of ✝︎

joneugster:
following leanprover/lean4#5174 use text-variant selector for ✝︎.

### DIFF
--- a/lean4-infoview/src/infoview/goals.tsx
+++ b/lean4-infoview/src/infoview/goals.tsx
@@ -18,7 +18,7 @@ import { useEvent } from './util'
 
 /** Returns true if `h` is inaccessible according to Lean's default name rendering. */
 function isInaccessibleName(h: string): boolean {
-    return h.indexOf('✝') >= 0
+    return h.indexOf('✝︎') >= 0
 }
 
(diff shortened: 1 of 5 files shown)


lean4-infoview/src/infoview/goals.tsx
```diff
@@ -18,7 +18,7 @@ import { useEvent } from './util'
 
 /** Returns true if `h` is inaccessible according to Lean's default name rendering. */
 function isInaccessibleName(h: string): boolean {
-    return h.indexOf('✝') >= 0
+    return h.indexOf('✝︎') >= 0
```
nomeata, line 21: Is this expected to work with previous versions of lean as well? Then maybe leave code as it (matches both) and leave a comment

mhuisi: Yes, vscode-lean4 needs to be backwards compatible with all prior Lean 4 releases.

joneugster: ofc that was an oversight!

Anyways, I closed https://github.com/leanprover/lean4/pull/5174 as I can't say with certainty that it won't introduce maintaince burden.

PR #548, opened 2024-11-16
moogle 2.0 + leandocsearch

jesse-michael-han:
Video demonstrating new UI / features:

https://github.com/user-attachments/assets/3685f63c-c519-42f5-8f62-3978c231e60b

### DIFF
--- a/vscode-lean4/loogleview/static/index.css
+++ b/vscode-lean4/loogleview/static/index.css
@@ -95,3 +95,22 @@ vscode-button {
 #suggestions a {
     text-decoration: var(--text-link-decoration);
 }
+
+.code-block {
+    background-color: #f6f8fa;
+    border: 1px solid #d0d7de;
+    border-radius: 6px;
+    margin: 1em 0;
+    padding: 16px;
+    overflow-x: auto;
+}
+
+.monospace {
+    font-family: monospace;
+    white-space: pre;
+}
+
+pre.monospace {
+    white-space: pre;
+    overflow-x: auto;
+}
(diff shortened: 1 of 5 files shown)


mhuisi (changes requested):
Other comments:
- Entering `&a` into the query field leads to a `TypeError: hits.map is not a function` error that is not cleared when sending a new query.
- The header of each doc search entry should tell users which documentation the entry is from, and ideally even what Lean version those docs are for.
- Doc quotes should not use a monospace font.
- In the theorem search, the start of the doc comment appears to be cut off (i.e. it displays as `-`).
- The UI should be explicit about which version of Mathlib it is targeting.
- Most doc searches yield very broken formatting. E.g. when searching for "difference between theorem and lemma", the first entry immediately has very broken formatting.
- Entering "introduction" into the doc search produces the TPIL ToC as a result, including links that don't work.
- Searching for "what does the infoview debounce setting do" in the doc search and clicking the "Introduction" entry yields an empty result.
- Most FPiL entries contain broken tags, e.g. `{{#example_in Examples/Intro.lean NatDoubleFour}}`.
- When searching for "what is mathlib" and clicking the "The Standard Library" entry, the "view online" link leads to a 404.
- The spinner should be directly next to the input container.
- I'd place the "Searching" widget below the search box, not next to it. It should also not be indented relative to the input box when the line needs to be broken in two.
- When clicking the search mode toggle, the change from a regular font to a bold font moves the toggle and the "Docs" text either to the left or the right. It should be static when clicking this button.

Broken doc formatting

PR #556, opened 2024-12-09
Lean4 code snippets

dalps:
These are some code snippets I defined to help me write Lean code, following the example of the [ocaml-platform](https://github.com/ocamllabs/vscode-ocaml-platform/blob/master/snippets/ocaml.json) extension.

From VS Code user guide's [article on snippets](https://code.visualstudio.com/docs/editor/userdefinedsnippets):
> Code snippets are templates that make it easier to enter repeating code patterns, such as loops or conditional-statements.

I think the Lean programming experience on VS Code could really benefit from some of these snippets.

### DIFF
--- a/snippets/lean4.json
+++ b/snippets/lean4.json
@@ -1,0 +1,111 @@
+{
+	"namespace": {
+		"prefix": "namespace",
+		"body": [
+			"namespace ${1:name}",
+			"$0",
+			"end $1"
+		]
+	},
+	"section": {
+		"prefix": "section",
+		"body": [
+			"section ${1:name}",
+			"$0",
+			"end $1"
+		]
+	},
+	"anonymous section": {
+		"prefix": "section",
+		"body": [
+			"section",
+			"$0",
+			"end"
+		]
+	},
+	"match .. with": {
+		"prefix": "match",
+		"body": [
+			"match ${1:scrutinee} with",
+			"| ${2:pattern} => ${0:value}"
+		]
+	},
+	"inductive type": {
+		"prefix": "inductive",
+		"body": [
+			"inductive ${1:name} where",
+			"\t| ${0:constructor} : $1"
+		]
+	},
+	"structure definition": {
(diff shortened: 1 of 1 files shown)


dalps:
If I can add my two cents, as an OCamler starting out with Lean, I'm used to ocaml-platform showing me a list of modules and functions from the stdlib or the packages I included in my dune project matching the letter I just typed; this happens wherever my cursor happens to be in an .ml file. I'm really missing this kind of assistance from the Lean extension.

In a Lean file, most of the time I type something I get a list of irrelevant workspace suggestions (those with the 'abc' icon), which I think is more annoying than accepting relevant suggestions by mistake on enter.

Hence these snippets, to fill that void when the Lean language server has nothing better to show.

mhuisi:
> In a Lean file, most of the time I type something I get a list of irrelevant workspace suggestions (those with the 'abc' icon), which I think is more annoying than accepting relevant suggestions by mistake on enter.

These are provided by VS Code, not vscode-lean4. You can disable them using the 'Editor: Word Based Suggestions' VS Code setting. We've considered disabling those language-wide in the past, but it turns out that some users do like word based suggestions, and VS Code makes it very difficult to change language-defaults, so we've decided against it for now.

If you are experiencing failures of auto-completion, please report those in the Lean 4 repository.

dalps:
Thank you, I've followed your advice and disabled word-based suggestions for Lean files, that helps.

After reading vscode-lean4 manual on auto-completion, my instance of the extension is indeed behaving correctly; also I didn't know that some snippets were already available through code actions.

I'd be happy to see the kind of snippet I proposed integrated in vscode-lean4 one day, or even keyword suggestions. But I agree that they should be done in a context-sensitive and configurable way. Until then, feel free to close this PR.

mhuisi:
Alright, I'll close this PR. Thanks for the suggestions regardless, there's definitely lots that still needs to be improved about the Lean editing experience :-)

Issue #470, opened 2024-06-18
RFC: indicator for out of date infoview


hargoniX:
### Proposal
Show some form of indicator for when the goal state that's currently being presented in the info view is out of date / being recomputed.

  - **User Experience**: Users can clearly judge when the information they are being presented is current. With recent features such as incrementality or the planned feature to show goals before the entire proof was elaborated this type of information becomes more important.

  - **Beneficiaries**: People that write large proofs on Lean releases with incrementality etc. know precisely when the information they are looking at is out of date/up to date.

  - **Maintainability**: Unclear.
 

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/vscode-lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others benefit from the changes in this proposal being added, please ask them to add :+1: to it.

Vtec234:
Have you noticed that a yellow tint appears on the location (`myfile.lean:12:34`) when it is loading? Are you requesting a more prominent indication?

hargoniX:
I do think a more prominent indication would be sensible yes. I talked to a few users and none of them (including myself) even noticed that this feature exists, let alone its meaning.

Issue #54, opened 2021-10-18
Keyboard accessibility warnings from Tippy


lovettchris:
We are getting these warnings in while debugging when hovering over a type and getting the tooltip popups:

"Interactive tippy element may not be accessible via keyboard navigation because it is not directly after the reference element in the DOM source order. Using a wrapper &lt;div> or &lt;span> tag around the reference element solves this by creating a new parentNode context. Specifying `appendTo: document.body` silences this warning, but it assumes you are using a focus management solution to handle keyboard navigation. See: https://atomiks.github.io/tippyjs/v6/accessibility/#interactivity
"

PR #561, opened 2024-12-22
feat: add a setting to disable the loading spinner animation

YnirPaz:
The loading spinner animation in the infoview can be distracting. This PR adds a setting, disabled by default, that removes the animation.

### DIFF
--- a/lean4-infoview-api/src/infoviewApi.ts
+++ b/lean4-infoview-api/src/infoviewApi.ts
@@ -86,6 +86,7 @@ export interface InfoviewConfig {
     emphasizeFirstGoal: boolean
     reverseTacticState: boolean
     showTooltipOnHover: boolean
+    disableLoadingSpinner: boolean
 }
 
 export const defaultInfoviewConfig: InfoviewConfig = {
@@ -97,6 +98,7 @@ export const defaultInfoviewConfig: InfoviewConfig = {
     emphasizeFirstGoal: false,
     reverseTacticState: false,
     showTooltipOnHover: true,
+    disableLoadingSpinner: false,
 }
 
 export type InfoviewActionKind =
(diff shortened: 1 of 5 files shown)


Vtec234:
You seem to have a different opinion than #470 :) IMO adding a new option for this rather specific preference is too much maintenance burden. However, the infoview CSS can be tweaked using the 'Lean 4 > Infoview: Style' option, so you should already be able to make the spinner invisible. But it might be a bit hard because it has no specific classes; so how about this: have the PR add a 'loading-spinner' which does nothing to `className`. Then you can easily make elements with that class invisible in 'Infoview: Style'.

YnirPaz:
> You seem to have a different opinion than #470 :) IMO adding a new option for this rather specific preference is too much maintenance burden. However, the infoview CSS can be tweaked using the 'Lean 4 > Infoview: Style' option, so you should already be able to make the spinner invisible. But it might be a bit hard because it has no specific classes; so how about this: have the PR add a 'loading-spinner' which does nothing to `className`. Then you can easily make elements with that class invisible in 'Infoview: Style'.

Alright, I'll do this. But I will say I really don't like having an accessibility "feature" only available for people who have the know-how to code it. I had to make my own version of the extension without the spinner just to be able to use lean remotely comfortably, and it's been frustrating to say the least.

Hopefully this doesn't come off as oversensitive.

Vtec234:
I didn't realize this was an accessibility issue rather than just personal preference - that certainly changes things. Owing to very limited dev resources, the accessibility story isn't great (see e.g. #54), but I don't see why not start fixing it here.

I still think just disabling the spinner sounds overly specific, though, and a more general accessibility option would be better. If it is okay to ask here, would it help you if the spinner icon did still appear, but didn't move (i.e., a static icon would show up)? If so, there could be an 'Accessibility > Reduce Motion' option which should set the CSS featutre [prefers-reduced-motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion). If this would not suffice, do you happen to know if there is another general accessibility option that completely disabling the spinner (i.e., not having it show up at all) could fall under?

Finally, it might be that the current maintainers (CC @mhuisi) are happy with the PR as-is, in which case I certainly wouldn't oppose it.

PR #496, opened 2024-07-06
feat: [lean4web] abstract lean client setup for websocket clients

abentkamp:
This is a pure refactoring. The behavior of the code has not changed.

This allows us to replace local lean clients by websocket lean clients.

PR #568, opened 2025-01-13
feat: [lean4web] abstract lean client setup for websocket clients

joneugster:
This is a pure refactoring. The behavior of the code has not changed.

This allows us to replace local lean clients by websocket lean clients.


joneugster:
recreation of #496

PR #579, opened 2025-02-10
feat: simplex abbreviations

gio256:
Adds abbreviations for `⦋` and `⦌`, which are used in mathlib to denote objects of the simplex category (as of [mathlib4#21565](https://github.com/leanprover-community/mathlib4/pull/21565)).


emilyriehl:
Just here to say I'm a fan of this PR!

PR #585, opened 2025-03-06
feat: editor decorations

mhuisi:
This PR adds several new editor decorations.

### Error and warning gutter decorations

To resolve the issue that warnings and errors can sometimes be a bit difficult to spot, especially when the corresponding diagnostic squiggly is small, this PR adds a system of gutter decorations to make errors and warnings much more visible. 

Error and warning gutter decorations

The circled error and warning symbols denote the line where an error or a warning starts, i.e. where the squiggly is displayed. A red vertical line denotes the *full range* of the error, i.e. the range in which the error is displayed under 'Messages' in the InfoView. The red line bends to the right whenever a full error range ends in that line of code. Single line errors are displayed without a red line.

These gutter decorations can be disabled using the 'Lean4: Show Diagnostic Gutter Decorations' setting. When the gutter icons of the Error Lens extension are enabled, the gutter decorations of the Lean 4 VS Code extension are disabled.

Once this PR is merged, this feature will be available in all Lean 4 versions.

### 'Unsolved goals' markers

To resolve the issue that it can sometimes be a bit difficult to tell where one should continue typing in a proof, this PR adds an end-of-line editor decoration that is displayed in the line where the full range of an 'unsolved goals' error ends. 

Unsolved goals markers

The 'unsolved goals' markers can be disabled using the 'Lean 4: Show Unsolved Goal Decoration' setting.

This feature needs language server support, so it will only be active for Lean versions that include the companion PR at leanprover/lean4#7366.

### Goals accomplished!

To re-introduce some positive reinforcement into Lean 4 after the 'Goals accomplished!' message from Lean 3 was removed due to often being confusing in Lean 4, this PR adds a new 'Goals accomplished!' mechanism when a theorem or a `Prop`-typed `example` has been proven.

Goals accomplished

When a theorem or a `Prop`-typed `example` contains no errors or `sorry`s anymore, two blue checkmarks appear next to the start of the declaration as a gutter decoration. Additionally, a 'Goals accomplished!' message appears under 'Messages' in the InfoView. The 'Goals accomplished!' message does not appear in 'All Messages' and is not displayed with a squiggly line in the editor.

The 'Goals accomplished!' gutter decoration can be configured or disabled using the 'Lean 4: Goals Accomplished Decoration Kind' setting. In addition to the default double checkmark decoration pictured above, this PR adds the following additional styles for the gutter decoration:

- Circled checkmark:

Circled checkmark

- Octopus:

Octopus

- Tada:

Tada

This feature needs language server support, so it will only be active for Lean versions that include the companion PR at leanprover/lean4#7366.

### Technical notes
- We intend to test this implementation for the 'unsolved goals' markers and 'Goals accomplished!' in core before v4.19.0-rc1 to iron out any remaining issues.
- All gutter decorations are implemented using .svgs, which is the only API that VS Code provides to display things in the gutter. Since there are lots of different cases to consider for the red vertical error range lines, as well as for light and dark themes, this PR ships 56 .svg files in total.
- The design language of the red lines for the error range gutter decorations is deliberately not perfectly unambiguous. For example, when multiple full error ranges overlap, it isn't clear which error range is terminated by the line bending to the right. The design in this PR is the best I could come up with for the limited size of the gutter without introducing too much visual complexity or noise.
- The task gutter had to be rewritten entirely for this PR, including the progress bar code.
- Since the line where the full range of an 'unsolved goals' error ends is relatively unstable as the document is being edited, this PR implements an edit delay of 3000ms for 'unsolved goals' markers, so that they do not constantly jump around while editing the document.
- The 'unsolved goals' markers use an emoji (specifically the 'Hammer and wrench' emoji U+1F6E0) instead of an icon. The reason for this is that VS Code's decoration API appears to be pretty broken for end-of-line icons and fails to vertically center them correctly.
- None of these decorations have an associated hover to explain the semantics of each decoration. The reason for this is that VS Code doesn't support gutter decoration hovers and for end-of-line decorations, it pollutes the hover of every hover in the whole line with the hover message of the end-of-line decoration.
- This PR includes the client-side for the protocol extensions defined at leanprover/lean4#7366.


mhuisi:
> This looks great! I think you should announce this on Zulip, and will get very much (positive) feedback.
> 
> I think users will also appreciate a the 🎉 emoji appended to the `Goals accomplished!` message.

I asked around a bit, but some users mentioned that they find the 🎉 emoji to be infantilizing, so I decided that it shouldn't be the default. Making the message configurable on a per-user basis is also a bit annoying, so IMO the message should just not contain any emojis, and users that like the emoji can use it in the corresponding editor decoration.

PatrickMassot:
Thank you so much Marc, it looks **great**! And it’s also really really useful.

PR #596, opened 2025-03-26
feat: add uri handler for the setup-guide

algebraic-dev:
I’m not certain whether this is the best place to add a new `registerUriHandler`. However, it's a hook for the URI `vscode://leanprover.lean4/setup-guide`, which will open the setup guide. This is useful for future documentation in the web, especially when detailing the installation process.

### DIFF
--- a/vscode-lean4/src/extension.ts
+++ b/vscode-lean4/src/extension.ts
@@ -1,6 +1,6 @@
 import * as os from 'os'
 import * as path from 'path'
-import { commands, ExtensionContext, extensions, TextDocument, window, workspace } from 'vscode'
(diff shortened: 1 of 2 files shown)


vscode-lean4/src/extension.ts
```diff
@@ -152,6 +152,14 @@ function activateAlwaysEnabledFeatures(context: ExtensionContext): AlwaysEnabled
     const elanCommandProvider = new ElanCommandProvider(outputChannel)
     context.subscriptions.push(elanCommandProvider)
 
+    window.registerUriHandler({
```
mhuisi, line 155: Needs to be added to `context.subscriptions`

PR #599, opened 2025-04-03
chore: rename the unused "8<" abbreviation as "scissor" to fix subscript ₈ conflict

euprunin:
Rename the unused "8<" abbreviation as "scissor".

This resolves the need for an extra space when typing "\8" to get ₈, and makes "\8" = ₈ work like all other subscript digits (e.g. "\0" = ₀, "\1" = ₁).

"8<" is not used in Mathlib or Lean.


mhuisi:
Why not rename it to `\scissor` instead?

PR #606, opened 2025-04-11
feat: revamp infoview options and actions

### LINKED ISSUE #592: Improve discoverability of InfoView settings and actions
We've heard some reports of users not being aware that features like the filter menu or the reverse list setting exist because they didn't find the icons in the InfoView.

As a compomise, we might make some of these functions available in the right-click menu or expose them in the forall menu so that they are more easily discoverable without also attracting attention at all times.

mhuisi:
This PR revamps the option and action UX of the InfoView. This work is motivated by several reports from users that they never knew that certain InfoView functionality existed, and intends to make these features more discoverable. Closes #592.

Specifically, the following adjustments are made:
- The filter menu was difficult to discover and users had trouble finding the 'reverse tactic state' functionality. To improve this, all goal state settings have been moved into a single 'Settings' menu that uses a standard cog icon, which should hopefully be much more recognizable. 
- All goal state settings are now also available in the 'Settings' submenu of the InfoView context menu.
- The 'Unselect All' context menu entry is now only visible when there is something to be unselected, as is the case for other context menu entries.
- All goal state settings can be persisted to the VS Code user settings with a context menu action or a button in the settings menu. Clients that do not support this functionality can hide the button in the settings menu by hiding the elements with `saveConfigButton` and `saveConfigLineBreak` classes.
- The `reverseTacticState`, `showGoalNames` and `emphasizeFirstGoal` settings have been made available in the InfoView goal state settings menu.
- All InfoView actions have been made available in the context menu.
- The 'Copy to comment' and 'Copy to clipboard' actions have been removed. As of [vscode#206529](https://github.com/microsoft/vscode/pull/206529), copying from VS Code webviews should work correctly out-of-the-box, so there is no need for these additional buttons that may unnecessarily complicate the UI anymore.
- The 'Refresh' action has been adjusted so that it is only displayed when the InfoView is paused to explicitly hint at the fact that it can be used to refresh the goal state while the InfoView is paused. Many users used to think that this was just the "Retry if InfoView is broken" button, which is practically never necessary.
- The tooltips of all InfoView icons have been adjusted to be a bit more descriptive.
- The `lean4.infoview.showExpectedType` setting has been deprecated in favor of the new `lean4.infoview.expectedTypeVisibility` setting, which allows hiding the expected type section entirely, as this section was reported to be confusing for students in a teaching environment.

<details>
<summary>Adjusted InfoView UI (Click to open)</summary>

Adjusted InfoView UI

</details>

<details>
<summary>New settings menu (Click to open)</summary>

New settings menu

</details>

<details>
<summary>New InfoView context menu (Click to open)</summary>

New InfoView context menu

</details>

### DIFF
--- a/lean4-infoview-api/package.json
+++ b/lean4-infoview-api/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@leanprover/infoview-api",
-    "version": "0.6.0",
+    "version": "0.7.0",
     "description": "Types and API for @leanprover/infoview.",
     "scripts": {
         "watch": "tsc --watch",
--- a/lean4-infoview-api/src/infoviewApi.ts
+++ b/lean4-infoview-api/src/infoviewApi.ts
@@ -19,7 +19,7 @@ export interface InfoviewConfig {
     reverseTacticState: boolean
     hideTypeAssumptions: boolean
     hideInstanceAssumptions: boolean
-    hideHiddenAssumptions: boolean
+    hideInaccessibleAssumptions: boolean
     hideLetValues: boolean
     showTooltipOnHover: boolean
 }
@@ -106,7 +106,7 @@ export const defaultInfoviewConfig: InfoviewConfig = {
     reverseTacticState: false,
     hideTypeAssumptions: false,
     hideInstanceAssumptions: false,
-    hideHiddenAssumptions: false,
+    hideInaccessibleAssumptions: false,
     hideLetValues: false,
     showTooltipOnHover: true,
 }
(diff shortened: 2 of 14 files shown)


Vtec234 (approved):
I have not looked at the implementation at all, but everything seems to work when tested locally.

ctchou:
Why were the 'Copy to comment' and 'Copy to clipboard' actions removed without any announcement or discussion in the Lean community Zulip?  I used those features often and was dismayed that they just suddenly disappeared.

PR #643, opened 2025-08-12
Fix `x[...]'...` notation

### LINKED ISSUE #609: Notation `x[...]'...` gets confused with a character literal
### Description

Indexing notation with proof is sometimes mistaken with a character literal, which can cause formatting to look weird later:

Image

### Context

I discovered this while working on hash map code:
https://github.com/leanprover/lean4/pull/7400#discussion_r2046149116

### Steps to Reproduce

```lean
@[simp]
def foo (x : Array Nat) (i : Nat) (h' : i < x.size) :
    0 ≤ x[i]'h' :=
  Nat.zero_le _

@[simp]
def bar (x : Array Nat) (i' : Nat) (h : i' < x.size) :
    0 ≤ x[i']'h :=
  Nat.zero_le _

@[simp]
def baz : 2 = 2 := rfl
```

**Expected behavior:**
The `[...]` should be detected as a closed pair of brackets, the `'` afterwards shouldn't get any special formatting.

**Actual behavior:**
In the first case, the `[..]` is detected as closed pair of brackets; however, the `'` afterwards gets highlighted like a character literal, the remainder of the "character literal" is overriden by the semantic highlighting information from the server.
In the second case it treats `[` as unclosed and `']'` as a character literal (although the first quote gets again overriden by semantic highlighting). Due to fact that it treats `[` as unclosed, the remainder has slightly wrong highlighting, e.g. `@[simp]` in `baz` isn't recognized as an attribute.

### Versions

Version 0.0.202

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/vscode-lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

mdbrnowski:
Closes #609.


mhuisi:
All good, TextMate grammars are really tough to get "right" (and by that I mean that there are fewer bugs, not no bugs) :-)

PR #694, opened 2025-12-08
test: add tests for `lakefile-toml-schema.json`

maxwell3025:
This pull request adds tests to make sure that [lakefile-toml-schema](https://raw.githubusercontent.com/leanprover/lean4/refs/heads/master/src/lake/schemas/lakefile-toml-schema.json) has no false positives and minimal false negatives *(or the other way around? Maintainer please confirm/deny)*.

As discussed in [this pull request](https://github.com/leanprover/lean4/pull/11536), tests for the `lakefile.toml` schema are placed here since the tests in this repo are better maintained and this location avoids adding another dependency to the test suite for the main project.

### DIFF
--- a/vscode-lean4/test/suite/lakefileTomlSchema/lakefileTomlSchema.test.ts
+++ b/vscode-lean4/test/suite/lakefileTomlSchema/lakefileTomlSchema.test.ts
@@ -18,7 +18,7 @@ suite('Tests', () => {
             logger.log(`=================== Ensure ${testFileName} is rejected ===================`)
             assert(vscode.workspace.workspaceFolders !== undefined, 'No workspace folder is opened')
             assert(vscode.workspace.workspaceFolders.length === 1, 'Exactly one workspace folder should be opened ')
-            const workspaceFolder = vscode.workspace.workspaceFolders[0].uri.path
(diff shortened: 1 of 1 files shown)


vscode-lean4/test/suite/lakefileTomlSchema/lakefileTomlSchema.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+            const document = await vscode.workspace.openTextDocument(lakefilePath)
+            await vscode.window.showTextDocument(document)
+
+            await waitForActiveExtension('tamasfe.even-better-toml')
+
+            // Wait for 5 seconds for diagnostics to appear
+            await sleep(5 * 1000)
+
+            const diagnostics = vscode.languages.getDiagnostics().flatMap(([, diagnostic]) => diagnostic)
```
mhuisi: I wonder if there is something a little more robust we could do here, like listening for diagnostic changes before opening the document and then using the first set of diagnostics that is reported by even-better-toml (or timing out after 60s), which would rely on the assumption that even-better-toml only reports diagnostics once, but the limitations of the VS Code API ensure that we won't really have a perfectly robust solution here either way.

If the 5s delay works consistently, then I'm fine with it :-)

PR #705, opened 2026-02-12
add icons for `.lean` files - default blue.

spaceunifyfifty:
Both a white and blue variant of the favicon logo found on the official Lean4 website are included, though only the blue variant is used by default.

image

image


mhuisi:
We recently updated our icons:
favicon-dark-64
favicon-light-64

Happy to merge once this PR uses these icons.

Issue #693, opened 2025-11-26
RFC: Automatic file reloading when dependencies change


hgoldstein95:
### Proposal

I'd like a setting that always calls "Restart Lean server for file" in open editor tabs when their dependencies change. Sometimes I'm working on one file `Foo.lean` with a library function/tactic and another `Bar.lean` that calls that function/tactic. It'd be nice if I could have a mode where I didn't need to manually open `Bar.lean` and run the "Restart Lean server for file" action to see if my change in Foo actually had the intended effect.

This has been a frustration for me for a while, but it's especially annoying for me now that I'm playing around with using an AI agent to edit code — the agent often reads LSP output from a file I haven't reloaded in a while and concludes that a change it made didn't work, even though the issue is really just that the file needs to be reloaded.

Concretely:

  - **User Experience**: Simplifies user workflow when working on multiple files that depend on one another, especially with the help of automated tools.

  - **Beneficiaries**: All Lean users who wish to split their code across files.

  - **Maintainability**: I could imagine a world where this is just default behavior, in which case it doesn't seem more or less maintainable. As a setting, I suppose it does add one more thing to maintain / test.

### Community Feedback

This was discussed in [this thread](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/Mode.20for.20automatically.20updating.20downstream.20files.3F/with/560400550).

mhuisi:
I have some further questions about this RFC:
1. Do you expect dependent files to be restarted only when the file you are looking at is saved, or automatically after a change has been made?
2. Do you only expect to see errors for open editor tabs, or for files that have not been opened as editor tabs as well? (I.e.: do you expect a full rebuild of the entire project, or only for the files that are currently open as tabs?)
3. Are you confident that your AI agent will correctly interpret the set of diagnostics produced by the language server once dependent files are being reloaded automatically? 
  (I'm asking because we've heard reports in the past that AI agents will not use the correct set of diagnostics even when looking at a single file, since LSP was not designed with AI agents in mind and there is no standardized way for push-diagnostics to designate themselves as being the "complete" set of diagnostics for the fully elaborated file that an AI agent should use, as opposed to one from the partial elaboration states)

[...]

PR #776, opened 2026-04-25
Auto restart files with stale dependencies

samuela:
When working in multi-file projects, I frequently find myself facing "Imports are out of date. Need to restart the file." errors. These are common enough to be quite annoying, esp. considering vscode-lean4 knows the solution and yet does not handle it automatically.

This PR adds a new setting that automatically detects import-out-of-date errors and automatically restarts the file. The flag defaults off in the interest of maintaining backwards compatibility.

### DIFF
--- a/vscode-lean4/src/leanclient.ts
+++ b/vscode-lean4/src/leanclient.ts
@@ -41,6 +41,7 @@ import {
 } from '@leanprover/infoview-api'
 import {
     allowedLoggingMethods,
+    automaticallyRestartFileOnDependencyChange,
     disallowedLoggingMethods,
     isLoggingEnabled,
     loggingDir,
@@ -56,6 +57,7 @@ import { DiagnosticChangeEvent, LeanClientDiagnosticCollection } from './diagnos
 import { formatCommandExecutionOutput } from './utils/batch'
 import {
     c2pConverter,
+    LeanDiagnostic,
     LeanImport,
     LeanModule,
     LeanModuleHierarchyImportedByParams,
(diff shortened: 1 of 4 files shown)


vscode-lean4/src/leanclient.ts
```diff
@@ -549,7 +553,9 @@ export class LeanClient implements Disposable {
             const vsCodeCollection = languages.createDiagnosticCollection('lean4')
             this.diagnosticCollection = new LeanClientDiagnosticCollection(vsCodeCollection)
             this.diagnosticCollection.onDidChangeDiagnostics((e: DiagnosticChangeEvent) => {
-                this.diagnosticsEmitter.fire(e.accumulatedParams())
+                const params = e.accumulatedParams()
+                this.diagnosticsEmitter.fire(params)
+                void this.maybeAutoRestartFileWithStaleDependencies(params)
```
SnirBroshi, line 558: Is this `void` necessary? Is there a lint for unused return value?

vscode-lean4/src/leanclient.ts
```diff
@@ -621,7 +627,46 @@ export class LeanClient implements Disposable {
...
+        }
+
+        const uri = fileUri.toString()
+        const hasStaleDependencyDiagnostic = params.diagnostics.some(d => this.isStaleDependencyDiagnostic(d))
+        if (!hasStaleDependencyDiagnostic) {
+            this.autoRestartedStaleDependencyUris.delete(uri)
+            return
+        }
+
+        if (!automaticallyRestartFileOnDependencyChange() || this.autoRestartedStaleDependencyUris.has(uri)) {
```
SnirBroshi: I think the setting check should be at the top of this function, otherwise we're doing work for nothing

vscode-lean4/src/leanclient.ts
samuela: yes exactly, the @typescript-eslint/no-floating-promises lint checks for this

vscode-lean4/src/leanclient.ts
samuela: good point, done in 85d45f26

medovina:
Issue #693 has a discussion about adding this feature.

PR #778, opened 2026-05-02
test: new test framework

mhuisi:
This PR adds a new test framework.

Most of this test framework was generated by Claude and then refined.

### Other changes
- The manifest parsing is slightly relaxed to allow for non-URL Git dependencies
- The InfoView gains some `data-id`s for better testability
- Logging calls in the extension itself are removed
- The "Server process exited with code 0" message in the output channel when restarting the server is suppressed


package-lock.json
mhuisi, line 1: (The diff size on `package-lock.json` is misleading - `git diff --histogram` shows mostly additions, as expected)

